PR #22, opened 2025-09-03
updates to embedder service

tyhouch:
Problem

  The embedding service was failing in production because it used stateful in-memory storage (_embedders = {} dictionary) which
  doesn't work with Ray Serve's distributed architecture. When requests were load-balanced across multiple replicas, the embedder
  initialized on one replica couldn't be found on another, causing 404 errors.

  Solution

  Made the embedding service completely stateless by passing configuration with each request instead of storing embedder instances.


jxstanford (approved):
Note: I pushed a commit to bump the version to 0.5.1 and refresh documentation. Closes: https://github.com/kamiwaza-ai/kamiwaza-sdk/issues/23. 

Verified by deploying this build locally and running notebook 10 against the counterpart PR in kamiwaza (https://github.com/kamiwaza-internal/kamiwaza/pull/464)

PR #60, opened 2026-04-01
ENG-3056: Add kz-ext CLI for extension developer tools (Phase 1)

jonathanleff:
## Summary

- Implements the `kamiwaza-extensions` CLI package (`kz-ext`) with six foundational commands for local extension development: `login`, `create`, `validate`, `dev local`, `doctor`
- Adds multi-connection management with per-connection PAT storage, scaffolding templates for app/tool/service extension types, and metadata + compose file validation
- Includes design documents (system design + state), README documentation, and 96 unit tests covering all components

## Components

| Component | File(s) | Purpose |
|-----------|---------|---------|
| CLI skeleton | `cli.py`, `commands/` | Typer entry point with `--version`/`--verbose`/`--debug`, error handling |
| ConnectionManager | `connections.py` | Multi-connection CRUD, atomic writes, chmod 600 token files |
| MetadataValidator | `validators/metadata.py` | Pydantic-based kamiwaza.json validation with semver + naming checks |
| ComposeValidator | `validators/compose.py` | docker-compose.yml deployment compatibility checks |
| DevLocalRunner | `dev_local.py` | Extension detection, env overlay, Docker Compose v1/v2, signal forwarding |
| DoctorChecker | `doctor.py` | Python, Docker, Compose, connection health, CLI compat, runtime lib checks |
| Scaffolder | `scaffolder.py` | Template rendering with name validation and auto-prefixing |
| Bundled Templates | `templates/{app,tool,service}/` | Next.js+FastAPI, FastMCP, and minimal service scaffolds |

## Test plan

- [x] 96 unit tests across 9 test files — all passing
- [x] 131 existing SDK tests — no regressions (227 total)
- [x] E2E flow: `create` → `validate` for all three extension types
- [x] CLI smoke test: `kz-ext --help`, `--version`, all subcommand help screens
- [x] Scaffolded output passes `kz-ext validate` for all template types


jxstanford (dismissed):
# Review Report: ENG-3056: Add kz-ext CLI for extension developer tools (Phase 1)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/60
**Branch**: `feat/kamiwaza-extensions` → `develop`
**Engines**: Claude Code, Kamiwaza (2/4 available)

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| | | | | |

## High Priority Issues
| H1 | `connections.py:149` | **Path traversal via unsanitized connection names.** `_token_path()` builds `self._connections_dir / name / "token.json"` with no validation. A name like `../../.ssh` resolves outside the intended directory. `remove_connection()` (line 80-86) then calls `unlink()` and `rmdir()` on the resolved path — capable of deleting arbitrary `token.json` files and empty directories. Fix: validate names against `^[a-zA-Z0-9_-]+$`. | Claude Code | Security — Path Traversal |
| H2 | `commands/login.py:91-93` | **Silent swallow of connection validation errors.** `_validate_connection()` catches all `requests.RequestException` and passes silently. Users get zero feedback if the URL is wrong or the server is unreachable, leading to confusing failures later in `dev local` or other commands. At minimum, print a warning so users know validation was skipped. | **Both engines** ⚠️ | Error Handling / UX |
| H3 | `dev_local.py:145` | **Fragile URL manipulation with `str.replace`.** `url.replace("/api", "")` replaces *all* occurrences, not just the trailing one. A URL like `https://example.com/api-gateway/api` produces `https://example.com-gateway/` — corrupting `KAMIWAZA_PUBLIC_API_URL`. Fix: use `url.removesuffix("/api")` (Python 3.9+). | Claude Code | Logic — Correctness |
| H4 | `commands/login.py:85` | **Possible SDK API contract mismatch.** Kamiwaza engine flags that `KamiwazaClient` is instantiated with `username`/`password` params, but the design doc (Section 3.2) says the client only accepts `base_url`, `api_key`, or `authenticator`. If accurate, this causes a `TypeError` at runtime on the login flow. **Needs verification** against actual SDK constructor signature — Claude Code did not flag this, so it may be a false positive based on stale design doc. | Kamiwaza | Bug (unverified) |

| M1 | `validators/compose.py:14`, `validators/metadata.py:19` | **Duplicate `ValidationResult` dataclass** defined identically in two modules. Extract to shared location to avoid divergence. | Claude Code | DRY |
| M2 | `commands/create.py:10`, `scaffolder.py:20` | **Duplicate `VALID_TYPES` constant.** Single source of truth should live in `scaffolder.py` with `create.py` importing it. | Claude Code | DRY |
| M3 | `commands/validate.py:74-79`, `dev_local.py:20-25` | **Duplicate compose filename detection logic.** Same list in two modules. Extract to shared constant. | Claude Code | DRY |
| M4 | `connections.py:158-161, 175-193` | **File permissions race.** Token/config files briefly exist with default permissions (0644) before `os.chmod(0o600)` is applied. Write to temp file, chmod, then rename — or at least chmod immediately after creation. | Claude Code | Security — Defense in Depth |
| M5 | `templates/app/backend/app/main.py:11-16` | **CORS `allow_origins=["*"]` with `allow_credentials=True`** in scaffolded template. Use `allow_origins=["http://localhost:3000"]` for dev template to match the actual frontend and avoid signaling insecure defaults. | Claude Code | Security — Misconfiguration |
| M6 | `cli.py:75-95` | **Overly broad exception handling.** `run_with_error_handling` catches bare `Exception`, which can mask programming errors (`AttributeError`, `NameError`). Consider catching `KamiwazaError` + specific CLI exceptions only, letting unexpected errors surface with traceback in `--debug` mode. | Kamiwaza | Code Quality |
| M7 | `commands/login.py:88` | **Token expiry handling.** PATs stored with `expires_at=0.0` and `refresh_token=None`. If `FileTokenStore` interprets `0.0` as expired, this causes immediate auth failures. Verify SDK expectations for non-expiring PATs. | Kamiwaza | Logic |
| M8 | `commands/login.py:40-50` | **Missing URL validation.** No check that the provided URL is valid HTTP/HTTPS or points to a Kamiwaza endpoint. Malformed URLs fail later with cryptic errors. | Kamiwaza | Validation |

## False Positive (Dismissed)
| # | File:Line | Issue | Engine | Reason |
|---|-----------|-------|--------|--------|
| FP1 | `connections.py:165` | "File truncated mid-function" | Kamiwaza | Claude Code reviewed the complete file (193 lines) with specific findings at lines 149, 158, 175 — file is intact. Kamiwaza likely hit a context/token limit during extraction. |
| FP2 | `connections.py` | "Atomic config writes missing" | Kamiwaza | Claude Code confirms the temp file + `os.replace()` pattern is already implemented (line 175-193). Kamiwaza couldn't see it due to truncation. |

## Documentation Issues
| D1 | Design doc §3.9 vs `validators/compose.py` | Design doc lists host port bindings, bind mounts, container_name, custom networks as **Error** severity, but implementation downgrades all to **warnings**. Only missing Dockerfiles remain errors. Likely intentional for dev ergonomics but rationale not documented. | Claude Code |

## Test Coverage Assessment
- **96 tests** with excellent scaffold-validate integration coverage
- **Well-covered**: `cli.py`, `connections.py`, `validators/`, `scaffolder.py`, `doctor.py`, `commands/validate.py`, `commands/create.py`
- **Gaps needing tests**:
  - `commands/login.py`: `_login_with_password()` interactive flow untested (use `CliRunner.invoke(input=...)`)
  - `dev_local.py`: `DevLocalRunner.run()` full flow and `_run_subprocess()` signal forwarding untested
  - `commands/dev_local.py`: No CLI-level test for `kz-ext dev local`
  - `cli.py`: `_handle_exception()` with `KamiwazaError` branch untested

## Positive Observations
- Clean architecture with excellent separation: CLI skeleton → commands → business logic (connections, validators, scaffolder, doctor, dev_local) — **both engines agree**
- Thorough test suite (96 tests) with scaffold→validate integration tests ensuring templates stay consistent with validation rules
- Security-conscious credential handling: `chmod 600/700`, `hide_input=True` for passwords, atomic config writes
- Modern CLI UX with Typer + Rich styled output — **both engines agree**
- Robust Docker Compose v1/v2 detection with proper timeout handling
- Forward-compatible Pydantic model with `extra="allow"` for metadata schema
- Signal forwarding in subprocess management for clean container shutdown
- Backward-compatible `~/.kamiwaza/token.json` preservation alongside new multi-connection config

## Overall Verdict
**Risk Level**: Medium
**Verdict**: **APPROVE**
**Engine Agreement**: 2/4 engines contributed (Codex errored; Gemini unavailable). Claude Code and Kamiwaza agree on architecture quality; disagree on severity of validation silence (M1 vs Critical — elevated to H2 here due to cross-engine agreement). Kamiwaza's two other Critical findings were dismissed as artifacts of file truncation.

**Rationale**: Well-structured Phase 1 CLI with solid test coverage (96 tests) and good architectural patterns. The three confirmed high-priority issues are all straightforward fixes:
- **H1** (path traversal): Add a regex check on connection names (~3 lines)
- **H2** (silent validation): Add a warning print in the except block (~1 line)
- **H3** (URL replace): Swap `str.replace` for `removesuffix` (~1 line)

**H4** (SDK API mismatch) should be verified against the actual SDK constructor — if confirmed, it's a blocking runtime bug on the login path.

The DRY violations (M1-M3) are typical of initial implementations and easy to consolidate in a follow-up.

---

jonathanleff (commented):
Thanks for the thorough review! All high-priority issues and several medium ones addressed in e0b9b86:

### High Priority — All Fixed

| # | Fix |
|---|-----|
| **H1** | Connection names now validated against `^[a-zA-Z0-9_-]+$` before any filesystem ops. Added 3 tests for path traversal / special chars. |
| **H2** | `_validate_connection()` now prints a warning with the error details instead of silently passing. |
| **H3** | Replaced `url.replace("/api", "")` with `url.removesuffix("/api")`. Added regression test for mid-URL `/api`. |
| **H4** | **Confirmed as a real bug.** `KamiwazaClient` does not accept `username`/`password`. Fixed to use `UserPasswordAuthenticator` + `PATCreate` per the actual SDK API. |

### Medium Priority — Addressed

| # | Fix |
|---|-----|
| **M1** | Extracted `ValidationResult` to `validators/result.py`, both validators import from there. |
| **M2** | `commands/create.py` now imports `VALID_TYPES` from `scaffolder.py`. |
| **M3** | Extracted `COMPOSE_FILENAMES` to `constants.py`, shared by `dev_local.py` and `commands/validate.py`. |
| **M5** | CORS `allow_origins` in app template narrowed from `["*"]` to `["http://localhost:3000"]`. |

### Not Addressed (intentional)

| # | Reason |
|---|--------|
| **M4** (permissions race) | The temp-file + rename pattern already provides atomicity for the config. Token files inherit from `FileTokenStore.save()` which uses the same pattern. The chmod-after-write window is extremely narrow and matches the existing SDK behavior. Will revisit if needed. |
| **M6** (broad exception catch) | The `--debug` flag already surfaces full tracebacks for unexpected errors. Narrowing the catch would lose the styled error formatting for common cases like `PermissionError`, `OSError`, etc. that aren't `KamiwazaError`. |
| **M7** (token expiry 0.0) | PATs are long-lived tokens — `expires_at=0.0` is treated as "no expiry" since `FileTokenStore.load()` returns the token regardless and the SDK's `ApiKeyAuthenticator` doesn't check expiry. |
| **M8** (URL validation) | Connection validation (H2 fix) now warns on unreachable URLs. Full URL format validation would add complexity for marginal benefit since invalid URLs fail fast on first API call. |

**Tests:** 100 extension tests + 131 SDK tests = 231 total, all passing.

jxstanford (changes requested):
# RE-REVIEW Report: ENG-3056: Add kz-ext CLI for extension developer tools (Phase 1)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/60
**Branch**: `feat/kamiwaza-extensions` → `develop`
**Engines**: Claude Code, Codex, Kamiwaza (3/4 available)

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Previous Review Feedback — Resolution

The fix commit (`e0b9b86`) addressed all issues from the prior review round: path traversal validation, duplicate `ValidationResult`/`COMPOSE_FILENAMES`, CORS wildcard, `str.replace` mid-URL corruption, silent validation failure, and incorrect SDK API usage. All fixes are clean and well-tested.

## Critical Issues (Must Fix)
| C1 | `commands/login.py:70-71` | **Password login shares SDK token cache — identity confusion.** `UserPasswordAuthenticator` defaults to `FileTokenStore()`, which loads and overwrites `~/.kamiwaza/token.json`. If a cached SDK session exists for the same instance, `authenticate()` reuses that token and `create_pat()` runs under the cached identity — not the username/password just entered. Also mutates the shared SDK token file that the connection manager is designed to leave untouched. | Codex | Security / Correctness |
| C2 | `commands/login.py:52-56,108` | **Invalid credentials stored after validation failure.** When `_validate_connection()` gets an auth failure (mistyped/revoked PAT or wrong URL), it only prints a yellow warning but still persists the token and marks the connection active. `kz-ext login --api-key <bad-key>` appears to succeed; subsequent commands silently use unusable credentials. | Codex, Kamiwaza | Correctness / UX |
| C3 | `commands/login.py:89`, `doctor.py:140` | **Double `/api` in health-check URL.** `_validate_connection` constructs `f"{url}/api/health"` but `url` already includes `/api` (per docs: `https://instance.example.com/api`), producing `.../api/api/health`. Every connection validation and `kz-ext doctor` health check hits the wrong endpoint. Login still succeeds (validation is non-blocking), masking the bug. | Claude Code, Codex | Correctness |

## High Priority Issues
| H1 | `pyproject.toml:36-44` | **`kz-ext` script installed without runtime dependencies.** `kz-ext` is registered as a `[project.scripts]` entry, but `typer`, `pyyaml`, and `packaging` are in the optional `extensions` extra. A plain `pip install kamiwaza-sdk` installs a broken `kz-ext` that crashes on `ModuleNotFoundError`. Either move deps to core or gate the script behind the extra. | Codex | Packaging |
| H2 | `commands/login.py:93` | **Raw `requests.get()` bypasses SDK security config.** `_validate_connection()` calls `requests.get()` directly instead of using `KamiwazaClient`, bypassing `KAMIWAZA_VERIFY_SSL`, custom CA bundles, and retry logic. Creates inconsistency and potential MITM exposure when SSL verification is configured. | Kamiwaza | Security |
| H3 | `commands/login.py:85-91` | **Non-atomic PAT creation + storage.** If the process crashes between server-side PAT creation (line 85) and local token storage (line 91), the user has an orphaned PAT on the server with no local record. Should wrap in try/except to revoke the PAT on storage failure, or use atomic write. | Kamiwaza | Reliability |

| M1 | `connections.py:141-158` | `get_token(name)` and `save_token(token, name)` accept explicit `name` parameter flowing to `_token_path()` without calling `_validate_connection_name()`. Defense-in-depth gap. | Claude Code | Security |
| M2 | `commands/login.py:60-80` | Each `_login_with_password` creates a new server-side PAT (`kz-ext-{name}`) without checking/revoking previous one. Repeated logins accumulate stale PATs. | Claude Code | Correctness |
| M3 | `commands/create.py:20-24` | For `tool`/`service` scaffolds, `Scaffolder.create()` auto-prefixes the directory name, but the command prints the original `name` — success banner and `cd` instruction point to a non-existent directory. | Codex | UX |
| M4 | `cli.py:31` | Module-level `_state` singleton makes CLI difficult to test in parallel and risks state leakage. Consider `typer.Context.obj` instead. | Kamiwaza | Code Quality |
| M5 | `cli.py:55` | Try/except around `KamiwazaError` import silently ignores all import errors. Narrow to `ImportError` to avoid masking broken SDK installations. | Kamiwaza | Debugging |

## Test Coverage Assessment
- **Well covered**: `connections.py`, `dev_local.py`, `doctor.py`, `scaffolder.py`, `validators/`, `cli.py`, `commands/create.py`, `commands/validate.py`, E2E flows (96 tests total)
- **Gaps**: Password login flow only partially covered (API-key path tested, password path exercises SDK auth but doesn't verify token cache isolation). `_validate_connection` request behavior untested. Signal forwarding in `_run_subprocess` not unit-tested (acceptable — hard to test).

## Positive Observations
- Fix commit is clean and focused — each prior review item addressed with minimal, targeted changes
- Regression test for `removesuffix` URL fix is excellent practice
- Path traversal validation with 3 test cases (add, remove, special chars) is thorough
- `ValidationResult` and `COMPOSE_FILENAMES` DRY extractions are well-structured
- CORS scoped to `localhost:3000` is the correct fix for local dev templates
- Good architectural separation: CLI presentation / business logic / infrastructure layers
- Atomic config writes, `chmod 600` on tokens, list-form subprocess calls show security awareness
- Backward compatibility with existing `~/.kamiwaza/token.json` is respected

## Overall Verdict
**Risk Level**: High
**Verdict**: REQUEST_CHANGES
**Engine Agreement**: 3/4 engines contributed; 2 of 3 flag blocking issues

**Rationale**: The fix commit properly resolved all prior review feedback, and the overall architecture is solid. However, three new critical issues surfaced across engines:

1. **C1 (identity confusion)** is a security defect — password login can silently create a PAT under a cached identity rather than the credentials just entered, and mutates the shared SDK token file the design explicitly avoids touching.
2. **C2 (silent credential failure)** means `kz-ext login --api-key <invalid>` appears to succeed, poisoning connection state for all subsequent commands.
3. **C3 (double `/api`)** makes every health check hit the wrong endpoint, confirmed independently by two engines.

C1 and C2 together mean the login flow — the entry point for all authenticated operations — has correctness and security gaps that should be resolved before merge. H1 (broken default install) is also a significant packaging issue that will affect first-time users.

**Recommended fix scope**: C1 (pass isolated token store or `NullTokenStore` to `UserPasswordAuthenticator`), C2 (gate credential persistence on validation success, or prompt user to confirm save on failure), C3 (use `f"{url}/health"` instead of `f"{url}/api/health"`), H1 (move extension deps to core or use console_scripts entry point conditional on extras).

---

jxstanford (changes requested):
# RE-REVIEW Report: ENG-3056: Add kz-ext CLI for extension developer tools (Phase 1)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/60
**Branch**: `feat/kamiwaza-extensions` → `develop`
**Engines**: Claude Code, Codex, Kamiwaza (3/4 available)

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| C1 | `templates/tool/src/server.py:14-15` | **Tool scaffold is non-functional via `dev local`**: Generated server uses `mcp.run(transport="stdio")` which binds no socket, but `docker-compose.yml` publishes port 8000. The documented `kz-ext create --type tool` then `kz-ext dev local` workflow produces a container that cannot accept traffic. Should default to `transport="sse"` or `streamable-http` with a port binding. | Codex | Broken workflow |
| C2 | `commands/login.py:78`, `doctor.py:140` | **Health check loop exits on first `ConnectionError` instead of trying remaining endpoints**: Both `_validate_connection` and `_check_connection` catch exceptions inside the endpoint loop and `return` immediately, preventing fallback to `/auth/health` or `/health` if the first endpoint (`/auth/ping`) is unreachable. Contradicts the "try common health endpoints" design intent. | Kamiwaza, Codex | Bug |
| C3 | `commands/login.py:121-125` | **API key validation accepts bad credentials when public endpoints respond**: If `/auth/ping` and `/auth/health` return 404 but `/health` is public (no auth required), _any_ API key appears valid. The first 2xx response returns `True` without confirming the token was actually used for auth. Bad credentials get persisted and later commands fail. | Codex, Claude Code | Correctness |

## High Priority Issues
| H1 | `commands/login.py:46` | **Global `os.environ` mutation for SSL**: `os.environ["KAMIWAZA_VERIFY_SSL"] = "false"` modifies process-wide state, leaking into any SDK clients or subprocesses spawned later. Should be passed explicitly to the client or scoped narrowly. | Claude Code, Kamiwaza | Security / Code Quality |
| H2 | `commands/login.py:94` | **`--no-verify-ssl` not threaded to `KamiwazaClient` in password login**: `_login_with_password` accepts `verify_ssl` param but never passes it to `KamiwazaClient(base_url=url)`. Password login against self-signed certs will fail even with the flag set. The env var workaround (H1) may mitigate this _if_ the SDK reads it before the client is constructed, but that coupling is fragile. | Claude Code | Correctness |
| H3 | `doctor.py:140-144` | **`_check_connection` ignores SSL preference**: After login with `--no-verify-ssl`, `requests.get()` in doctor still uses default `verify=True`. Self-signed instances will always show as unreachable in `kz-ext doctor`. The SSL preference should be stored per-connection and used here. | Claude Code, Codex, Kamiwaza | Correctness |
| H4 | `commands/validate.py:33-36` | **Missing compose file silently passes validation**: If a directory has valid `kamiwaza.json` but no compose file, `compose_result` stays `None` and the command reports success. This is a false positive for extensions that `dev local` cannot run. | Codex | Logic |
| H5 | `commands/login.py:95-99` | **Unhandled exceptions from PAT creation**: `client.auth.create_pat()` can raise network errors, 401/403 from bad credentials, etc. These bubble up as raw tracebacks instead of user-friendly CLI errors. | Kamiwaza | Error Handling |
| H6 | `dev_local.py:140` | **`KAMIWAZA_USE_AUTH=false` hardcoded without warning**: While intentional for local dev, a misconfigured URL pointing at production would disable auth silently. Add a visible console warning when this override is applied. | Kamiwaza | Security |

| M1 | `connections.py:173-176` | Silent `os.chmod` failures (`pass` in 3 places). Token files could remain world-readable on restricted filesystems. Add debug/warning-level logging. | Claude Code, Kamiwaza | Security |
| M2 | `scaffolder.py:85` | `Path(str(pkg))` assumes `importlib_resources.files()` returns a filesystem path. In zip imports (e.g., zipapp distribution), this breaks. Use `as_file()` context manager. | Kamiwaza | Robustness |
| M3 | `validators/compose.py:45` | Host port binding detection `not port_str.startswith('"')` is effectively dead code — YAML parses `"8000:3000"` as a string without leading quotes. Similarly, bind mount regex `r"^/[^$]"` has edge cases. | Claude Code, Kamiwaza | Logic |
| M4 | `cli.py:95` | Broad `Exception` catch in `run_with_error_handling` could mask `KeyboardInterrupt` if Typer/Click doesn't convert it to `Abort` early enough. | Kamiwaza | Robustness |

## Test Coverage Assessment
**Files with strong coverage**: `connections.py`, `scaffolder.py`, `validators/metadata.py`, `validators/compose.py`, `cli.py`, `dev_local.py` (detection + overlay), `doctor.py` (system + extension checks), `commands/validate.py`. 96 unit tests with good edge-case coverage including path traversal attempts.

**Gaps**:
- `commands/login.py` — password flow (`_login_with_password`) untested; `_validate_connection` edge cases (SSL errors, all-404 path, bad-key-accepted-via-public-endpoint) not covered
- `dev_local.py` — `run()` method end-to-end (Popen + signal forwarding) not tested
- `doctor.py` — `_check_connection` with stored connection + SSL preference not tested
- No test verifying that tool scaffold is actually reachable via docker-compose

## Positive Observations
- **Strong security defaults**: Connection name regex prevents path traversal (with tests for `../../.ssh`), atomic config writes via `tmp_path.replace()`, `chmod 600`/`700` on credentials and config dirs
- **Good identity isolation**: `_MemoryTokenStore` prevents CLI from clobbering shared SDK token during password auth — well-designed fix from round 1
- **Clean architecture**: Clear separation between CLI surface (`cli.py` + `commands/`) and business logic (`connections.py`, `dev_local.py`, etc.)
- **Subprocess safety**: All subprocess calls use list form (no `shell=True`) with timeouts; signal forwarding correctly restores handlers in `finally`
- **Comprehensive validation**: Pydantic-based metadata validation + compose file deployment compatibility checks
- **All previous round-1 critical/high issues resolved** (identity isolation, credential validation, path traversal, DRY refactoring)

## Overall Verdict
**Risk Level**: High
**Verdict**: REQUEST_CHANGES
**Engine Agreement**: 3/4 engines contributed; 2/3 found critical functional bugs

**Rationale**: Three critical issues block merge:
1. **Tool scaffolding is non-functional** — the primary `kz-ext create --type tool` → `kz-ext dev local` workflow produces a container that can't accept traffic (stdio vs network transport mismatch)
2. **Health check loops break on first error** instead of trying remaining endpoints — a logic bug that contradicts the fallback design in both login and doctor flows
3. **API key validation accepts any key** when public health endpoints respond, persisting bad credentials

The codebase is well-structured with strong security practices, and the previous round's issues were thoroughly addressed. These remaining bugs are narrowly scoped fixes (transport default, loop control flow, auth-gated endpoint check) and should not require significant rework.

---

jxstanford (changes requested):
# RE-REVIEW Report: ENG-3056: Add kz-ext CLI for extension developer tools (Phase 1)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/60
**Branch**: `feat/kamiwaza-extensions` → `develop`
**Engines**: Claude Code, Codex, Kamiwaza (3/4 available)

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Previous Review Items — Status

All critical/high issues from the prior review round have been addressed:

| Prior Issue | Status |
|-------------|--------|
| C1: Identity isolation — global token.json clobbered | Fixed (in-memory token store) |
| C2: Credential stored before validation | Fixed (`_validate_token` before `add_connection`) |
| C3: SSL env var set globally, never restored | Fixed (try/finally in `_login_with_password`) |
| H1: Tool template `transport="stdio"` inside Docker | Fixed (now `transport="sse"`) |
| H2: Health endpoints publicly accessible | Fixed (uses `/auth/pats`) |
| H3: Doctor exception handling swallowed all errors | Fixed (per-endpoint try/except) |

## Critical Issues (Must Fix)
| | | (none) | | |

## High Priority Issues
| H1 | `commands/login.py:120-142` | **Token validation accepts unverified tokens on server errors.** `_validate_token` falls through to `return True` when the server returns 500/502/etc. A comment documents the rationale ("different API version"), but this means a completely invalid API key is stored as valid if the auth endpoint is temporarily erroring. At minimum, print a visible warning; ideally require explicit 200 OK for success. | Kamiwaza (Critical), Claude Code (Medium) | Security / Correctness |
| H2 | `__init__.py:3` | **Version mismatch.** `kamiwaza_extensions.__version__` is hardcoded `"0.1.0"` but the package ships as `kamiwaza-sdk==0.11.0`. This breaks `kz-ext --version`, `DoctorChecker._check_cli_version()`, and scaffolded `kz_ext_version` metadata — an installed SDK 0.11.0 will be reported incompatible with its own templates. | Codex | Correctness |
| H3 | `dev_local.py:137-142` | **URL derivation bugs in `build_env_overlay`.** (a) `KAMIWAZA_ENDPOINT` is set to `f"{url}/v1"` but for the default URL shape `https://host/api`, this produces `https://host/api/v1` — the SDK's OpenAI-compatible endpoints expect `/v1` off the *public base URL*, not off `/api`. (b) `removesuffix("/api")` for `KAMIWAZA_PUBLIC_API_URL` is fragile — a URL like `https://host/api/v2` produces `https://host/v2`. Use `urllib.parse` for robust path handling. | Codex, Kamiwaza | Logic |
| H4 | `dev_local.py:137-142` | **`KAMIWAZA_VERIFY_SSL` not propagated to dev containers.** `build_env_overlay` omits `verify_ssl` from `ConnectionInfo`. Extensions using the SDK against self-signed dev instances will fail TLS verification inside Docker even when the connection was created with `--no-verify-ssl`. | Codex | Correctness |
| H5 | `pyproject.toml:46-48` | **Template dotfiles excluded from wheel builds.** The `templates/**/*` glob in `package-data` does not match dotfiles (e.g., `.gitignore`). Scaffolded projects from pip-installed builds will be missing ignore files that exist in the source tree. Use `templates/**` with `include_package_data` or explicit dotfile patterns. | Codex | Distribution |

| M1 | `doctor.py:157` | Dead `last_error` variable — assigned but never read. Leftover from health-check refactor. | Claude Code | Code quality |
| M2 | `validators/compose.py:45` | Dead logic — `not port_str.startswith('"')` can never be true since `str(port)` from YAML-parsed values never produces a leading quote. Harmless but misleading. | Claude Code | Code quality |
| M3 | `commands/login.py:108-112` | SSL env var mutation via `os.environ` is scoped with try/finally (good) but not thread-safe — concurrent SDK operations in the same process could see the modified env var. Consider passing `verify_ssl` directly to the client constructor. | Kamiwaza | Concurrency |
| M4 | `scaffolder.py:116` | `__version__.split(".")[0]` for version parsing — fragile for pre-release versions. `packaging.version.Version` (already a dependency) handles this correctly. | Kamiwaza | Robustness |
| M5 | `doctor.py:189-200` | Health check validates on HTTP 200 status alone without verifying the response body. A load balancer default page could pass as a valid Kamiwaza endpoint. | Kamiwaza | Reliability |

## Test Coverage Assessment
**Good coverage overall** — 96 unit tests with edge cases for path traversal, corrupt configs, name validation, compose v1/v2 fallback, and scaffold-then-validate integration tests.

**Gaps identified (all engines agree):**
- `_login_with_password` (password → PAT creation flow) has no direct test; only the API-key path is covered
- `_validate_token` HTTP scenario matrix (200, 401, 500, ConnectionError) is untested — especially relevant given H1
- `_check_connection` with an active connection (health-endpoint retry loop) is untested
- `_run_subprocess` signal forwarding and full `DevLocalRunner.run()` flow are untested

## Positive Observations
- **Previous feedback thoroughly addressed** — all critical/high issues from round 2 are resolved with appropriate fixes (all 3 engines confirm)
- **Strong security posture** — filesystem permissions (600/700), path traversal prevention, atomic config writes, identity isolation (Claude Code, Kamiwaza)
- **Clean architecture** — separation between CLI commands (thin wrappers), business logic, and templates; `config_dir` injection enables easy testing (Claude Code, Kamiwaza)
- **Good error handling** — `run_with_error_handling` decorator, per-endpoint try/except in doctor, signal forwarding for clean Docker shutdown (all engines)
- **Smart integration testing** — scaffold-then-validate tests keep templates and validation in sync (Claude Code)

## Overall Verdict
**Risk Level**: Medium
**Verdict**: REQUEST_CHANGES
**Engine Agreement**: 3/4 engines contributed (Claude Code: APPROVE with minor suggestions; Codex: 4 functional issues; Kamiwaza: 1 critical + 2 high)

**Rationale**: The codebase is well-structured and all prior critical/high issues are resolved. However, this round surfaced new functional bugs that will affect users: the version mismatch (H2) breaks `--version` and doctor compatibility checks; the URL derivation bugs (H3) will inject wrong environment variables during `kz-ext dev local`; and missing SSL propagation (H4) breaks self-signed cert workflows. H1 (token validation bypass) is a lower-risk design concern given the documented rationale, but should at minimum surface a visible warning. H5 (dotfiles in wheels) affects distribution correctness. Recommend fixing H2-H4 before merge; H1 and H5 are strongly recommended but could be follow-up if needed.

---

jonathanleff (commented):
Addressed in a54cc05. Responses to each high-priority item:

### H1 — Token validation fallthrough on server errors
**Fixed.** `_validate_token` now prints a visible warning when the server returns 5xx or the request fails, making it clear the token was not verified. We still store the token because server errors are transient — blocking login permanently on a 502 during a deploy would be worse UX than storing an unverified token. The warning gives the user the information to act on.

### H2 — Version mismatch (0.1.0 vs 0.11.0)
**No change needed — working as designed.** `kz-ext` is intentionally versioned separately from `kamiwaza-sdk`. The extensions CLI starts at `0.1.0` with its own release cadence. The doctor check compares `kz-ext.__version__` against `kz_ext_version` in `kamiwaza.json`, and scaffolded templates set `kz_ext_version: ">=0.1.0,<1.0.0"` — which correctly matches. Added a clarifying comment in `__init__.py`.

### H3 — URL derivation in build_env_overlay
**No change.** The `removesuffix("/api")` pattern matches the actual Kamiwaza URL convention (`https://host/api`). The concern about `https://host/api/v2` is theoretical — Kamiwaza doesn't use that pattern. `KAMIWAZA_ENDPOINT` as `{url}/v1` matches the current extensions-template behavior. Adding `urllib.parse` for this would be overengineering for the actual URL shapes in use.

### H4 — KAMIWAZA_VERIFY_SSL not in env overlay
**Fixed.** `build_env_overlay` now includes `KAMIWAZA_VERIFY_SSL=false` when the connection was created with `--no-verify-ssl`. Extensions using the SDK inside Docker will respect the SSL setting.

### H5 — Dotfiles excluded from wheels
**Fixed.** Added `templates/**/.*` pattern to `pyproject.toml` package-data so `.gitignore` files are included in wheel builds.

### Also fixed
- **M1**: Removed dead `last_error` variable in doctor
- **M2**: Removed dead quote-check logic in compose port validation

jxstanford (changes requested):
# RE-REVIEW Report: ENG-3056: Add kz-ext CLI for extension developer tools (Phase 1)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/60
**Branch**: `feat/kamiwaza-extensions` → `develop`
**Engines**: Claude Code, Codex, Kamiwaza (3/4 available)

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| | | *(none)* | | |

## High Priority Issues
| H1 | `commands/login.py:91-94` | **Pre-existing `KAMIWAZA_VERIFY_SSL=false` leaks into login when `--no-verify-ssl` is NOT passed.** The code only sets the env var when `verify_ssl=False` but never forces it to `"true"` otherwise. If a user's shell has `KAMIWAZA_VERIFY_SSL=false` and they run `kz-ext login` against a production URL, `KamiwazaClient` reads the env and silently skips certificate verification. Fix: add `else: os.environ["KAMIWAZA_VERIFY_SSL"] = "true"` so the flag always reflects the CLI argument. | Codex, Kamiwaza | Security/Correctness |
| H2 | `templates/*/docker-compose.yml` | **`KAMIWAZA_VERIFY_SSL` missing from all compose template environment sections.** `build_env_overlay` correctly sets it on the host (round 4 fix), but none of the three template compose files (`app`, `tool`, `service`) include `KAMIWAZA_VERIFY_SSL=${KAMIWAZA_VERIFY_SSL:-}`. Extensions running inside containers won't inherit the SSL override, so `--no-verify-ssl` / `--dev` won't propagate to runtime libs in containers. | Claude Code, Codex | Correctness |
| H3 | `commands/validate.py:23-25` | **`validate --json` outputs non-JSON on missing `kamiwaza.json`.** The early-exit path prints a Rich-formatted error string and calls `typer.Exit(1)` regardless of whether `--json` was passed. Callers expecting machine-readable output on every invocation get unparseable text for the most common failure mode. Wrap in `if json_output: print(json.dumps({"passed": false, "errors": [...]}))`. | Codex | Correctness |
| H4 | `commands/login.py:~140-160` | **Token stored despite 5xx server errors during validation.** `_validate_token` warns but returns `True` (stores credentials) when the server returns 5xx. While connection errors correctly reject, a 5xx could indicate a compromised or MITM'd server. Consider failing closed: reject on 5xx with a message suggesting `--force` or retry. | Kamiwaza | Security |

| M1 | `dev_local.py:139-140` | **`KAMIWAZA_ENDPOINT` derived from API URL, not public URL.** When connection URL is `https://host/api`, endpoint becomes `https://host/api/v1`. The function already derives `KAMIWAZA_PUBLIC_API_URL=https://host`, and platform model-serving endpoints are typically at `{public_url}/v1`, not `{api_url}/v1`. Consider deriving from the public URL instead. | Codex | Correctness |
| M2 | `connections.py:180-183` | **Token file permission race window.** File is created with default umask before `chmod(600)`. The parent dir is already 700, limiting exposure, but using `os.open(path, O_CREAT\|O_WRONLY, 0o600)` + `os.fdopen` would eliminate the window entirely. | Kamiwaza | Security |
| M3 | `doctor.py:217-220` | **Doctor recommends wrong package name.** Fix message suggests `pip install 'kamiwaza-extensions{spec}'` but the CLI ships inside `kamiwaza-sdk`. Users following this guidance would install a nonexistent package. | Codex | Correctness |
| M4 | `cli.py:5-6` | **Unused imports `sys` and `traceback`** — dead imports from earlier iterations. | Claude Code | Code quality |
| M5 | `cli.py:82-98` | **`_handle_exception` has identical `KamiwazaError` and generic `Exception` branches.** Both print error, optionally print traceback, and exit 1. Remove the `KamiwazaError` check or add differentiating behavior. | Claude Code | Code quality |
| M6 | `dev_local.py:139` | **`url.removesuffix("/api")` edge case.** URLs like `https://api.example.com/api` would become `https://api.example.com`, losing the subdomain semantics. Use `urllib.parse` for proper URL manipulation. | Kamiwaza | Correctness |
| M7 | `connections.py:195-200` | **Corrupt config silently returns empty dict.** If the config JSON is malformed, `_load_config` returns `{}`, effectively erasing all connections. Consider backing up the corrupt file and warning the user. | Kamiwaza | Data integrity |

## Test Coverage Assessment
**102 test methods across 9 test files** — strong for Phase 1.

- **Well covered**: `connections.py` (12 tests incl. path traversal, corrupt config, token round-trip), `validators/metadata.py` (12), `validators/compose.py` (8), `scaffolder.py` (15 incl. scaffold-then-validate), `dev_local.py` (10), `doctor.py` (10), `cli.py` (8), `validate_cmd` (5), `e2e_flows` (7), `login` (7)
- **Gaps (non-blocking for Phase 1)**:
  - `_login_with_password` SDK client chain — only partially tested via CLI happy path
  - `_validate_token` status-code branching (500 → warn+store, ConnectionError → reject) — no direct unit tests
  - `DevLocalRunner.run()` subprocess lifecycle (signal forwarding, exit code propagation)
  - H1 scenario: pre-existing `KAMIWAZA_VERIFY_SSL=false` in env during `verify_ssl=True` login

## Positive Observations
- **Identity isolation**: `_MemoryTokenStore` prevents kz-ext from corrupting shared `~/.kamiwaza/token.json` — well-designed (all engines)
- **Atomic config writes**: `_save_config` uses tmp file + `replace()` to avoid corruption (Claude Code, Kamiwaza)
- **Security hardening**: chmod 600 on tokens, chmod 700 on config dir, connection name regex against path traversal (all engines)
- **Token validation redesign**: Single auth-required endpoint (`/auth/pats`) with per-failure-mode warnings is a significant improvement over multi-endpoint ping (Claude Code)
- **Scaffold-then-validate tests**: Ensuring all three template types pass the validator after scaffolding catches template/schema drift early (Claude Code)
- **Clean architecture**: CLI layer (Typer) → command implementations → core logic — testable at each layer (Claude Code, Kamiwaza)
- **Version compatibility checking**: Proper use of `packaging.specifiers` for CLI version validation (Kamiwaza)

## Overall Verdict
**Risk Level**: Medium
**Verdict**: REQUEST_CHANGES
**Engine Agreement**: 3/4 engines contributed; 2/3 recommend changes

**Rationale**: The architecture and test coverage are strong for Phase 1, and previous review rounds addressed many issues well. However, three high-priority issues warrant another pass before merge:

- **H1** is a security logic bug: SSL verification can be silently bypassed during login if the user's shell has a pre-existing env var — a one-line fix.
- **H2** means `--dev` / `--no-verify-ssl` doesn't actually work for containerized extensions — the most common use case for `kz-ext dev local`. Adding one line to each compose template fixes it.
- **H3** breaks the `--json` contract for the most common validation failure, which CI/automation consumers will hit immediately.

All three fixes are small and surgical. H4 (token-on-5xx) is worth discussing but could be deferred to Phase 2 if the team prefers warn+store semantics for dev tooling.

---

jonathanleff (commented):
Addressed H2, H3, M3, M4 in 8f162bd. Responses to all high-priority items:

### Fixed

**H2 — KAMIWAZA_VERIFY_SSL missing from compose templates**: Added `KAMIWAZA_VERIFY_SSL=${KAMIWAZA_VERIFY_SSL:-}` to all three template compose files (app, tool, service). Containers now inherit the SSL setting from `kz-ext dev local`.

**H3 — validate --json outputs non-JSON on missing kamiwaza.json**: Early-exit path now outputs valid JSON when `--json` is passed.

### Skipped (intentional)

**H1 — Pre-existing KAMIWAZA_VERIFY_SSL=false leaks into login**: This is standard env var precedence, not a bug. If a user has `KAMIWAZA_VERIFY_SSL=false` in their shell, that's their intent — the SDK respects it everywhere. Overriding it to `"true"` when `--no-verify-ssl` isn't passed would break users who set the env var deliberately (e.g. in `.bashrc` for a dev environment). The `try/finally` in `_login_with_password` already scopes the mutation correctly for the `--no-verify-ssl` flag case.

**H4 — Token stored despite 5xx**: Already addressed in round 4 with visible warnings. Adding a `--force` flag introduces complexity for an edge case — if the server is 5xx'ing during a deploy, the user wants their creds stored so they can work once it recovers. The warning gives them the information to act on. Can revisit in Phase 2 if this causes real user confusion.

### Also fixed
- **M3**: Doctor version message now references `kamiwaza-sdk` (not nonexistent `kamiwaza-extensions`)
- **M4**: Removed unused `sys` and `traceback` imports

jxstanford (approved):
# RE-REVIEW Report: ENG-3056: Add kz-ext CLI for extension developer tools (Phase 1)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/60
**Branch**: `feat/kamiwaza-extensions` → `develop`
**Engines**: Claude Code, Kamiwaza (2/4 available)

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| | | (none) | | |

## High Priority Issues
| H1 | `commands/login.py:147-152` | **Transport failures treated as successful validation**: When `_validate_token` encounters `SSLError`, timeout, or other `RequestException` (e.g. self-signed cert without `--no-verify-ssl`), it returns `True` and stores the connection. `kz-ext login` prints success but the stored credentials are unusable by later commands (`doctor`, `dev local`). Transport-level failures should fail validation, not silently pass. | Codex, Kamiwaza | Security/Logic |
| H2 | `dev_local.py:95-117` | **Signal handler race + missing ProcessLookupError guard**: (a) Signal handlers are installed *after* `Popen` — a signal arriving in the gap won't be forwarded. (b) `proc.send_signal(signum)` at line 116 doesn't catch `ProcessLookupError`, so if the child exits between signal receipt and the call, an unhandled exception propagates in signal-handler context. | Claude Code, Kamiwaza | Reliability |

| M1 | `commands/login.py:102` | **PAT minted without audience/scope defaults**: Password-login path calls `PATCreate(name=...)` without `scope="openid"` or `aud="kamiwaza-platform"`. The SDK CLI and integration tests always set these. On deployments that enforce audience/scope mapping, the stored PAT will get 401/403 from subsequent API calls. | Codex | Logic |
| M2 | `validators/compose.py:33-35` | **Compose files with no services pass validation**: When `services` is missing or empty, `validate` returns `passed=True` with only a warning. These files cannot be started by `docker compose up`, so validation gives a false green-light. Should be an error. | Codex | Logic |
| M3 | `commands/login.py:78-92` | **Global `os.environ` mutation for SSL setting**: `_login_with_password` mutates `os.environ["KAMIWAZA_VERIFY_SSL"]` which is process-global. Low risk for a CLI (single-threaded), but if the function is ever reused in a library context this becomes a concurrency hazard. | Kamiwaza | Security |
| M4 | `pyproject.toml:21-30` | **`rich` used but not declared as explicit dependency**: `rich` is imported directly in 7+ files but isn't listed in `dependencies` — it works via `typer`'s transitive dep. Explicit declaration is safer against upstream changes. | Claude Code | Packaging |
| M5 | `validators/compose.py:75-82` | **Absolute Dockerfile path may bypass validation**: When `dockerfile` is an absolute path, `ext_dir / context / dockerfile` resolves to the absolute path on some OSes, bypassing the intended relative check. | Kamiwaza | Logic |
| M6 | `connections.py:180-183` | **Silent permission failures**: `chmod 600`/`700` silently passes on `OSError`. On restrictive filesystems, tokens could remain world-readable without warning. | Kamiwaza | Security |
| M7 | `cli.py:80-96` | **`KamiwazaError` branch produces identical output to generic branch**: The conditional adds no value currently — could be simplified or extended with error-specific handling. | Claude Code | Code quality |

## Test Coverage Assessment
- **102 unit tests** with thorough edge case coverage across all major modules
- **Well-covered**: `connections.py` (13 tests incl. path traversal, corrupt config), `validators/metadata.py` (12), `validators/compose.py` (8), `scaffolder.py` (14 incl. scaffold→validate integration), `dev_local.py` (12), `doctor.py` (11), `cli.py` (8), `commands/login.py` (6), `commands/validate.py` (5), `e2e flows` (7)
- **Gaps**: `_login_with_password` path (username/password flow not unit-tested beyond prompt abort), `_validate_token` edge cases with `verify_ssl=False` and transport errors (directly related to H1)

## Positive Observations
- Clean separation between CLI entry points (`commands/*.py`) and business logic modules
- Security fundamentals are solid: connection name allowlist regex prevents path traversal, token files `chmod 600`, atomic config writes via `tmp+rename`, `yaml.safe_load` for compose parsing
- Good use of Pydantic with `extra="allow"` for forward-compatible metadata validation
- `_MemoryTokenStore` in login avoids clobbering the shared SDK token — good isolation
- All prior review feedback (4 items) addressed cleanly in a minimal delta (8 insertions, 4 deletions)
- README documentation verified: all 5 commands, flags, and extension types match implementation

## Overall Verdict
**Risk Level**: Low
**Verdict**: APPROVE
**Engine Agreement**: 2/4 engines contributed — both approve. All engines agree on the token-validation bypass (H1) as the most impactful finding.

**Rationale**: The codebase demonstrates solid architecture, strong security practices, and comprehensive test coverage (102 tests). H1 (transport failures stored as valid credentials) and H2 (signal handler robustness) are real issues but both are edge-case scenarios in a Phase 1 CLI tool — H1 produces a misleading success message rather than a security breach, and H2 requires precise signal timing. Recommend addressing H1 and M1-M2 in a fast-follow since they affect the login and validation happy paths.

---

PR #74, opened 2026-04-19
fix(sdk): preserve server 401 detail for non-refreshable auth; fix apps session-token test

yod-kz:
## Summary

UAT surfaced a confusing error from `POST /apps/sessions/heartbeat`:

```
POST /apps/sessions/heartbeat → 401 {"detail":"Invalid session token"}
SDK raises: AuthenticationError: Authentication failed after token refresh.
```

Two issues, both on the SDK side:

1. **Test is wrong.** `tests/integration/test_apps_live.py::test_apps_deploy_and_deployment_error_paths` was sending a bogus session token in the JSON body and asserting `status_code == 404`. Server PR [#1376](https://github.com/kamiwaza-ai/kamiwaza/pull/1376) (ENG-2902, merged Mar 30 2026) intentionally removed body-based tokens from `kamiwaza/dependencies/session_auth.py`. App session endpoints now only accept tokens via `Authorization: Bearer` or `X-App-Session-Token` headers, and an unknown token correctly returns **401** "Invalid session token" at the auth boundary.

2. **SDK error message is misleading.** `ApiKeyAuthenticator.refresh_token` is a no-op (PATs can't be refreshed client-side). On 401 the SDK still called it, retried once, then raised `"Authentication failed after token refresh"` — implying a refresh was attempted. The server's actual response body was also swallowed.

## Changes

- `kamiwaza_sdk/authentication.py`: add `Authenticator.can_refresh()` (default `True`). Override to `False` on `ApiKeyAuthenticator`.
- `kamiwaza_sdk/client.py`: in the 401 branch, if the authenticator can't actually refresh, raise immediately with the endpoint + server response body preserved. Duck-typed (`callable()`-aware, attribute-style `can_refresh` also respected, no attribute means default `True` for legacy auths). For refreshable authenticators, the post-refresh error now also includes endpoint + server body.
- `tests/integration/test_apps_live.py`: assert the correct 401 auth-boundary semantics (via `AuthenticationError`) with a documenting comment about ENG-2902.
- `tests/unit/test_client_401_handling.py`: new unit coverage for all four 401 paths (PAT, refreshable, legacy duck-typed, attribute-style, no-auth).

## Six-mode analysis (standard/RBAC/ReBAC × no-workroom/in-workroom)

The server session-token validator in `kamiwaza/dependencies/session_auth.py` is **mode-agnostic**: it hashes the token and does a direct `DBAppDeployment.app_session_token_hash` lookup. No Keycloak, RBAC/ReBAC guard, or workroom scope is applied before the 401. All six auth × workroom combinations therefore behave identically on this negative path — an unknown token is 401 everywhere. The SDK change is purely in how the client reacts to the 401, so it is also mode-agnostic.

A real post-teardown heartbeat also correctly returns 401 because the deployment's token hash has been removed. That's the intended auth-boundary signal, not a 404 "deployment gone" — the caller can't prove it owned the now-missing deployment without a valid token. Not changing that semantics here.

## Council consults

Two model-council runs (Codex + Gemini in parallel).

1. **Root cause** — confirmed SDK test is wrong (post-#1376 contract), confirmed 401 is right for all six modes, called out the auth-header precedence footgun: the SDK always attaches `Authorization: Bearer <PAT>`, so even if `X-App-Session-Token` is set the server reads the PAT first and still 401s (same outcome for an unknown-token negative test).
2. **Fix review** — confirmed minimal correct fix, flagged (and I addressed) a `callable()` edge case in the duck-typed `can_refresh` resolution and tightened tests to assert endpoint + server body in all error messages.

## Test plan

- [x] `uv run pytest -m "not integration and not live and not e2e"` → 705 passed, 8 skipped, 172 deselected (up from 703 baseline; +2 new tests, no regressions).
- [x] `uv run pytest tests/unit/test_client_401_handling.py -v` → 5 passed.
- [ ] Live regression: re-run `tests/integration/test_apps_live.py::test_apps_deploy_and_deployment_error_paths` against the UAT cluster (will verify in next kzuat iteration).
- [ ] Optional broader live run to confirm no other PAT-authenticated endpoint started relying on the retry-then-fail path.

## Out of scope / follow-ups

- Server-side: no change needed. The 401 is the correct boundary response for unknown session tokens across all modes.
- Potential future improvement: a `KamiwazaClient.post(..., skip_auth=True)` (or similar) convenience so session-token flows can opt out of attaching the PAT, letting the server's `X-App-Session-Token` path be exercised without the PAT preceding it. Not required for this fix.

---

Tagging @m9e for the five-agent multi-review.

### DIFF
--- a/kamiwaza_sdk/authentication.py
+++ b/kamiwaza_sdk/authentication.py
@@ -6,7 +6,7 @@ import logging
 from abc import ABC, abstractmethod
 from datetime import datetime, timedelta, timezone
 import time
-from typing import Optional, TYPE_CHECKING
+from typing import Optional
 
 import requests
 
--- a/kamiwaza_sdk/authentication.py
+++ b/kamiwaza_sdk/authentication.py
@@ -182,3 +182,13 @@ class OAuthAuthenticator(Authenticator):
 
     def refresh_token(self, session: requests.Session) -> None:  # pragma: no cover - not implemented
         raise NotImplementedError("OAuth authentication is not yet implemented.")
+
+    def can_refresh(self) -> bool:
+        """Placeholder: OAuth refresh is not yet implemented.
+
+        Returning ``False`` prevents the HTTP 401 handler from invoking
+        ``refresh_token`` (which would raise ``NotImplementedError``) and
+        lets the client surface a clean ``AuthenticationError`` instead.
+        Flip this back to the default ``True`` once OAuth refresh lands.
+        """
+        return False
--- a/kamiwaza_sdk/authentication.py
+++ b/kamiwaza_sdk/authentication.py
@@ -33,6 +33,18 @@ class Authenticator(ABC):
         """Return the active access token, refreshing when necessary."""
         return None
 
+    def can_refresh(self) -> bool:
+        """Whether ``refresh_token`` can actually obtain new credentials.
+
+        Defaults to ``True`` for strategies that exchange credentials for
+        short-lived tokens (password/OAuth). ``ApiKeyAuthenticator`` overrides
+        this to ``False`` because a PAT/API-key cannot be refreshed client-side.
+        The HTTP layer uses this to avoid a pointless retry on 401 and to
+        surface the server's actual response instead of a misleading
+        "after token refresh" message.
+        """
+        return True
+
 
 class ApiKeyAuthenticator(Authenticator):
     """Simple bearer token authenticator backed by a PAT/API key."""
@@ -45,6 +57,10 @@ class ApiKeyAuthenticator(Authenticator):
 
     def refresh_token(self, session: requests.Session) -> None:  # pragma: no cover - nothing to refresh
         pass
+
+    def can_refresh(self) -> bool:
+        """PATs/API keys cannot be refreshed client-side."""
+        return False
 
     def get_access_token(self, session: requests.Session) -> Optional[str]:
         return self.api_key
--- a/kamiwaza_sdk/client.py
+++ b/kamiwaza_sdk/client.py
@@ -36,6 +36,21 @@ from .services.skills import SkillsService
 logger = logging.getLogger(__name__)
 
 _AUTH_ERROR_DETAIL_MAX_LEN = 500
+_AUTH_ERROR_DETAIL_TRUNCATED_SUFFIX = "... [truncated]"
+
+
+def _truncate_with_suffix(
+    value: str, max_len: int = _AUTH_ERROR_DETAIL_MAX_LEN
+) -> str:
+    """Truncate ``value`` to ``max_len`` chars, appending a suffix when cut.
+
+    A naked slice is ambiguous — a 500-char return is indistinguishable from
+    a legitimately short body that happens to fit. The suffix makes the
+    truncation explicit to anyone reading logs or exception messages.
+    """
+    if len(value) <= max_len:
+        return value
+    return value[:max_len] + _AUTH_ERROR_DETAIL_TRUNCATED_SUFFIX
 
 
 def _extract_server_detail(response, max_len: int = _AUTH_ERROR_DETAIL_MAX_LEN) -> str:
@@ -43,21 +58,22 @@ def _extract_server_detail(response, max_len: int = _AUTH_ERROR_DETAIL_MAX_LEN) 
 
     Prefers the JSON ``detail`` field (FastAPI convention) so the caller sees
     the server's actual message. Falls back to the serialized JSON body, then
-    raw text. Output is always truncated to ``max_len`` characters to prevent
-    multi-KB proxy/gateway HTML error pages from bloating log lines and
-    exception strings.
+    raw text. Output is always truncated to ``max_len`` characters (with an
+    explicit ``... [truncated]`` suffix when cut) to prevent multi-KB
+    proxy/gateway HTML error pages from bloating log lines and exception
+    strings.
     """
     try:
         body = response.json()
     except (ValueError, AttributeError):
-        return (response.text or "")[:max_len]
+        return _truncate_with_suffix(response.text or "", max_len)
 
     if isinstance(body, dict) and "detail" in body:
         detail = body["detail"]
         if isinstance(detail, str):
-            return detail[:max_len]
-        return str(detail)[:max_len]
-    return str(body)[:max_len]
+            return _truncate_with_suffix(detail, max_len)
+        return _truncate_with_suffix(str(detail), max_len)
+    return _truncate_with_suffix(str(body), max_len)
 
 
 class KamiwazaClient:
--- a/kamiwaza_sdk/client.py
+++ b/kamiwaza_sdk/client.py
@@ -210,7 +210,8 @@ class KamiwazaClient:
                         f"{_extract_server_detail(response)}"
                     )
                 logger.warning(
-                    f"Received 401 Unauthorized. Response: {response.text}"
+                    f"Received 401 Unauthorized. Response: "
+                    f"{_extract_server_detail(response)}"
                 )
                 if self.authenticator:
                     # Only attempt a refresh-and-retry if the authenticator can
--- a/kamiwaza_sdk/client.py
+++ b/kamiwaza_sdk/client.py
@@ -35,6 +35,30 @@ from .services.skills import SkillsService
 
 logger = logging.getLogger(__name__)
 
+_AUTH_ERROR_DETAIL_MAX_LEN = 500
+
+
+def _extract_server_detail(response, max_len: int = _AUTH_ERROR_DETAIL_MAX_LEN) -> str:
+    """Extract a short, embeddable description of a server error response.
+
+    Prefers the JSON ``detail`` field (FastAPI convention) so the caller sees
+    the server's actual message. Falls back to the serialized JSON body, then
+    raw text. Output is always truncated to ``max_len`` characters to prevent
+    multi-KB proxy/gateway HTML error pages from bloating log lines and
+    exception strings.
+    """
+    try:
+        body = response.json()
+    except (ValueError, AttributeError):
+        return (response.text or "")[:max_len]
+
+    if isinstance(body, dict) and "detail" in body:
+        detail = body["detail"]
+        if isinstance(detail, str):
+            return detail[:max_len]
+        return str(detail)[:max_len]
+    return str(body)[:max_len]
+
 
 class KamiwazaClient:
     _RECENT_DATASET_TTL_SECONDS = 30.0
@@ -182,9 +206,12 @@ class KamiwazaClient:
             if response.status_code == 401:
                 if skip_auth:
                     raise AuthenticationError(
-                        f"Unauthenticated request failed for {endpoint}: {response.text}"
+                        f"Unauthenticated request failed for {endpoint}: "
+                        f"{_extract_server_detail(response)}"
                     )
-                logger.warning(f"Received 401 Unauthorized. Response: {response.text}")
+                logger.warning(
+                    f"Received 401 Unauthorized. Response: {response.text}"
+                )
                 if self.authenticator:
                     # Only attempt a refresh-and-retry if the authenticator can
                     # actually obtain new credentials. PAT/API-key auth cannot,
@@ -207,7 +234,7 @@ class KamiwazaClient:
                     if did_refresh:
                         raise AuthenticationError(
                             f"Authentication failed after token refresh for "
-                            f"{endpoint}: {response.text}"
+                            f"{endpoint}: {_extract_server_detail(response)}"
                         )
                     # Non-refreshable authenticator (e.g. PAT): surface the
                     # server's actual response so callers can distinguish
@@ -215,7 +242,8 @@ class KamiwazaClient:
                     # (e.g. app session-token endpoints returning 401 for an
                     # unknown session token).
                     raise AuthenticationError(
-                        f"Authentication failed for {endpoint}: {response.text}"
+                        f"Authentication failed for {endpoint}: "
+                        f"{_extract_server_detail(response)}"
                     )
                 raise AuthenticationError(
                     "Authentication failed. No authenticator provided."
--- a/kamiwaza_sdk/client.py
+++ b/kamiwaza_sdk/client.py
@@ -186,12 +186,36 @@ class KamiwazaClient:
                     )
                 logger.warning(f"Received 401 Unauthorized. Response: {response.text}")
                 if self.authenticator:
-                    if not did_refresh:
+                    # Only attempt a refresh-and-retry if the authenticator can
+                    # actually obtain new credentials. PAT/API-key auth cannot,
+                    # so a retry is wasted and the "after token refresh" error
+                    # message is misleading when it fails. Duck-typed: accept
+                    # either a callable method or a plain attribute/property;
+                    # default to True for legacy authenticators that don't
+                    # expose the hook at all.
+                    can_refresh_attr = getattr(
+                        self.authenticator, "can_refresh", True
+                    )
+                    can_refresh = (
+                        can_refresh_attr() if callable(can_refresh_attr)
+                        else bool(can_refresh_attr)
+                    )
+                    if can_refresh and not did_refresh:
                         did_refresh = True
                         self.authenticator.refresh_token(self.session)
                         continue
+                    if did_refresh:
+                        raise AuthenticationError(
+                            f"Authentication failed after token refresh for "
+                            f"{endpoint}: {response.text}"
+                        )
+                    # Non-refreshable authenticator (e.g. PAT): surface the
+                    # server's actual response so callers can distinguish
+                    # invalid PAT from endpoint-specific auth boundary errors
+                    # (e.g. app session-token endpoints returning 401 for an
+                    # unknown session token).
                     raise AuthenticationError(
-                        "Authentication failed after token refresh."
+                        f"Authentication failed for {endpoint}: {response.text}"
                     )
                 raise AuthenticationError(
                     "Authentication failed. No authenticator provided."
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -806,6 +806,25 @@ def _password_auth_works(
         return False, str(exc)
 
 
+def _api_key_auth_works(base_url: str, api_key: str) -> tuple[bool, str]:
+    """Validate a PAT/API key by calling /auth/users/me.
+
+    Stale PATs left over from a prior platform install (e.g. ``.env.local`` from
+    before Keycloak's signing keys were rotated during a reinstall) will parse
+    as valid JWTs but fail verification with 401 ``Not authenticated`` because
+    the signing ``kid`` is unknown to the new platform. Probing /auth/users/me
+    lets the fixture chain detect that case and fall through to password-based
+    PAT creation instead of handing every test a dead token.
+    """
+
+    client = KamiwazaClient(base_url, api_key=api_key)
+    try:
+        client.auth.get_current_user()
+        return True, ""
+    except (AuthenticationError, APIError) as exc:
+        return False, str(exc)
+
+
 def _resolve_live_password_once(
     *,
     live_server_available: str,
@@ -1179,12 +1198,29 @@ def live_session_api_key(
 
     Auth-specific tests still use live_username/live_password directly, but the
     shared client fixture should not re-run password grants across the whole suite.
+
+    Env-supplied ``KAMIWAZA_API_KEY`` values are validated against
+    ``/auth/users/me`` before use. Stale PATs left behind by a prior platform
+    install (same ``.env.local``, new Keycloak signing keys after an
+    ``install-dev.sh`` reinstall) would otherwise poison every test with the
+    generic ``AuthenticationError: Authentication failed after token refresh``
+    signature. If the env PAT fails probe, fall through to password-based PAT
+    creation instead.
     """
 
     api_key = live_api_key.strip()
     if api_key:
-        yield api_key
-        return
+        ok, error = _api_key_auth_works(live_server_available, api_key)
+        if ok:
+            yield api_key
+            return
+        # Stale PAT (e.g. signed by a pre-reinstall Keycloak key). Fall through
+        # to password-based PAT creation so the rest of the suite can run.
+        print(
+            f"\n[live_session_api_key] env KAMIWAZA_API_KEY rejected by platform "
+            f"({error}); falling back to password-based PAT creation.",
+            flush=True,
+        )
 
     username = live_username.strip()
     password = resolved_live_password.strip()
@@ -1232,7 +1268,10 @@ def live_session_write_key(
     starts requiring admin, these tests will surface the regression as a 403.
     """
 
-    if live_api_key.strip():
+    # Only skip when the env PAT is usable — otherwise a stale env PAT would
+    # also prevent the write-scope regression guard from running.
+    api_key = live_api_key.strip()
+    if api_key and _api_key_auth_works(live_server_available, api_key)[0]:
         yield ""
         return
 
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1160,7 +1160,17 @@ def resolved_live_password(
     """
     Resolve live password with kube-derived credentials first (kz-login),
     then explicit configured password as fallback.
+
+    Fast-path: when ``KAMIWAZA_API_KEY`` is present and validates against
+    ``/auth/users/me``, skip the password resolver entirely. The valid PAT
+    is the only credential the integration suite needs, and running
+    ``kz-login`` on every session just to fill an unused string can trip
+    Keycloak rate limits / account lockout across many test runs.
     """
+
+    env_api_key = live_api_key.strip()
+    if env_api_key and _api_key_auth_works(live_server_available, env_api_key)[0]:
+        return ""
 
     password, error = _resolve_live_password_once(
         live_server_available=live_server_available,
@@ -1168,7 +1178,7 @@ def resolved_live_password(
         live_username=live_username,
         configured_password=str(pytestconfig.getoption("live_password")),
     )
-    if password or live_api_key.strip():
+    if password or env_api_key:
         return password
 
     username = live_username.strip()
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -831,11 +831,6 @@ def _resolve_live_password_once(
     if cached is not None:
         return cached
 
-    if live_api_key.strip():
-        result = ("", None)
-        _LIVE_PASSWORD_CACHE[cache_key] = result
-        return result
-
     username = live_username.strip()
     configured_password = configured_password.strip()
     if not username:
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -69,6 +69,14 @@ RUNTIME_HOST_ALIAS = os.environ.get(
 _COMPOSE_ENV_CACHE: dict[str, str] | None = None
 _COMPOSE_ENV_ERROR: str | None = None
 _LIVE_PASSWORD_CACHE: dict[tuple[str, str, str, str], tuple[str, str | None]] = {}
+# Memoize PAT probe (``GET /auth/users/me``) results for the session.
+# ``_api_key_auth_works`` is called once in ``resolved_live_password``,
+# once in ``live_session_api_key``, and once in ``live_session_write_key``
+# (plus any future fixture that needs to validate a PAT). Without caching,
+# the same PAT gets probed against Keycloak three times per session, which
+# is noise at best and a lockout vector at worst. Cache both success and
+# failure so a bad PAT doesn't keep re-probing either.
+_API_KEY_PROBE_CACHE: dict[tuple[str, str], tuple[bool, str]] = {}
 _HTTP_TRACE_FLAG = "KAMIWAZA_HTTP_TRACE"
 _HTTP_TRACE_FILE_ENV = "KAMIWAZA_HTTP_TRACE_FILE"
 _TEXT_BODY_MARKERS = (
@@ -815,14 +823,26 @@ def _api_key_auth_works(base_url: str, api_key: str) -> tuple[bool, str]:
     the signing ``kid`` is unknown to the new platform. Probing /auth/users/me
     lets the fixture chain detect that case and fall through to password-based
     PAT creation instead of handing every test a dead token.
+
+    Result (success or failure) is memoized on ``(base_url, api_key)`` for the
+    session, symmetric with ``_LIVE_PASSWORD_CACHE``. Multiple fixtures probe
+    the same PAT per session and we don't want to hammer Keycloak for it.
     """
+
+    cache_key = (base_url, api_key)
+    cached = _API_KEY_PROBE_CACHE.get(cache_key)
+    if cached is not None:
+        return cached
 
     client = KamiwazaClient(base_url, api_key=api_key)
     try:
         client.auth.get_current_user()
-        return True, ""
+        result: tuple[bool, str] = (True, "")
     except (AuthenticationError, APIError) as exc:
-        return False, str(exc)
+        result = (False, str(exc))
+
+    _API_KEY_PROBE_CACHE[cache_key] = result
+    return result
 
 
 def _resolve_live_password_once(
@@ -1161,17 +1181,16 @@ def resolved_live_password(
     Resolve live password with kube-derived credentials first (kz-login),
     then explicit configured password as fallback.
 
-    Fast-path: when ``KAMIWAZA_API_KEY`` is present and validates against
-    ``/auth/users/me``, skip the password resolver entirely. The valid PAT
-    is the only credential the integration suite needs, and running
-    ``kz-login`` on every session just to fill an unused string can trip
-    Keycloak rate limits / account lockout across many test runs.
+    Session-scoped and backed by ``_LIVE_PASSWORD_CACHE`` inside
+    ``_resolve_live_password_once`` so kz-login / password grants run at most
+    once per session. Password-authentication tests
+    (``test_password_authentication_allows_whoami``, PAT-lifecycle, CLI login)
+    consume this fixture directly, so it must always resolve a real password
+    when one is available — returning an empty short-circuit string here
+    regresses those tests.
     """
 
     env_api_key = live_api_key.strip()
-    if env_api_key and _api_key_auth_works(live_server_available, env_api_key)[0]:
-        return ""
-
     password, error = _resolve_live_password_once(
         live_server_available=live_server_available,
         live_api_key=live_api_key,
@@ -1226,9 +1245,13 @@ def live_session_api_key(
             return
         # Stale PAT (e.g. signed by a pre-reinstall Keycloak key). Fall through
         # to password-based PAT creation so the rest of the suite can run.
+        # Cap the error text so a gateway 502/503 HTML body (or any other
+        # verbose non-401 failure surfaced via ``APIError.__str__``) doesn't
+        # flood CI logs / pytest output.
+        truncated_error = error if len(error) <= 500 else error[:500] + "... [truncated]"
         print(
             f"\n[live_session_api_key] env KAMIWAZA_API_KEY rejected by platform "
-            f"({error}); falling back to password-based PAT creation.",
+            f"({truncated_error}); falling back to password-based PAT creation.",
             flush=True,
         )
 
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -70,12 +70,12 @@ _COMPOSE_ENV_CACHE: dict[str, str] | None = None
 _COMPOSE_ENV_ERROR: str | None = None
 _LIVE_PASSWORD_CACHE: dict[tuple[str, str, str, str], tuple[str, str | None]] = {}
 # Memoize PAT probe (``GET /auth/users/me``) results for the session.
-# ``_api_key_auth_works`` is called once in ``resolved_live_password``,
-# once in ``live_session_api_key``, and once in ``live_session_write_key``
-# (plus any future fixture that needs to validate a PAT). Without caching,
-# the same PAT gets probed against Keycloak three times per session, which
-# is noise at best and a lockout vector at worst. Cache both success and
-# failure so a bad PAT doesn't keep re-probing either.
+# ``_api_key_auth_works`` is called once in ``live_session_api_key`` and
+# once in ``live_session_write_key`` (plus any future fixture that needs
+# to validate a PAT). Without caching, the same PAT gets probed against
+# Keycloak on every call, which is noise at best and a lockout vector at
+# worst. Cache both success and failure so a bad PAT doesn't keep
+# re-probing either.
 _API_KEY_PROBE_CACHE: dict[tuple[str, str], tuple[bool, str]] = {}
 _HTTP_TRACE_FLAG = "KAMIWAZA_HTTP_TRACE"
 _HTTP_TRACE_FILE_ENV = "KAMIWAZA_HTTP_TRACE_FILE"
--- a/tests/integration/test_apps_live.py
+++ b/tests/integration/test_apps_live.py
@@ -5,6 +5,7 @@ from uuid import uuid4
 
 import pytest
 
+from kamiwaza_sdk import KamiwazaClient
 from kamiwaza_sdk.exceptions import APIError, AuthenticationError
 
 pytestmark = [pytest.mark.integration, pytest.mark.live, pytest.mark.withoutresponses]
@@ -71,7 +72,9 @@ def test_apps_template_crud_and_images(live_kamiwaza_client) -> None:
                 pass
 
 
-def test_apps_deploy_and_deployment_error_paths(live_kamiwaza_client) -> None:
+def test_apps_deploy_and_deployment_error_paths(
+    live_kamiwaza_client, live_server_available
+) -> None:
     client = live_kamiwaza_client
     bogus_id = uuid4()
 
@@ -115,17 +118,20 @@ def test_apps_deploy_and_deployment_error_paths(live_kamiwaza_client) -> None:
     # App session endpoints use a dedicated session-token auth dependency
     # (server PR #1376, ENG-2902). Tokens must be supplied via the
     # ``Authorization: Bearer`` or ``X-App-Session-Token`` header and must
-    # match a deployment's hashed token in the database. A bogus token is
-    # rejected at the auth boundary with 401 "Invalid session token" — the
-    # SDK surfaces this as ``AuthenticationError`` because the client's PAT
-    # authenticator cannot recover (no-op refresh). This is a negative
-    # auth-boundary test: no deployment is involved, so 404 is not reachable.
+    # match a deployment's hashed token in the database. Server precedence
+    # validates ``Authorization: Bearer`` *first*, so if we sent this request
+    # with the test's PAT attached the 401 would be for "no deployment
+    # matches the PAT owner" rather than a true session-token rejection —
+    # a false positive. Build a throwaway client with **no authenticator**
+    # so only the ``X-App-Session-Token`` header drives the auth decision.
     bogus_session_token = f"sdk-session-{uuid4().hex}"
     session_headers = {"X-App-Session-Token": bogus_session_token}
+    anon_client = KamiwazaClient(live_server_available)
+    anon_client.authenticator = None  # defeat env-PAT fallback
     with pytest.raises(AuthenticationError):
-        client.post("/apps/sessions/heartbeat", headers=session_headers)
+        anon_client.post("/apps/sessions/heartbeat", headers=session_headers)
     with pytest.raises(AuthenticationError):
-        client.post(
+        anon_client.post(
             "/apps/sessions/end",
             headers=session_headers,
             json={"reason": "sdk-test"},
--- a/tests/integration/test_apps_live.py
+++ b/tests/integration/test_apps_live.py
@@ -128,9 +128,16 @@ def test_apps_deploy_and_deployment_error_paths(
     session_headers = {"X-App-Session-Token": bogus_session_token}
     anon_client = KamiwazaClient(live_server_available)
     anon_client.authenticator = None  # defeat env-PAT fallback
-    with pytest.raises(AuthenticationError):
+    # Match on the server's session-token rejection detail
+    # (``_extract_session_token`` raises 401 with ``"Invalid session token"``).
+    # A pure-anon 401 produced by the SDK itself says "No authenticator
+    # provided.", so matching on "session token" confirms the request
+    # actually reached the server's session-token dependency rather than
+    # being short-circuited anywhere else in the stack.
+    session_token_pattern = r"(?i)session token"
+    with pytest.raises(AuthenticationError, match=session_token_pattern):
         anon_client.post("/apps/sessions/heartbeat", headers=session_headers)
-    with pytest.raises(AuthenticationError):
+    with pytest.raises(AuthenticationError, match=session_token_pattern):
         anon_client.post(
             "/apps/sessions/end",
             headers=session_headers,
--- a/tests/integration/test_apps_live.py
+++ b/tests/integration/test_apps_live.py
@@ -5,7 +5,7 @@ from uuid import uuid4
 
 import pytest
 
-from kamiwaza_sdk.exceptions import APIError
+from kamiwaza_sdk.exceptions import APIError, AuthenticationError
 
 pytestmark = [pytest.mark.integration, pytest.mark.live, pytest.mark.withoutresponses]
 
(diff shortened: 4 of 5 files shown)


m9e (commented):
# PR Review Report: fix(sdk): preserve server 401 detail for non-refreshable auth; fix apps session-token test

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/74
**Branch**: fix/uat-apps-heartbeat-session-token-401 -> release/0.12.1
**Scope**: 5 files, +327/-19

**Reviewers run**: Superpowers (SP), Claude (CL), Codex (CX). Kamiwaza (KZ) errored (no streamed content). Gemini (GM) unavailable (CLI not installed). Effective N = 3.

---

_None. No reviewer flagged a critical-severity blocker._

| 1 | `tests/integration/test_apps_live.py:123-132` | The new negative-path test sets `X-App-Session-Token` in per-request headers, but `_request` calls `authenticator.authenticate(session)` first, so the session already carries `Authorization: Bearer <PAT>`. The outgoing request sends **both** headers. Depending on server precedence (PR #1376), the PAT path may be validated first and return 401 for the wrong reason, or the test becomes a false positive. Use an unauthenticated client, explicitly override `Authorization`, or at minimum document the precedence assumption. | SP, CL, CX (3/3) | Test Correctness |
| 2 | `tests/integration/conftest.py:849-853` (approx, `live_kamiwaza_client` / `live_session_api_key`) | Removing the `live_api_key` short-circuit means `_resolve_live_password_once()` / `kz-login` now runs every session even when a valid `KAMIWAZA_API_KEY` is present. If `KAMIWAZA_PASSWORD` is stale, this adds an unnecessary failed login on every test session and can trip auth lockout / rate-limit policies. Reinstate the API-key fast-path. | CX (single-source, high-confidence side-effect regression) | Test Infra / Correctness |
| 3 | `kamiwaza_sdk/client.py:208-219` | Raised `AuthenticationError` now embeds `response.text` verbatim. Typical 401s are short JSON, but gateway/HTML error pages or large proxy responses would produce huge, noisy exception strings (and could reflect input on some flows). Consider `response.text[:500]`, or parse `detail` from JSON first with fallback. | SP (Medium-High), CL (Medium, 55) — 2/3 | Defensive / Ergonomics |

| 1 | `kamiwaza_sdk/authentication.py` (OAuthAuthenticator) | `OAuthAuthenticator` inherits the default `can_refresh() == True` but its `refresh_token()` raises `NotImplementedError`. A 401 on an OAuth flow will attempt refresh and crash with `NotImplementedError` rather than a clean `AuthenticationError` — the exact failure mode this PR tries to eliminate. Pre-existing, but a natural follow-up: override `can_refresh()` to `False` until OAuth refresh is implemented. | CL (75) | Consistency / Correctness |
| 2 | `kamiwaza_sdk/client.py:196-202` | Duck-typing `can_refresh` via `getattr(..., True)` + `callable()` is defensive for non-ABC authenticators. Since `Authenticator` is an `ABC` with a default implementation, the complexity is only for external subclasses that sidestep the ABC. Could be simplified to `getattr(self.authenticator, "can_refresh", lambda: True)()` if the external-subclass extension point isn't load-bearing. | SP (Low), CL (35) — 2/3 | Simplicity / YAGNI |
| 3 | `tests/integration/test_apps_live.py:125-132` | Test asserts `AuthenticationError` tightly coupled to current server auth boundary. If server moves to 403/404 for unknown session tokens later, it regresses. Consider `pytest.raises((AuthenticationError, APIError))` and branch on `exc.value.status_code`. | CL (55) | Test Robustness |
| 4 | `tests/unit/test_client_401_handling.py` | The four 401 scenarios share substantial scaffolding (`_StubResponse`, `_client_with_responses`). Parametrizing into a single `pytest.mark.parametrize` matrix would reduce duplication. Also consider adding a test that `Authenticator.can_refresh()` default returns `True` to pin the ABC contract. | SP (Low), CL (40) | Test Quality |
| 5 | `tests/unit/test_client_401_handling.py:60-67` | `_StubResponse.json()` raises `ValueError`; modern `requests.Response.json()` raises `requests.exceptions.JSONDecodeError` (a `ValueError` subclass). Match the real exception type to avoid divergence if the client ever catches `JSONDecodeError` specifically. | SP (Low) | Test Fidelity |
| 6 | `tests/integration/conftest.py:1198-1223` | `_api_key_auth_works` is called twice (once each for `live_session_api_key` and `live_session_write_key`), each instantiating a fresh client and hitting `/auth/users/me`. Cache the probe result analogous to `_LIVE_PASSWORD_CACHE` to avoid duplicated network calls. | SP (Low) | Performance (tests) |
| 7 | `tests/integration/conftest.py:1268-1274` | `live_session_write_key` now silently falls through when the env PAT is stale; add the same log line the sibling fixture has to aid diagnosis when the downstream password fallback also fails. | SP (Low) | Diagnostics |
| 8 | `kamiwaza_sdk/authentication.py:36-46` | `can_refresh` is a new public extension hook — document it in a CHANGELOG / 0.12.1 release notes entry and in the class docstring (not just the method's) so downstream authenticator authors know to override when applicable. | SP (Low) | API Design / Docs |

## Test Coverage Assessment

- **Coverage verdict**: Adequate
- **Covered**: All four 401 paths (PAT / refreshable / legacy no-hook / attribute-style / no-authenticator) and regression guard that `"after token refresh"` is absent from the PAT-path message.
- **Missing / suggested**:
  - Direct test of `Authenticator.can_refresh()` ABC default returning `True`.
  - Guard test for `OAuthAuthenticator.can_refresh()` (would currently assert `True`, which surfaces the follow-up noted above).
  - Test that very large / HTML `response.text` produces a sensibly-bounded error message (only relevant if truncation is added per High #3).
- The unchanged `_request` path outside the 401 branch is not re-exercised, which is fine for this change's scope.

## Reviewer Agreement Matrix

| Category | Superpowers | Claude | Codex | Consensus |
|----------|-------------|--------|-------|-----------|
| Dual-header / test correctness | High | Medium-High | High (P3) | **Agreed (3/3), High** |
| `response.text` size / leakage | Medium-High | Medium | — | 2/3, Medium-High |
| Conftest password-probe regression | — | — | High (P3) | 1/3, High single-source |
| OAuthAuthenticator inconsistency | — | Medium (75) | — | 1/3, Medium |
| Duck-typing complexity | Low | Low | — | 2/3, Low |
| Test polish / parametrization | Low | Low | — | 2/3, Low |
| Security (OWASP, injection, auth) | Clean | Clean | Clean | **Agreed: no vulnerabilities** |
| Positive: error message improvement | Yes | Yes | (implicit) | **Agreed** |

- Surgical, well-motivated fix: the misleading "Authentication failed after token refresh" is replaced with endpoint + server body for non-refreshable auth, while preserving retry-on-refresh for genuinely-refreshable authenticators.
- Clear docstrings on `can_refresh()` explaining both semantics and the HTTP-layer implication, plus inline rationale in the `_request` branch.
- Strong backward-compatibility posture: `getattr(..., "can_refresh", True)` + `callable()` preserves behavior for legacy/third-party authenticators, with an explicit regression test.
- Thorough unit coverage across all four 401 code paths including an explicit assertion that `"after token refresh"` is **not** in the PAT-path message.
- Integration test comment cites server PR #1376 / ENG-2902 so future debuggers can reason about why 404 is no longer reachable.
- Conftest addition validates `KAMIWAZA_API_KEY` via `/auth/users/me` before use — addresses a real operational footgun (stale PATs after Keycloak reinstall).
- Endpoint is now included in both the post-refresh and non-refreshable error messages — meaningful debuggability improvement.
- No secrets/PII introduced into logs beyond pre-existing `logger.warning` behavior.

## Overall Verdict

**Risk Level**: Low
**Recommendation**: Request Changes (on the strength of the three high-priority items, any one of which would be sufficient by itself)
**Reviewer Consensus**: 3/3 flag the dual-header test issue. The fix itself is sound and well-tested; the blocking concerns are in the test harness (integration test dual-header, conftest password-probe regression) plus the `response.text` unbounded-embed ergonomics.

### Recommended next steps
1. Address dual-header in the new integration test (use an auth-less client or clear `Authorization` before sending `X-App-Session-Token`), or add an inline comment pinning the server-precedence assumption.
2. Restore the `live_api_key` short-circuit in `conftest.py` so `_resolve_live_password_once()` is not called when an API key is present.
3. Truncate / structure `response.text` in the new `AuthenticationError` messages (parse `detail` from JSON when possible; cap raw fallback at ~500 chars).
4. (Follow-up, optional) Override `can_refresh()` to `False` on `OAuthAuthenticator` until OAuth refresh is actually implemented.

---

Legend: SP=Superpowers, CL=Claude, CX=Codex, KZ=Kamiwaza (errored), GM=Gemini (unavailable)

*Review synthesized from 3 of 5 possible reviewers. Kamiwaza model returned "No review content in streamed response"; Gemini CLI not installed locally.*

yod-kz:
Addressed review findings on ec7d1d8f9d7f6008d9eed817892f318772415f4f.

**Landed (3 high + 1 medium):**

- **HIGH #1 — dual-header false positive in session-token test.** `test_apps_deploy_and_deployment_error_paths` now builds a throwaway `KamiwazaClient` with `authenticator = None` for the two `X-App-Session-Token` assertions so only the session-token header drives the 401. Eliminates the "valid PAT with no matching deployment" 401 that was masquerading as a session-token rejection.
- **HIGH #2 — `live_api_key` short-circuit regression.** `resolved_live_password` in `tests/integration/conftest.py` now short-circuits to `""` when the env PAT validates against `/auth/users/me`. The PAT-probe safety net added in the original PR (for stale-Keycloak-key PATs) is preserved; we just no longer run `kz-login` / password-grant when the PAT is already good, avoiding Keycloak lockout exposure on heavy test sessions.
- **HIGH #3 — unbounded `response.text` in `AuthenticationError`.** New `_extract_server_detail(response, max_len=500)` helper prefers JSON `detail`, falls back to serialized body, then raw text, always truncated. Applied to all three 401 raise sites (`skip_auth`, `did_refresh`, non-refreshable). Two new unit tests cover JSON-detail surfacing and non-JSON truncation.
- **MEDIUM — `OAuthAuthenticator.can_refresh()`.** Returns `False` until OAuth refresh is implemented; prevents a 401 from invoking the placeholder `refresh_token()` and raising `NotImplementedError` instead of a clean `AuthenticationError`. Pinned with a unit test.

**Verification:** `uv run pytest -m "not integration and not live and not e2e"` — 708 passed, 8 skipped (up from 705, zero regressions). Scope kept surgical; no unrelated files touched.

**Deferred by design (per review plan):** parametrizing unit tests; caching the `_api_key_auth_works` probe; CHANGELOG entry; any `AuthenticationError` → `APIError` swap; `JSONDecodeError` vs `ValueError` tweaks. All low-priority polish that can be addressed separately if desired.

m9e (commented):
# PR Re-Review Report: fix(sdk): preserve server 401 detail for non-refreshable auth; fix apps session-token test

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/74
**Branch**: fix/uat-apps-heartbeat-session-token-401 -> release/0.12.1
**Scope**: 5 files, +458/-23
**Prior review**: 2026-04-19 20:49 UTC (3 HIGH + mediums)
**Follow-up commit claimed**: `ec7d1d8`

**Reviewers run**: Superpowers (SP), Claude (CL), Codex (CX). Kamiwaza (KZ) errored (no streamed content, 2 attempts). Gemini (GM) unavailable (CLI not installed). Effective N = 3.

This report focuses on the **delta** since the prior review — whether the author's follow-up commit resolves prior feedback, and any new issues introduced. Unchanged code is not re-reviewed.

---

## Prior Feedback Resolution Summary

| Prior Item | SP | CL | CX | Consensus |
|---|---|---|---|---|
| HIGH #1 — dual-header in `test_apps_live.py` | Resolved | Resolved (structure), but loose assertion | Partially (loose assertion = still papers over) | **Structurally resolved; assertion still weak** |
| HIGH #2 — `live_api_key` short-circuit in `conftest.py` | **Partially — regression re-introduced** | **Partial / regression re-introduced** | Fixed (surface) | **Re-introduces prior regression** (2/3 strongly dissent; see below) |
| HIGH #3 — unbounded `response.text` in `AuthenticationError` | Resolved (w/ logger.warning residue) | Resolved (w/ logger.warning residue) | Fixed | **Resolved; one minor log-hygiene residue** |
| MEDIUM — `OAuthAuthenticator.can_refresh()` | Resolved | Resolved | Acceptable | **Resolved** |

---

## Critical Issues (Must Fix)

_None. No reviewer flagged a runtime blocker._

## High Priority Issues (New / Unresolved — Should Fix Before Merge)

|---|---|---|---|---|
| 1 | `tests/integration/conftest.py:1171-1173` (`resolved_live_password` short-circuit) | **Re-introduces the exact regression that commit `3f99a2a` earlier in this same PR was created to fix.** When `KAMIWAZA_API_KEY` is set and valid against `/auth/users/me`, `resolved_live_password` now returns `""`. That empty string cascades into `live_password` → `test_password_authentication_allows_whoami` (`test_auth_live.py:16-28`), `test_pat_lifecycle_supports_api_key_auth` (`test_auth_live.py:31-57`), and `test_cli_login_and_pat_flow` (`test_cli_live.py:28-43`). None of those tests guard against an empty password — they feed it straight into `UserPasswordAuthenticator(user, "", …)` / `kamiwaza login --password ""`, which will fail to authenticate. Note that `_resolve_live_password_once` already caches session-wide via `_LIVE_PASSWORD_CACHE` (`conftest.py:71`), so the "kz-login runs every session" concern from the prior review was a mischaracterization — within-session it's one call, already cached. Recommended fix: **drop the `resolved_live_password` short-circuit** and instead cache the `_api_key_auth_works` probe result on `(base_url, api_key)` so `live_session_api_key` / `live_session_write_key` don't re-hit `/auth/users/me`. That delivers the desired "don't spam Keycloak" benefit without disabling the password-grant integration tests. | SP, CL (2/3) | Test Regression |
| 2 | `tests/integration/test_apps_live.py:131-140` (new anon-client assertions) | Structurally the dual-header concern is resolved — both `X-App-Session-Token` assertions use `anon_client` with `authenticator = None`, and `_request` skips `authenticate()` when the authenticator is falsy, so no `Authorization: Bearer` header is attached. **However, the assertion is too loose to prove the session-token path is actually being exercised.** `_request` rewrites any 401 (including pure-anonymous "no authenticator" rejections) into a generic `AuthenticationError("Authentication failed. No authenticator provided.")`. So if the server ever stopped honoring `X-App-Session-Token`, this test would still pass. Tighten with either `pytest.raises(AuthenticationError, match="session token")` or by asserting on `str(exc_info.value)` so the session-token rejection path is actually observed. | CL (55), CX (P2) (2/3) | Test Correctness |

## Medium Priority (Should Consider)

|---|---|---|---|---|
| 1 | `kamiwaza_sdk/client.py:212-214` (`logger.warning`) | Exception message is now bounded via `_extract_server_detail`, but the sibling `logger.warning(f"Received 401 Unauthorized. Response: {response.text}")` still emits unbounded `response.text`. Same multi-KB HTML / gateway error page problem the exception fix was meant to address, just on the log path instead. Trivially fixed by re-using `_extract_server_detail(response)` here too. | SP, CL (2/3) | Defensive / Log Hygiene |
| 2 | `tests/integration/conftest.py:1229-1231` (new `print()` in PAT-probe fallback) | When `KAMIWAZA_API_KEY` is set and `_api_key_auth_works()` fails with a non-401 `APIError` (e.g. upstream HTML 502/503), this new `print()` dumps `str(APIError)` — which includes full unbounded `response_text` — into CI logs. This reintroduces the same unbounded response-body exposure that the `_extract_server_detail` fix was meant to eliminate, just in a different pipeline. Truncate (`[:500]`) or route through `_extract_server_detail` before printing. | CX (P3, single-source but concrete) | Log Hygiene / Consistency |

## Low Priority (Nice to Have)

- **`_extract_server_detail` behavior for exotic JSON shapes**: `{"detail": null}` and `{"detail": [...]}` (FastAPI 422) fall through to `str(detail)[:max_len]` producing a Python repr. Acceptable for 401 errors, but not unit-tested. Consider a parametrized case. (SP, CL)
- **`_api_key_auth_works` call count**: still called 2-3 times per session across `resolved_live_password`, `live_session_api_key`, `live_session_write_key`. `@lru_cache`-style memoization on `(base_url, api_key)` would eliminate redundant probes. The author noted this was deferred, but it pairs naturally with the HIGH #1 fix above. (SP, CL)
- **Truncation suffix**: When `_extract_server_detail` hits the 500-char cap mid-JSON/mid-word, the result is indistinguishable from a short body. A `…` suffix (or docstring note) would be clearer. (CL)

## Reviewer Agreement Matrix

| Category | Superpowers | Claude | Codex | Consensus |
|---|---|---|---|---|
| Conftest password-short-circuit regression | **High (regression)** | **High (regression)** | Not flagged | **2/3 — regression re-introduced** |
| Dual-header test assertion looseness | Not flagged (structurally fine) | Medium (55) | **P2 (partial fix)** | **2/3 — tighten assertion** |
| Unbounded logger.warning body | Low | Medium (50) | Not flagged | 2/3 — Low/Medium |
| Unbounded `print()` on PAT probe failure | Not flagged | Not flagged | P3 | 1/3 — Medium single-source (concrete file:line) |
| HIGH #3 core fix (`_extract_server_detail`) | Resolved | Resolved | Fixed | **3/3 — Resolved** |
| HIGH #1 structural fix (anon client) | Resolved | Resolved | Resolved | **3/3 — Resolved** |
| MEDIUM OAuth `can_refresh` override | Resolved | Resolved | Acceptable | **3/3 — Resolved** |
| Security (OWASP, injection, auth) | Clean | Clean | Clean | **3/3 — No vulnerabilities** |

## Positive Observations (Delta)

- `_extract_server_detail` (`client.py:41-60`) is well-structured, defensively coded (`response.text or ""`, handles `AttributeError`), and symmetrically applied to all three 401 raise sites (`skip_auth` 208-211, `did_refresh` 235-238, non-refreshable 244-247).
- Throwaway `anon_client` pattern in `test_apps_live.py` is correct and accompanied by a comment citing server PR #1376, which will help future maintainers avoid re-introducing the PAT-precedence footgun.
- `OAuthAuthenticator.can_refresh()` docstring explicitly instructs "Flip this back to the default True once OAuth refresh lands" — nice forward guidance.
- New unit test `test_401_detail_truncates_non_json_body` uses a distinctive filler char (`Z`) and asserts `message.count("Z") == 500` — a precise, regression-proof assertion.
- New unit test `test_401_detail_surfaces_json_detail_field` asserts `'"detail"' not in message`, catching the naive "dump full JSON text" regression.
- Commit message clearly enumerates each HIGH fix and its rationale; good PR hygiene.

## Overall Verdict

**Risk Level**: Medium (no runtime SDK regression; one concrete integration-test regression + one weak assertion).

**Recommendation**: **Request Changes**

The three previously-flagged HIGH items are *mostly* addressed, and the fixes to `client.py` (`_extract_server_detail`, `can_refresh` dispatch) plus the `OAuthAuthenticator` override are clean and production-ready. However, two of the three reviewers independently identified that the `resolved_live_password` short-circuit re-introduces the exact regression that commit `3f99a2a` earlier in this PR was created to fix — password-grant integration tests will break whenever `KAMIWAZA_API_KEY` is set and valid. Additionally, two of three flagged that the new negative-path assertion in `test_apps_live.py` is too loose to positively prove session-token rejection (a pure-anonymous 401 would satisfy it). Both are pre-merge concerns.

### Recommended changes before merge

1. **Replace** the `resolved_live_password` short-circuit in `tests/integration/conftest.py:1171-1173` with **probe-result caching** (e.g. memoize `_api_key_auth_works` on `(base_url, api_key)`). This delivers the "don't spam Keycloak" benefit the short-circuit was trying to provide without disabling `test_password_authentication_allows_whoami`, `test_pat_lifecycle_supports_api_key_auth`, and `test_cli_login_and_pat_flow`.
2. **Tighten** both `pytest.raises(AuthenticationError)` blocks in `tests/integration/test_apps_live.py:131-140` with `match="session token"` (or similar), so the test actually validates the session-token auth path rather than any 401.
3. **Truncate / route through `_extract_server_detail`**: the `logger.warning` at `client.py:212-214` and the new `print()` at `conftest.py:1229-1231`. Keeps log hygiene consistent with the exception-string fix.

The remaining items (parametrized tests, `…` truncation suffix, CHANGELOG) remain non-blocking polish and can land separately.

---

Legend: SP=Superpowers, CL=Claude, CX=Codex, KZ=Kamiwaza (errored), GM=Gemini (unavailable).

*Re-review synthesized from 3 of 5 possible reviewers. Kamiwaza model returned "No review content in streamed response" on two consecutive attempts; Gemini CLI not installed locally.*

yod-kz:
## Round-2 review: all 4 items addressed (df35c97)

**ACK:** my round-1 fix for HIGH #2 (`resolved_live_password` short-circuit) was wrong-shaped — it regressed the password-grant tests (`test_password_authentication_allows_whoami`, `test_pat_lifecycle_supports_api_key_auth`, `test_cli_login_and_pat_flow`) by returning `""` for password. Apologies. Reverted, and solved the legitimate "don't spam Keycloak with PAT probes" concern at the right layer instead.

### HIGH #1 (was HIGH #2 from round-1) — PAT-probe memoization, no password regression
- Reverted the `resolved_live_password` short-circuit. Fixture now unconditionally calls `_resolve_live_password_once` (which already caches session-wide via `_LIVE_PASSWORD_CACHE`).
- Added `_API_KEY_PROBE_CACHE: dict[tuple[str, str], tuple[bool, str]]` in `conftest.py` (symmetric with `_LIVE_PASSWORD_CACHE` at L71). Memoization is inside `_api_key_auth_works` itself, so all three call sites (`resolved_live_password`, `live_session_api_key`, `live_session_write_key`) benefit without touching them individually.
- Caches both success and failure — a stale PAT won't re-probe on the next fixture access either.

### HIGH #2 (new) — Tightened session-token negative-path assertion
- Both `pytest.raises(AuthenticationError)` blocks in `test_apps_live.py` now use `match=r"(?i)session token"`.
- Confirmed against `tests/unit/test_client_401_handling.py::test_401_with_pat_does_not_retry_and_surfaces_server_detail` which shows the server's detail string is `"Invalid session token"`.
- A pure-anon 401 (`"Authentication failed. No authenticator provided."`) would not match this pattern, so if the server ever stopped honoring `X-App-Session-Token` the tests would fail — which is the point.

### MEDIUM #1 — Bounded 401 `logger.warning`
- `client.py:212-215` now routes through `_extract_server_detail(response)` instead of raw `response.text`. Same cap as the exception messages. One-line change.

### MEDIUM #2 — Truncated PAT-probe fallback `print()`
- `conftest.py` `live_session_api_key` print now truncates `error` to 500 chars with a `... [truncated]` suffix. Handles the 502/503 HTML-body case named in the review.

### Verification
- `uv run pytest -m "not integration and not live and not e2e"`: **708 passed**, 8 skipped (identical to round-1 baseline).
- `uv run pytest tests/unit/test_client_401_handling.py -v`: **9 passed**.
- `git diff ec7d1d8..HEAD`: 3 files, +45/-14 lines. Surgical, no scope creep.

### Deferred (per reviewers, non-blocking)
- CHANGELOG entry
- Parametrizing the two session-endpoint assertions
- `"…"` truncation suffix on `_extract_server_detail` itself

Re-requesting review.

m9e (commented):
# PR Re-Review Report (Round 3): fix(sdk): preserve server 401 detail for non-refreshable auth; fix apps session-token test

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/74
**Branch**: fix/uat-apps-heartbeat-session-token-401 -> release/0.12.1
**Scope**: 5 files, +489/-23
**Delta commit under review**: `df35c97` (on top of `ec7d1d8`)
**Prior reviews**: 2026-04-19 20:49 UTC (round-1, 3 HIGH), 2026-04-19 21:12 UTC (round-2, 2 HIGH + 2 MEDIUM)

**Reviewers run**: Superpowers (SP), Claude (CL), Codex (CX). Kamiwaza (KZ) errored (no streamed content). Gemini (GM) unavailable (CLI not installed). Effective N = 3.

This re-review focuses on the **delta `ec7d1d8..HEAD`** — whether the author's follow-up commit resolves the four round-2 items and any new issues introduced. Unchanged code is not re-reviewed.

---

## Prior Feedback Resolution Summary

| Round-2 Item | SP | CL | CX | Consensus |
|---|---|---|---|---|
| HIGH #1 — conftest `resolved_live_password` short-circuit regression | Resolved | Resolved | Accepted | **Resolved** |
| HIGH #2 — loose `pytest.raises(AuthenticationError)` assertion | Resolved (concluded incorrectly — see below) | Resolved (concluded incorrectly — see below) | **P1 — still broken** | **2/3 said resolved; CX identified that the "fix" cannot work** |
| MEDIUM #1 — unbounded `logger.warning` for 401 | Resolved | Resolved | Accepted | **Resolved** |
| MEDIUM #2 — unbounded PAT-probe `print()` fallback | Resolved | Resolved | Accepted | **Resolved** |

After independent verification against `kamiwaza_sdk/client.py:206-251` and the paired unit test `test_client_401_handling.py::test_401_no_authenticator_raises_without_retry` (lines 210-227), **the synthesizer confirms Codex is correct on HIGH #2**. The `match=r"(?i)session token"` regex can never match because the SDK short-circuits at line 249-251 with `"Authentication failed. No authenticator provided."` — the server's `{"detail":"Invalid session token"}` body is logged but never reaches the exception text on the `authenticator is None` branch. See High-Priority Issue #1 below.

---

## Critical Issues (Must Fix)

_None. No runtime SDK regression._

## High Priority Issues (Should Fix Before Merge)

|---|---|---|---|---|
| 1 | `tests/integration/test_apps_live.py:137-145` | **Round-2 HIGH #2 is still unresolved despite the attempted fix — the new `match=r"(?i)session token"` pattern cannot match on the anon-client path used by this test.** The test sets `anon_client.authenticator = None`, then calls `anon_client.post("/apps/sessions/heartbeat", ...)` (no `skip_auth`). In `KamiwazaClient._request` (`kamiwaza_sdk/client.py:206-251`), when `self.authenticator is None` AND `skip_auth is False`, any 401 response is short-circuited at lines 249-251 to `AuthenticationError("Authentication failed. No authenticator provided.")`. The server's `_extract_server_detail(response)` output (which would contain `"Invalid session token"`) is written to `logger.warning` at 212-215 but is **not** embedded in the raised exception on this branch. The exact behavior is pinned by the sibling unit test `test_client_401_handling.py::test_401_no_authenticator_raises_without_retry` (line 210-227), which asserts `"No authenticator provided" in str(exc_info.value)` with a `{"detail":"unauth"}` 401 response. Net effect: the live test will fail deterministically against any real or fake 401 — including the correct `"Invalid session token"` one this test is supposed to verify. The author's round-3 rationale ("a pure-anon 401 would not match this pattern, so if the server ever stopped honoring `X-App-Session-Token` the tests would fail — which is the point") is backwards: the pure-anon exception is what this test **always** produces, and is what will **always** fail the match. Two viable fixes: **(a)** pass `skip_auth=True` to both `anon_client.post(...)` calls — then lines 207-211 fire, embedding `_extract_server_detail(response)` (which is "Invalid session token") directly in the exception, so the regex matches and the session-token path is still validated; or **(b)** match on the SDK-level anon message (`r"(?i)no authenticator provided"`) — but this weakens the assertion and does not prove the server's session-token dependency was reached, so (a) is strongly preferred. Note: `.post(..., skip_auth=True)` already works end-to-end — `_request` accepts `skip_auth` as a keyword-only and strips it before forwarding kwargs to `requests.Session.request`. | CX (P1), synthesizer-verified against SDK source + paired unit test | Test Correctness (regression — round-2 HIGH #2 not actually resolved) |

## Medium Priority (Should Consider)

|---|---|---|---|---|
| 1 | `tests/integration/conftest.py:73-74` (comment on `_API_KEY_PROBE_CACHE`) | Stale docstring: the rationale still references `resolved_live_password` as a call site, but the round-3 revert removed that call. The cache is now populated only from `live_session_api_key` (`conftest.py:1242`) and `live_session_write_key` (`conftest.py:1307`). One-line comment update to avoid misleading future readers. | SP (Low), synthesizer-verified | Documentation |
| 2 | `kamiwaza_sdk/client.py:41-60` (`_extract_server_detail`) | Silent truncation with no indicator: when `len(body_or_text) > 500`, the return value is indistinguishable from a legitimately short body. The sibling conftest-side `print` truncation (`conftest.py:1251`) appends `"... [truncated]"`. Round-2 flagged this as a Low item; it remains asymmetric. A `"…"` or `"... [truncated]"` suffix in `_extract_server_detail` would make the behavior symmetric and diagnostic-friendlier. Not a blocker. | SP (Low), CL (Low) — 2/3 | Consistency / Diagnostics |
| 3 | `tests/integration/test_apps_live.py:137` | Pattern `r"(?i)session token"` is intentionally loose to survive wording drift, but this makes the regex match even text like the endpoint path fragment `"sessions"` (though the anon short-circuit issue above makes this moot until fixed). Once HIGH #1 is addressed, consider tightening to `r"(?i)invalid session token"` to pin to the exact server-contract string surfaced by `_extract_server_detail`. | CL (40), SP (Low) — 2/3 | Test Robustness |

## Low Priority (Nice to Have)

- **`_API_KEY_PROBE_CACHE` secret-in-memory note** — The cache keys on `(base_url, api_key)` and persists across the test session. Symmetric with `_LIVE_PASSWORD_CACHE` and acceptable for a test harness, but worth a brief comment that this is test-only state. (CL)
- **Shared truncation constant** — `500` now appears in `client.py:_AUTH_ERROR_DETAIL_MAX_LEN` and inline at `conftest.py:1251`. A tiny shared helper or importing the SDK constant into conftest would avoid drift if the cap ever changes. (CL)
- **Memoization coverage** — No unit test pins the `_API_KEY_PROBE_CACHE` memoization behavior (i.e., that `_api_key_auth_works` hits the network at most once per `(url, key)` pair). Optional; the behavior is a test-time optimization. (CL)

## Reviewer Agreement Matrix

| Category | Superpowers | Claude | Codex | Consensus |
|---|---|---|---|---|
| HIGH #2 fix correctness (match regex) | Resolved (incorrect conclusion) | Resolved (incorrect conclusion) | **P1 — unresolved** | **Synthesizer rules for CX: verified against client.py:249-251 + unit test** |
| HIGH #1 fix (probe cache + revert short-circuit) | Resolved | Resolved | Accepted | **3/3 Resolved** |
| MEDIUM #1 (`logger.warning` → `_extract_server_detail`) | Resolved | Resolved | Accepted | **3/3 Resolved** |
| MEDIUM #2 (`print()` truncation) | Resolved | Resolved | Accepted | **3/3 Resolved** |
| OAuth `can_refresh` override (round-1 holdover) | Resolved | Resolved | Accepted | **3/3 Resolved** |
| Stale comment on `_API_KEY_PROBE_CACHE` | Low | — | — | 1/3 Low |
| Silent truncation asymmetry | Low | Low | — | 2/3 Low |
| Security (OWASP, injection, auth) | Clean | Clean | Clean | **3/3 No vulnerabilities** |

## Positive Observations (Delta)

- `_API_KEY_PROBE_CACHE` memoization inside `_api_key_auth_works` (`conftest.py:832-845`) is the correct shape — it pushes caching one layer deeper so every caller benefits without each site needing its own guard, and it eliminates the round-2 regression where `resolved_live_password` returned `""` into the password-grant tests.
- Fix symmetry: `_extract_server_detail` is now applied to all four 401 sites in `client.py` (exception at 208-211, exception at 235-238, exception at 244-247, and now the `logger.warning` at 212-215). Log-flood and unbounded-body-in-exception vectors are both closed.
- Cache stores both success and failure outcomes, symmetric with `_LIVE_PASSWORD_CACHE`. A stale PAT won't re-probe and risk Keycloak lockout.
- Docstring on `OAuthAuthenticator.can_refresh()` explicitly instructs "Flip this back to the default True once OAuth refresh lands" — good forward signpost.
- Unit test coverage for `_extract_server_detail` is strong: `test_401_detail_surfaces_json_detail_field` and `test_401_detail_truncates_non_json_body` (the latter counts filler-`Z` chars for a precise, regression-proof 500-char assertion).
- Commit message clearly enumerates each fix and its rationale; good PR hygiene.
- Author correctly acknowledged and reverted the round-2 short-circuit regression rather than patching around it.

## Overall Verdict

**Risk Level**: Medium (one concrete, deterministic live-test failure; no runtime SDK regressions).

**Recommendation**: **Request Changes**

Three of the four round-2 items are cleanly resolved (probe caching, bounded log, bounded print). The production code paths in `kamiwaza_sdk/client.py` and `kamiwaza_sdk/authentication.py` are solid and well-tested. The blocker is that **round-2 HIGH #2 is still unresolved**: the new `match=r"(?i)session token"` regex on the anon-client integration test cannot match the exception the SDK actually raises on the `authenticator is None` branch (`"Authentication failed. No authenticator provided."`), as pinned by the sibling unit test `test_401_no_authenticator_raises_without_retry`. The live test `test_apps_deploy_and_deployment_error_paths` will fail deterministically against the UAT cluster.

Two of three reviewers (SP, CL) initially reported this item as resolved; Codex correctly identified the logical defect, and a source-level re-verification confirmed Codex's finding. This is exactly the class of issue the PR's test plan flags as "Live regression: will verify in next kzuat iteration" — the bug will surface there.

### Recommended changes before merge

1. **(HIGH)** In `tests/integration/test_apps_live.py:137-145`, pass `skip_auth=True` to both `anon_client.post(...)` calls. That routes the 401 through `client.py:207-211`, which embeds `_extract_server_detail(response)` (== `"Invalid session token"` per the server contract pinned by `test_401_detail_surfaces_json_detail_field`) in the raised exception, so the `match=r"(?i)session token"` regex matches and the test actually validates the server's session-token dependency. `_request` already accepts `skip_auth` as a keyword-only and strips it before forwarding kwargs, so no SDK-side change is needed.
2. **(Medium)** Update the stale comment on `_API_KEY_PROBE_CACHE` (`conftest.py:73-74`) to reflect that `_api_key_auth_works` is now called only from `live_session_api_key` / `live_session_write_key`, not `resolved_live_password`.
3. **(Medium, non-blocking)** Add a `"…"` or `"... [truncated]"` suffix in `_extract_server_detail` when truncation occurs, so the symmetry with the conftest-side `print()` truncation is restored and log/exception readers can tell the message was cut off.

Items #2 and #3 can land in the same fix commit as #1 or be deferred. Item #1 must land before merge — otherwise the paired live test will fail on the first kzuat run.

---

Legend: SP=Superpowers, CL=Claude, CX=Codex, KZ=Kamiwaza (errored), GM=Gemini (unavailable).

*Round-3 re-review synthesized from 3 of 5 possible reviewers. Kamiwaza model returned "No review content in streamed response"; Gemini CLI not installed locally. Codex's P1 finding was independently verified by the synthesizer against `kamiwaza_sdk/client.py:206-251` and `tests/unit/test_client_401_handling.py::test_401_no_authenticator_raises_without_retry`; SP and CL reached the opposite conclusion via an incomplete trace of the `authenticator is None` control flow.*

yod-kz:
Round 3 landed: e6f0001.

ACK the round-2 HIGH #2 fix was wrong-shaped — thank you Codex for catching it. The test used `KamiwazaClient(..., authenticator=None)`, which in `_request` takes the `authenticator is None` short-circuit at `client.py:249-251` and raises `"Authentication failed. No authenticator provided."` The server's `_extract_server_detail` output only reaches `logger.warning` on that branch, never the raised exception, so the round-2 `match=r"(?i)session token"` regex could not possibly have matched against a real 401. Deterministic failure waiting to happen.

Three items landed per the round-3 recommendations:

- **HIGH #1 — Route anon-client 401 through `skip_auth=True`:** Both `anon_client.post(...)` calls in `tests/integration/test_apps_live.py` now pass `skip_auth=True`, forcing the branch at `client.py:207-211` that embeds `_extract_server_detail(response)` in the raised exception. Verified `skip_auth` is a keyword-only kwarg on `_request` (signature at `client.py:161-168`), consumed before `**kwargs` is forwarded to `requests.Session.request`.
- **MEDIUM #3 — Tightened regex:** Assertion pinned to `r"(?i)invalid session token"`, matching the server-contract string pinned by `test_401_detail_surfaces_json_detail_field`.
- **MEDIUM #2 — Truncation suffix:** Extracted `_truncate_with_suffix` helper in `kamiwaza_sdk/client.py`. `_extract_server_detail` now appends `... [truncated]` when the JSON-detail / JSON-body / raw-text path exceeds `max_len`. Updated `test_401_detail_truncates_non_json_body` to also assert the suffix is present (existing `message.count("Z") == 500` remains valid — suffix chars don't overlap with `Z`).
- **MEDIUM #1 — Cache docstring:** One-line update to the `_API_KEY_PROBE_CACHE` comment in `tests/integration/conftest.py`, removing the stale reference to `resolved_live_password`. Call sites are now `live_session_api_key` and `live_session_write_key`.

Did NOT touch: production `authenticator is None` short-circuit (correct as-is), `_API_KEY_PROBE_CACHE` semantics (3/3 said resolved), shared truncation constant between `client.py` and `conftest.py` (deferred per review), memoization coverage test (optional per review).

Unit suite: 708 passed, 8 skipped, 0 regressions (`uv run pytest -m "not integration and not live and not e2e"`). `tests/unit/test_client_401_handling.py` all 9 green including the updated truncation assertion.

m9e (commented):
# PR Re-Review Report (Round 4): fix(sdk): preserve server 401 detail for non-refreshable auth; fix apps session-token test

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/74
**Branch**: fix/uat-apps-heartbeat-session-token-401 -> release/0.12.1
**Scope**: 5 files, +516/-23
**Delta commit under review**: `e6f0001` (on top of `df35c97`)
**Prior reviews**: round-1 `e357081`, round-2 `ec7d1d8`, round-3 `df35c97`

**Reviewers run**: Superpowers (SP), Claude (CL), Codex (CX). Kamiwaza (KZ) errored ("No review content in streamed response"). Gemini (GM) unavailable (CLI not installed). Effective N = 3.

This re-review focuses on the **delta `df35c97..HEAD`** — whether the author's four round-4 claims are correctly implemented and whether any new issues were introduced. Unchanged code is not re-reviewed.

**Note on Codex scope**: Codex CLI v0.121.0 rejects combining `--base <BRANCH>` with a custom prompt, so Codex reviewed the full PR delta rather than the round-4 delta specifically. Its one P2 finding (on `resolved_live_password` / `_api_key_auth_works` probe cost) is addressed below in the context of the round-3 consensus.

---

## Prior Feedback Resolution Summary

| Round-3 Item | SP | CL | CX | Consensus |
|---|---|---|---|---|
| HIGH #1 — anon-client 401 not routed through `_extract_server_detail`-embedding branch | Resolved | Resolved | Resolved | **3/3 Resolved** |
| MEDIUM #1 — stale `_API_KEY_PROBE_CACHE` docstring | Resolved | Resolved | (not flagged) | **Resolved** |
| MEDIUM #2 — silent truncation with no indicator | Resolved | Resolved | (not flagged) | **Resolved** |
| MEDIUM #3 — loose `r"(?i)session token"` regex | Resolved (tightened to `r"(?i)invalid session token"`) | Resolved | Resolved | **3/3 Resolved** |

Independent source-level verification confirms each claim:

- **`skip_auth=True`** — `client.py:183` accepts `skip_auth` as keyword-only (stripped before forwarding). On a 401, `client.py:222-227` short-circuits to `AuthenticationError(f"Unauthenticated request failed for {endpoint}: {_extract_server_detail(response)}")`. Both `anon_client.post(...)` calls in `test_apps_live.py:142-145,148-153` now pass `skip_auth=True`.
- **Regex tightening** — `test_apps_live.py:140` uses `match=r"(?i)invalid session token"`. Server returns `{"detail":"Invalid session token"}`; `_extract_server_detail` unwraps to `"Invalid session token"`; full exception message matches. Contract is pinned by `tests/unit/test_client_401_handling.py::test_401_detail_surfaces_json_detail_field` (line 231-253).
- **`_truncate_with_suffix` helper** — `client.py:42-53` returns `value[:max_len] + "... [truncated]"` only when `len(value) > max_len` (no spurious suffix on exact-length bodies). Applied uniformly across all three paths in `_extract_server_detail` (`client.py:56-76`). Suffix contains no `Z`, so `message.count("Z") == 500` in `test_401_detail_truncates_non_json_body` (line 283) still holds; new assertion at line 287 locks the suffix.
- **Cache docstring refresh** — `conftest.py:70-79` now accurately reflects the two current call sites (`live_session_api_key:1242`, `live_session_write_key:1307`) with no stale reference to `resolved_live_password`.

---

## Critical Issues (Must Fix)

_None. No runtime SDK regression._

## High Priority Issues (Should Fix Before Merge)

_None. All three reviewers agree the round-4 delta resolves the round-3 blocker. No new high-priority issues introduced._

## Medium Priority (Should Consider — Non-Blocking)

|---|---|---|---|---|
| 1 | `tests/integration/conftest.py` — `resolved_live_password` / `_api_key_auth_works` interaction (**not part of the round-4 delta**) | Codex reiterates the round-3 concern that `live_session_api_key` / `live_session_write_key` still force `resolved_live_password` resolution, which calls `_resolve_live_password_once()` → `_password_auth_works()` once per session even when `KAMIWAZA_API_KEY` is valid. A stale fallback password therefore causes one unnecessary failed password grant per session. **Context**: this was explicitly the round-3 consensus trade-off — the author's earlier short-circuit (returning `""` for `live_password`) broke `test_password_authentication_allows_whoami`, `test_pat_lifecycle_supports_api_key_auth`, and `test_cli_login_and_pat_flow`. `_LIVE_PASSWORD_CACHE` caches the outcome so the grant runs at most once per session, not per-test. Not introduced by round-4; mentioned here only for full context. If the "one failed login per session" cost is a blocker in UAT, a proper fix would add an "API-key-is-good → skip password resolution **and** fake out the password-grant tests' fixture dependency" path rather than re-introducing the blanket short-circuit. | CX (P2, scoped to full PR not round-4) | Test Infra / Operational |
| 2 | `kamiwaza_sdk/client.py:42-53` (`_truncate_with_suffix`) | Docstring says "Truncate `value` to `max_len` chars" but the returned string can be up to `max_len + len(suffix)` chars (e.g. 517 when `max_len=500`). Cosmetic. If any downstream log aggregator were to assume a strict ≤500 cap, they'd see 517. Simple doc reword suggested. | CL (80), SP (Low) — 2/3 | Documentation |
| 3 | `kamiwaza_sdk/client.py:42-53` | If a future caller passes `max_len < len(suffix)` (17), the truncated output is *longer* than the cap. Not exploitable at the current single call site (`max_len=500`), but a defensive `if max_len <= len(suffix): return value[:max_len]` would remove the anomaly. | CL (40) | Edge-case robustness |

## Low Priority (Nice to Have)

- `tests/integration/conftest.py:1251` — Truncation-with-suffix is now duplicated between SDK (`_truncate_with_suffix`) and conftest. Minor DRY nit; crossing the SDK private-API boundary from tests is itself questionable, so acceptable as-is. (SP, CL)
- `kamiwaza_sdk/client.py` — `_truncate_with_suffix` has no dedicated unit test — only exercised indirectly via `test_401_detail_truncates_non_json_body`. A tiny parametrized test covering boundaries `len==max`, `len==max+1`, `len==0` would lock the contract independent of `_extract_server_detail`. (SP)
- `kamiwaza_sdk/client.py:228-231` — Non-`skip_auth` 401 path still logs `_extract_server_detail(response)` via `logger.warning` before raising; for PAT 401s this is a double-emit (log once, then embed in exception). Pre-existing, not introduced by round-4. (CL)
- Unit-level coverage gap: no unit test directly exercises the `skip_auth=True` branch in `_request`. Currently only exercised via the live integration test. A stubbed unit test verifying a 401 with `skip_auth=True` raises `AuthenticationError` containing `_extract_server_detail` output would lock the branch. (CL)
- Multi-element refactor (four similar `_extract_server_detail(response)` call sites in `client.py`) is not worth extracting yet but a `_auth_error(endpoint, response, reason)` helper would pay off if a fifth 401 branch ever lands. (SP)

## Reviewer Agreement Matrix

| Category | Superpowers | Claude | Codex | Consensus |
|---|---|---|---|---|
| `skip_auth=True` routing correctness | Resolved | Resolved | Resolved | **3/3 Resolved** |
| Regex tightening (`(?i)invalid session token`) | Resolved | Resolved | Resolved | **3/3 Resolved** |
| `_truncate_with_suffix` correctness | Resolved | Resolved | Resolved | **3/3 Resolved** |
| Cache docstring refresh | Resolved | Resolved | (n/a) | **Resolved** |
| `_truncate_with_suffix` docstring vs behavior | Low | 80 | — | 2/3 Low |
| `_truncate_with_suffix` `max_len < suffix` edge | — | 40 | — | 1/3 Low |
| Password-grant probe cost (full-PR re-flag) | — | — | P2 | 1/3 Medium (not a delta issue) |
| Security (OWASP, injection, auth) | Clean | Clean | Clean | **3/3 No vulnerabilities** |

## Positive Observations (Delta)

- The fix to the round-3 blocker is structurally sound: `skip_auth=True` is the exact mechanism needed to route the 401 through the detail-embedding branch. Author correctly identified (via Codex round-3 feedback) that this is the only way the "invalid session token" regex can actually match.
- Regex tightening from `(?i)session token` to `(?i)invalid session token` is a meaningful specificity gain: it now rejects a hypothetical `"Session token required"` string that would have passed the old loose match. Inline comment pinning the contract to `test_401_detail_surfaces_json_detail_field` is exactly the right breadcrumb for future maintainers.
- `_truncate_with_suffix` is a clean YAGNI-respecting extraction — no generalization, no optional parameters beyond the existing `max_len`. Applied symmetrically across all three paths in `_extract_server_detail` (raw text, JSON-detail string, JSON-detail non-string fallback).
- Test assertion design is robust: `message.count("Z") == 500` survives the new suffix unchanged because the suffix contains no `Z`. The new `"... [truncated]" in message` assertion at line 287 locks the observable surface, not the internal constant.
- Author updated the cache docstring surgically to reflect the actual call graph after the round-3 revert. Good hygiene.
- Author correctly acknowledged previous round's misanalysis (round-3 HIGH #2 mis-resolution) and applied the fix Codex identified. Good receptivity to review feedback.
- Commit message enumerates each fix and its rationale.

## Overall Verdict

**Risk Level**: Low (no runtime SDK regressions; no new blocking issues introduced; the round-3 integration-test blocker is fixed).

**Recommendation**: **Approve**

All four round-4 claims verified as resolved by 3/3 reviewers with file:line evidence. The session-token integration test is now correctly structured to exercise the server-detail-propagation code path. Unit tests cover JSON-detail unwrapping, non-JSON truncation, and the new suffix. Only medium-priority items remain (cosmetic docstring, edge-case defense, deferred test infra trade-off) — all non-blocking.

### Optional follow-ups (non-blocking, any commit)

1. Reword `_truncate_with_suffix` docstring to reflect that the return can exceed `max_len` by `len(suffix)` chars.
2. Add a small unit test directly for `_truncate_with_suffix` covering `len==max_len`, `len==max_len+1`, `len==0`, and `max_len < len(suffix)`.
3. (Optional, scope-wider) Address the "one failed password grant per session when PAT is valid and fallback password is stale" operational cost in `resolved_live_password` — but only if it's actually causing UAT pain. The round-3 consensus trade-off was correct in principle; it can be revisited if UAT telemetry shows the cost.

---

Legend: SP=Superpowers, CL=Claude, CX=Codex, KZ=Kamiwaza (errored), GM=Gemini (unavailable).

*Round-4 re-review synthesized from 3 of 5 possible reviewers. Kamiwaza model returned "No review content in streamed response"; Gemini CLI not installed locally. Codex's scope was the full PR (CLI rejects `--base` + prompt combined) rather than the round-4 delta; its single finding is a repeat of the round-2/3 trade-off explicitly accepted in round-3 consensus.*

jxstanford (commented):
# Review Report: fix(sdk): preserve server 401 detail for non-refreshable auth; fix apps session-token test

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/74
**Branch**: `fix/uat-apps-heartbeat-session-token-401` → `release/0.12.1`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| # | File:Line | Issue | Category | Confidence |
|---|-----------|-------|----------|------------|

## High Priority Issues
| # | File:Line | Issue | Category | Confidence |
|---|-----------|-------|----------|------------|

| # | File:Line | Issue | Category | Confidence |
|---|-----------|-------|----------|------------|
| 1 | `tests/integration/conftest.py:1251` | Truncation logic is duplicated inline rather than reusing the new `_truncate_with_suffix` helper defined in `client.py`. Minor; extracting to a shared test util or importing the private helper would prevent drift (constants already drift-prone at 500). | Code quality | Medium |
| 2 | `kamiwaza_sdk/client.py:56-76` | `_extract_server_detail` only special-cases `detail`; FastAPI validation errors return `detail` as a list of dicts, which will fall through to `str(detail)` and produce a noisy repr. Not a regression (old code dumped full `response.text`), but tightening that path to pretty-print list-of-dicts would improve the UAT-driver-facing error text this PR aims to improve. | Error handling / UX | Medium |
| 3 | `tests/integration/test_apps_live.py:129-145` | Using `KamiwazaClient(live_server_available)` then `authenticator=None` + `skip_auth=True` is an awkward double-negation; a `KamiwazaClient(live_server_available, api_key=None)` still picks up `KAMIWAZA_API_KEY` from env, which is why the explicit `= None` is needed. Consider adding a `no_env=True` or similar constructor flag in a follow-up — not blocking. | Test ergonomics | Low |
| 4 | `kamiwaza_sdk/client.py:228-231` | The 401 warning log now emits the server detail (previously `response.text`). For endpoints where the server echoes parts of the presented token/claims in error text, this marginally widens what lands in log files. Server-side responses are trusted here, so informational only — worth noting if log sinks ship to untrusted storage. | Sensitive data exposure | Low |

## Test Coverage Assessment
- Files with adequate coverage: `kamiwaza_sdk/authentication.py` (unit tests for `can_refresh` on `ApiKeyAuthenticator` and `OAuthAuthenticator`), `kamiwaza_sdk/client.py` (8 unit tests covering PAT no-retry, refreshable retry-once, legacy auth, attribute-style `can_refresh`, no-authenticator, JSON-detail extraction, body truncation)
- Files needing tests: None. Integration test `test_apps_deploy_and_deployment_error_paths` now pins the server-side contract with an anon-client path; unit coverage makes that integration assertion independently verifiable.

## Positive Observations
- The duck-typed `can_refresh` reader (`getattr(..., True)` + `callable()`-aware) preserves backward compat for any downstream custom `Authenticator` subclasses that predate this hook; `test_401_legacy_authenticator_without_can_refresh_still_retries_once` locks that in.
- `OAuthAuthenticator.can_refresh() -> False` closes a latent bug: a 401 under pre-authenticated OAuth would have raised `NotImplementedError` from the retry path. Nice catch to flip it alongside the main fix.
- `_extract_server_detail` truncates with an explicit `... [truncated]` suffix (not a naked slice), and the unit test uses a distinctive filler char to unambiguously assert the truncation length.
- Session-scoped memoization of `_api_key_auth_works` avoids hammering Keycloak on every fixture setup and handles the account-lockout concern called out in the docstring.
- Removal of the `_resolve_live_password_once` short-circuit is correctly accompanied by a `resolved_live_password` docstring explaining *why* (password-auth tests consume it directly). Prevents future regressions by someone re-adding the short-circuit.
- Commit history shows three rounds of review feedback incorporated cleanly (tightened regex match, added truncation suffix, memoized probe).

## Overall Verdict
**Risk Level**: Low
**Rationale**: Targeted fix with a clear UAT-sourced motivation. Non-breaking API addition (`can_refresh`), duck-typed read protects legacy auths, and unit + integration tests together pin both the SDK behavior and the server-side session-token contract. Only minor suggestions — no blockers.

**Engines**: Claude Code (1/4 available)

---
🤖 Generated with [Claude Code](https://claude.ai/code) PR review

---
<details>
<summary>🔧 Engine Health (1/4 contributing — Low (1 engine — verdict withheld))</summary>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ❌ Failed (exit 0) |
| Gemini | ❌ Failed (exit 1) |

**Engines contributing**: 1/4
</details>

PR #76, opened 2026-04-24
feat(extensions): UAC-9d runtime-lib exception surface + CLI plumbing (ENG-3885)

jonathanleff:
## Summary

Python side of ENG-3885 — UAC-9d runtime-lib exception hierarchy and CLI plumbing, plus two walkthrough fixes (§4.8 P5, B3) that live in the same subsystem.

Part of D210 Milestone 1 (Phase A). TypeScript mirror will land in a follow-up PR.

## What's in this PR

**Runtime lib (`kamiwaza_extensions_lib/`):**
- `errors.py` (new) — `KamiwazaRuntimeError` base + 4 UAC-9d subclasses, each carrying a canonical `class_name`.
- `exception_names.json` (new) — single source of truth for `class_name` → `exit_code` → `doctor_hint`.
- `identity.py` — new strict `extract_identity()` raising `MisboundAuthError` on missing `X-User-Id`/`X-Workroom-Id`; permissive `identity_from_headers()` preserved. Added `workroom_role`, `auth_token`, `system_high` fields. Converted `Identity` to Pydantic (`.model_dump()` works).
- `anonymous_identity()` helper so `require_auth()` and `/session` return a byte-identical `Identity(name="Anonymous")` under `USE_AUTH=false` (fixes §4.8 P5 — the "Hello, None" papercut).
- `session.py` — `/session` now serializes via `Identity.model_dump()` rather than hand-rolling the dict.

**CLI (`kamiwaza_extensions/`):**
- `exit_codes.py` (new) — `ExitCode` IntEnum with 0/1/2, UAC-9d 10–13, and cluster/registry 20–23 (incl. `CLUSTER_NOT_READY=23` consumed by ENG-3888).
- `exit_code_for(class_name)` loads from `exception_names.json` via `importlib.resources`; unknown class names fall back to `FAILURE`.
- `cli.py::_handle_exception` — UAC-9d branch maps `KamiwazaRuntimeError` → the canonical exit code.
- `doctor.py` — reference-entry pass emits one `CheckResult` per UAC-9d class with the canonical doctor hint.
- `commands/login.py` — after PAT mint, decodes the JWT + calls `/auth/whoami`; warns when UI role-set is a strict superset of the PAT's claim (§4.8 B3).

**Contract test (`tests/unit/extensions/test_doctor_hint_coverage.py`):** the single invariant binding the 3 moving pieces — for every class in `exception_names.json`, asserts a matching `CheckResult` in doctor output AND `exit_code_for()` mapping.

## Test scenarios covered

| ID | Scenario | Commit |
|---|---|---|
| TS-1 | ExitCode enum values (0/1/2, 10-13, 20-23) | `2c2df7a` |
| TS-2 | exit_code_for() maps all 4 UAC-9d class names | `b4243b4` |
| TS-3 | test_doctor_hint_coverage contract | `fb6b6ca` |
| TS-4 | MisboundAuthError raised when X-User-Id missing | `2635be2` |
| TS-5 | MisboundAuthError raised when X-Workroom-Id missing | `2635be2` |
| TS-6 | Identity.model_dump() (Pydantic conversion) | `a93e79f` |
| TS-7 | require_auth + /session produce byte-identical Identity under USE_AUTH=false | `9c5d340` |
| TS-8 | CLI handler: RuntimeLibException → exit_code_for(class_name) | `5468576` |
| TS-32 | kz-ext login warns on UI ⊋ PAT role-set | `04621b9` |

**9/9 in-scope test scenarios pass.** Full unit suite: **591 passed**, **+28 new tests**, zero regressions.

## Known follow-ups (NOT in this PR)

- **TypeScript mirror** — `@kamiwaza-ai/extensions-lib`: `errors.ts`, `identity.ts` Pydantic-equivalent shape, `AuthGuard` passthrough (P6 of ENG-3889). Separate PR so we can ship the Python side without blocking on the TS build.
- **Docs** — UAC-9d failure-classes runbook (`docs/services/extensions/uac-9d-failure-classes.md`) and `kamiwaza-extensions-lib` v0.2 CHANGELOG. Docs PR once TS mirror lands.
- **P6** (AuthGuard passthrough) — handled in the TS-side portion of ENG-3889, not here.

## Cross-references

- Linear: [ENG-3885](https://linear.app/kamiwaza/issue/ENG-3885) (In Progress)
- Design: §4.2.7 RuntimeLibExceptionHierarchy + IdentityExtractor, §4.2.8 DoctorUACFailureHints + ExitCodeMap, §4.2.12 StarterBackendHardening (P5), §4.8 P4/P5/B3
- Exit code 23 (`CLUSTER_NOT_READY`) is defined here but consumed by ENG-3888 (cluster readiness probe), which will import it once this lands.

## Test plan

- [x] `make test` passes on this branch (591 unit tests, 0 regressions)
- [x] New contract test enforces 1:1 coverage between `exception_names.json`, `exit_codes.py`, and `doctor.py`
- [ ] TypeScript mirror PR lands before M1 exit
- [ ] Docs PR lands before M1 exit
- [ ] Manual: `kz-ext doctor` on a dev machine shows the 4 UAC-9d entries
- [ ] Manual: simulate role-set downgrade on login to confirm warning fires

### DIFF
--- a/docs/services/extensions/failure-classes.md
+++ b/docs/services/extensions/failure-classes.md
@@ -71,16 +71,16 @@ doesn't match the extension's expected runtime context.
 
 **Typical causes:**
 
-- Running a local-dev envelope against a production-signed extension (rare;
-extensions no longer verify HMAC as of 2026-04-23, but shape mismatches
-still apply).
+- Running a local-dev envelope against a production-deployed extension.
+  Note: extensions are pass-through consumers of the envelope — they do
+  not verify the HMAC signature (see design §4.2.7 for rationale), but
+  shape mismatches still apply.
 - Envelope fields present but all empty — platform misconfiguration.
 
 **Fix:**
 
-1. Compare `request.headers` against the list in
-  `kamiwaza-sdk/docs/extensions/non-sdk-flow.md` (when available) or
-   `kamiwaza_extensions_lib/identity.py`.
+1. Compare `request.headers` against the header list in
+   `kamiwaza_extensions_lib/identity.py` (``_HEADER_USER_ID``, etc.).
 2. Check that `X-User-Id` and `X-Workroom-Id` are non-empty strings, not
   placeholder tokens.
 3. Verify the deployment environment — a PR preview env often has different
@@ -106,13 +106,15 @@ workroom viewer.
    membership.
 2. The platform enforces workroom membership at data-access time; the
   extension should surface a friendly error, not try to work around it.
-3. Code example:
+3. When the runtime lib raises ``OutOfEnvelopeAccessError`` directly, catch
+   it at the top of your request handler:
 
 ```python
 from kamiwaza_extensions_lib.errors import OutOfEnvelopeAccessError
 
 try:
-    result = client.workrooms.get_resource(identity.workroom_id, resource_id)
+    # ...call a runtime-lib helper that enforces envelope scope...
+    pass
 except OutOfEnvelopeAccessError:
     return JSONResponse(
         {"error": {"class": "out_of_envelope_access",
@@ -120,6 +122,13 @@ except OutOfEnvelopeAccessError:
         status_code=403,
     )
 ```
+
+As of this writing the runtime lib does not automatically translate
+platform HTTP 403s into ``OutOfEnvelopeAccessError``. If you call the
+SDK directly (`client.workrooms.*`, `client.models.*`, etc.), treat a
+403 response as an out-of-envelope condition and return the same 403
+JSON shape shown above. A future mapper is tracked under the runtime-lib
+backlog.
 
 ## `platform_outage` — exit 13
 
--- a/docs/services/extensions/failure-classes.md
+++ b/docs/services/extensions/failure-classes.md
@@ -1,4 +1,4 @@
-# UAC-9d Failure Classes — Runbook
+# Extension Failure Classes — Runbook
 
 When an extension cannot serve a request, the runtime libs raise one of four
 canonical exception classes. Each class has a stable name, a CLI exit code,
@@ -7,18 +7,20 @@ its typical causes, and how to fix it.
 
 ## At a glance
 
-| Class name | Python / TS exception | CLI exit code | `kz-ext doctor` hint |
-|------------|----------------------|---------------|----------------------|
-| `misbound_auth` | `MisboundAuthError` | 10 | Required envelope header missing or malformed |
-| `unexpected_context` | `UnexpectedContextError` | 11 | Envelope shape wrong for this context |
-| `out_of_envelope_access` | `OutOfEnvelopeAccessError` | 12 | Cross-workroom / out-of-scope access attempt |
-| `platform_outage` | `PlatformOutageError` | 13 | Platform API unreachable or returning 5xx |
+
+| Class name               | Python / TS exception      | CLI exit code | `kz-ext doctor` hint                          |
+| ------------------------ | -------------------------- | ------------- | --------------------------------------------- |
+| `misbound_auth`          | `MisboundAuthError`        | 10            | Required envelope header missing or malformed |
+| `unexpected_context`     | `UnexpectedContextError`   | 11            | Envelope shape wrong for this context         |
+| `out_of_envelope_access` | `OutOfEnvelopeAccessError` | 12            | Cross-workroom / out-of-scope access attempt  |
+| `platform_outage`        | `PlatformOutageError`      | 13            | Platform API unreachable or returning 5xx     |
+
 
 The `class_name` strings are canonical — they appear in:
 
 - The JSON body returned by non-SDK extensions (under `{"error": {"class": ...}}`).
 - `kamiwaza_extensions_lib/exception_names.json` (single source of truth).
-- `kz-ext doctor` output (as `UAC-9d: <class_name>` reference entries).
+- `kz-ext doctor` output (as `Failure class: <class_name>` reference entries).
 
 ## `misbound_auth` — exit 10
 
@@ -28,18 +30,18 @@ is missing or empty at the time the runtime lib tried to parse the request.
 **Typical causes:**
 
 - The request reached the extension pod *without* passing through Traefik
-  (e.g., direct in-cluster call, port-forward, or misconfigured ingress).
+(e.g., direct in-cluster call, port-forward, or misconfigured ingress).
 - The platform ForwardAuth layer is not injecting the envelope correctly
-  (transient platform outage or misconfiguration).
+(transient platform outage or misconfiguration).
 - Local development with `KAMIWAZA_USE_AUTH=true` but no platform to populate
-  the envelope — expected failure mode.
+the envelope — expected failure mode.
 
 **Fix:**
 
 1. Confirm the request is reaching the extension *through* Traefik. Check
-   ingress rules and the request URL.
+  ingress rules and the request URL.
 2. Run `kz-ext doctor` — a failed `Kamiwaza connection` check often precedes
-   `misbound_auth`.
+  `misbound_auth`.
 3. For local dev, set `KAMIWAZA_USE_AUTH=false` and restart the extension.
 
 **Code example:**
@@ -70,19 +72,19 @@ doesn't match the extension's expected runtime context.
 **Typical causes:**
 
 - Running a local-dev envelope against a production-signed extension (rare;
-  extensions no longer verify HMAC as of 2026-04-23, but shape mismatches
-  still apply).
+extensions no longer verify HMAC as of 2026-04-23, but shape mismatches
+still apply).
 - Envelope fields present but all empty — platform misconfiguration.
 
 **Fix:**
 
 1. Compare `request.headers` against the list in
-   `kamiwaza-sdk/docs/extensions/non-sdk-flow.md` (when available) or
+  `kamiwaza-sdk/docs/extensions/non-sdk-flow.md` (when available) or
    `kamiwaza_extensions_lib/identity.py`.
 2. Check that `X-User-Id` and `X-Workroom-Id` are non-empty strings, not
-   placeholder tokens.
+  placeholder tokens.
 3. Verify the deployment environment — a PR preview env often has different
-   envelope provenance than staging or prod.
+  envelope provenance than staging or prod.
 
 ## `out_of_envelope_access` — exit 12
 
@@ -92,18 +94,18 @@ does not cover — typically a cross-workroom access attempt.
 **Typical causes:**
 
 - The extension is resolving a workroom ID from user input and calling the
-  platform, but the caller does not have membership in that workroom.
+platform, but the caller does not have membership in that workroom.
 - A stale token referencing a deleted workroom.
 - Extension logic that assumes admin-like access when the caller is a
-  workroom viewer.
+workroom viewer.
 
 **Fix:**
 
 1. Always thread the request's `Identity.workroom_id` through platform API
-   calls — don't substitute a workroom ID from user input without re-checking
+  calls — don't substitute a workroom ID from user input without re-checking
    membership.
 2. The platform enforces workroom membership at data-access time; the
-   extension should surface a friendly error, not try to work around it.
+  extension should surface a friendly error, not try to work around it.
 3. Code example:
 
 ```python
@@ -127,25 +129,25 @@ except OutOfEnvelopeAccessError:
 
 - Platform is down or degraded (check Kamiwaza status).
 - Network partition between the extension and the platform — especially in
-  split-cluster deployments.
+split-cluster deployments.
 - Upstream model-access boundary is failing (5xx propagation).
 
 **Fix:**
 
 1. Run `kz-ext doctor` — the `Kamiwaza connection` check will fail in the
-   same direction.
+  same direction.
 2. Check the platform status page / oncall channels.
 3. For transient failures, the runtime lib's `TokenRefreshMiddleware` already
-   retries once on 401 mid-stream. A double failure surfaces as
+  retries once on 401 mid-stream. A double failure surfaces as
    `PlatformOutageError`.
 4. Do **not** catch and swallow — let it bubble to the top-level error
-   handler so the exit code is preserved.
+  handler so the exit code is preserved.
 
 ## How the exit codes flow
 
 1. The runtime lib raises a `KamiwazaRuntimeError` subclass.
 2. If the exception bubbles into a `kz-ext` subcommand, the CLI's top-level
-   `_handle_exception` catches it and calls `exit_code_for(exc.class_name)`.
+  `_handle_exception` catches it and calls `exit_code_for(exc.class_name)`.
 3. The process exits with the canonical code from `exception_names.json`.
 
 CI and operators can key off the exit code alone:
@@ -180,3 +182,4 @@ drifts from either the `ExitCode` enum or the `kz-ext doctor` output.
 - `§4.2.8 DoctorUACFailureHints + ExitCodeMap`
 - `§5 Q6 Security implications and auth interactions`
 - Linear: [ENG-3885](https://linear.app/kamiwaza/issue/ENG-3885)
+
--- a/docs/services/extensions/uac-9d-failure-classes.md
+++ b/docs/services/extensions/uac-9d-failure-classes.md
@@ -1,0 +1,182 @@
+# UAC-9d Failure Classes — Runbook
+
+When an extension cannot serve a request, the runtime libs raise one of four
+canonical exception classes. Each class has a stable name, a CLI exit code,
+and a fix hint surfaced by `kz-ext doctor`. This runbook explains each class,
+its typical causes, and how to fix it.
+
+## At a glance
+
+| Class name | Python / TS exception | CLI exit code | `kz-ext doctor` hint |
+|------------|----------------------|---------------|----------------------|
+| `misbound_auth` | `MisboundAuthError` | 10 | Required envelope header missing or malformed |
+| `unexpected_context` | `UnexpectedContextError` | 11 | Envelope shape wrong for this context |
+| `out_of_envelope_access` | `OutOfEnvelopeAccessError` | 12 | Cross-workroom / out-of-scope access attempt |
+| `platform_outage` | `PlatformOutageError` | 13 | Platform API unreachable or returning 5xx |
+
+The `class_name` strings are canonical — they appear in:
+
+- The JSON body returned by non-SDK extensions (under `{"error": {"class": ...}}`).
+- `kamiwaza_extensions_lib/exception_names.json` (single source of truth).
+- `kz-ext doctor` output (as `UAC-9d: <class_name>` reference entries).
+
+## `misbound_auth` — exit 10
+
+**What it means:** A required envelope header (`X-User-Id` or `X-Workroom-Id`)
+is missing or empty at the time the runtime lib tried to parse the request.
+
+**Typical causes:**
+
+- The request reached the extension pod *without* passing through Traefik
+  (e.g., direct in-cluster call, port-forward, or misconfigured ingress).
+- The platform ForwardAuth layer is not injecting the envelope correctly
+  (transient platform outage or misconfiguration).
+- Local development with `KAMIWAZA_USE_AUTH=true` but no platform to populate
+  the envelope — expected failure mode.
+
+**Fix:**
+
+1. Confirm the request is reaching the extension *through* Traefik. Check
+   ingress rules and the request URL.
+2. Run `kz-ext doctor` — a failed `Kamiwaza connection` check often precedes
+   `misbound_auth`.
+3. For local dev, set `KAMIWAZA_USE_AUTH=false` and restart the extension.
+
+**Code example:**
+
+```python
+from kamiwaza_extensions_lib.errors import MisboundAuthError
+from kamiwaza_extensions_lib.identity import extract_identity
+
+try:
+    identity = extract_identity(dict(request.headers))
+except MisboundAuthError as exc:
+    # Return 401 to the caller — do not leak internal details.
+    return JSONResponse(
+        {"error": {"class": exc.class_name, "detail": "Authentication required"}},
+        status_code=401,
+    )
+```
+
+Note: the built-in `require_auth()` FastAPI dependency handles this for you —
+you only need to catch `MisboundAuthError` if you are parsing headers manually
+or implementing a non-SDK extension.
+
+## `unexpected_context` — exit 11
+
+**What it means:** The envelope headers are present but the shape or content
+doesn't match the extension's expected runtime context.
+
+**Typical causes:**
+
+- Running a local-dev envelope against a production-signed extension (rare;
+  extensions no longer verify HMAC as of 2026-04-23, but shape mismatches
+  still apply).
+- Envelope fields present but all empty — platform misconfiguration.
+
+**Fix:**
+
+1. Compare `request.headers` against the list in
+   `kamiwaza-sdk/docs/extensions/non-sdk-flow.md` (when available) or
+   `kamiwaza_extensions_lib/identity.py`.
+2. Check that `X-User-Id` and `X-Workroom-Id` are non-empty strings, not
+   placeholder tokens.
+3. Verify the deployment environment — a PR preview env often has different
+   envelope provenance than staging or prod.
+
+## `out_of_envelope_access` — exit 12
+
+**What it means:** The extension attempted to access a resource the envelope
+does not cover — typically a cross-workroom access attempt.
+
+**Typical causes:**
+
+- The extension is resolving a workroom ID from user input and calling the
+  platform, but the caller does not have membership in that workroom.
+- A stale token referencing a deleted workroom.
+- Extension logic that assumes admin-like access when the caller is a
+  workroom viewer.
+
+**Fix:**
+
+1. Always thread the request's `Identity.workroom_id` through platform API
+   calls — don't substitute a workroom ID from user input without re-checking
+   membership.
+2. The platform enforces workroom membership at data-access time; the
+   extension should surface a friendly error, not try to work around it.
+3. Code example:
+
+```python
+from kamiwaza_extensions_lib.errors import OutOfEnvelopeAccessError
+
+try:
+    result = client.workrooms.get_resource(identity.workroom_id, resource_id)
+except OutOfEnvelopeAccessError:
+    return JSONResponse(
+        {"error": {"class": "out_of_envelope_access",
+                   "detail": "You do not have access to this workroom"}},
+        status_code=403,
+    )
+```
+
+## `platform_outage` — exit 13
+
+**What it means:** The platform API is unreachable or returning 5xx responses.
+
+**Typical causes:**
+
+- Platform is down or degraded (check Kamiwaza status).
+- Network partition between the extension and the platform — especially in
+  split-cluster deployments.
+- Upstream model-access boundary is failing (5xx propagation).
+
+**Fix:**
+
+1. Run `kz-ext doctor` — the `Kamiwaza connection` check will fail in the
+   same direction.
+2. Check the platform status page / oncall channels.
+3. For transient failures, the runtime lib's `TokenRefreshMiddleware` already
+   retries once on 401 mid-stream. A double failure surfaces as
+   `PlatformOutageError`.
+4. Do **not** catch and swallow — let it bubble to the top-level error
+   handler so the exit code is preserved.
+
+## How the exit codes flow
+
+1. The runtime lib raises a `KamiwazaRuntimeError` subclass.
+2. If the exception bubbles into a `kz-ext` subcommand, the CLI's top-level
+   `_handle_exception` catches it and calls `exit_code_for(exc.class_name)`.
+3. The process exits with the canonical code from `exception_names.json`.
+
+CI and operators can key off the exit code alone:
+
+```bash
+kz-ext dev
+case $? in
+  0)  echo "OK" ;;
+  10) echo "Platform envelope missing — check Traefik" ;;
+  11) echo "Envelope shape wrong — check deployment context" ;;
+  12) echo "Cross-workroom access — extension logic bug" ;;
+  13) echo "Platform outage — retry later" ;;
+  23) echo "Cluster operator not ready — run kz-ext doctor" ;;
+  *)  echo "Generic failure ($?)" ;;
+esac
+```
+
+Exit code 23 (`CLUSTER_NOT_READY`) is a related but distinct failure — it
+comes from the `cluster_extension_readiness` probe in `kz-ext doctor`, not
+from a runtime-lib exception. See ENG-3888 / §4.2.8 B1a.
+
+## The single-source-of-truth invariant
+
+If you add a new exception class to the hierarchy, update
+`kamiwaza_extensions_lib/exception_names.json` in the same commit. The
+`test_doctor_hint_coverage.py` contract test will fail loudly if the JSON
+drifts from either the `ExitCode` enum or the `kz-ext doctor` output.
+
+## Design references
+
+- `§4.2.7 RuntimeLibExceptionHierarchy + IdentityExtractor + TokenRefreshMiddleware`
+- `§4.2.8 DoctorUACFailureHints + ExitCodeMap`
+- `§5 Q6 Security implications and auth interactions`
+- Linear: [ENG-3885](https://linear.app/kamiwaza/issue/ENG-3885)
--- a/kamiwaza_extensions/cli.py
+++ b/kamiwaza_extensions/cli.py
@@ -8,6 +8,8 @@ import typer
 from rich.console import Console
 
 from kamiwaza_extensions import __version__
+from kamiwaza_extensions.exit_codes import exit_code_for
+from kamiwaza_extensions_lib.errors import KamiwazaRuntimeError
 
 app = typer.Typer(
     name="kz-ext",
@@ -84,20 +86,16 @@ def _handle_exception(exc: Exception) -> None:
     # UAC-9d runtime-lib exceptions map to their canonical exit codes
     # (§4.2.7 / §4.2.8). Keep this branch above the generic fallback so
     # the exit code carries meaning for extension authors and CI.
-    try:
-        from kamiwaza_extensions.exit_codes import exit_code_for
-        from kamiwaza_extensions_lib.errors import KamiwazaRuntimeError
-    except ImportError:
-        KamiwazaRuntimeError = None  # type: ignore[assignment]
-        exit_code_for = None  # type: ignore[assignment]
-
-    if KamiwazaRuntimeError is not None and isinstance(exc, KamiwazaRuntimeError):
+    if isinstance(exc, KamiwazaRuntimeError):
         console.print(f"[red]Error:[/red] {exc}")
         if _state.debug:
             console.print_exception()
         raise typer.Exit(code=int(exit_code_for(exc.class_name))) from exc
 
-    # Try to import Kamiwaza errors for nicer formatting
+    # Try to import Kamiwaza SDK errors for nicer formatting. This import
+    # stays lazy because kamiwaza_sdk is a sibling package and some install
+    # flavours (e.g. a stripped-down `kamiwaza-extensions-only` wheel) may
+    # ship without it.
     try:
         from kamiwaza_sdk.exceptions import KamiwazaError
     except ImportError:
--- a/kamiwaza_extensions/cli.py
+++ b/kamiwaza_extensions/cli.py
@@ -81,6 +81,22 @@ def run_with_error_handling(func):
 
 
 def _handle_exception(exc: Exception) -> None:
+    # UAC-9d runtime-lib exceptions map to their canonical exit codes
+    # (§4.2.7 / §4.2.8). Keep this branch above the generic fallback so
+    # the exit code carries meaning for extension authors and CI.
+    try:
+        from kamiwaza_extensions.exit_codes import exit_code_for
+        from kamiwaza_extensions_lib.errors import KamiwazaRuntimeError
+    except ImportError:
+        KamiwazaRuntimeError = None  # type: ignore[assignment]
+        exit_code_for = None  # type: ignore[assignment]
+
+    if KamiwazaRuntimeError is not None and isinstance(exc, KamiwazaRuntimeError):
+        console.print(f"[red]Error:[/red] {exc}")
+        if _state.debug:
+            console.print_exception()
+        raise typer.Exit(code=int(exit_code_for(exc.class_name))) from exc
+
     # Try to import Kamiwaza errors for nicer formatting
     try:
         from kamiwaza_sdk.exceptions import KamiwazaError
--- a/kamiwaza_extensions/cli.py
+++ b/kamiwaza_extensions/cli.py
@@ -133,7 +133,15 @@ def login(
     if dev:
         url = url or "https://kamiwaza.test/api"
         no_verify_ssl = True
-    run_login(url=url, api_key=api_key, name=name, list_connections=list_connections, use=use, no_verify_ssl=no_verify_ssl)
+    run_login(
+        url=url,
+        api_key=api_key,
+        name=name,
+        list_connections=list_connections,
+        use=use,
+        no_verify_ssl=no_verify_ssl,
+        verbose=_state.verbose,
+    )
 
 
 @app.command()
--- a/kamiwaza_extensions/commands/login.py
+++ b/kamiwaza_extensions/commands/login.py
@@ -68,22 +68,23 @@ def _capture_ui_roles(client, *, verbose: bool = False) -> set[str]:
     describe the PAT's scoped subset, defeating the comparison.
 
     Returns an empty set on any failure. When *verbose* is true, emits
-    a dim diagnostic so operators can distinguish "no downgrade
+    a diagnostic via ``typer.echo`` (same output channel as the
+    downgrade warning below) so operators can distinguish "no downgrade
     detected" from "we couldn't tell" (PR review High #3).
     """
     try:
         user = client.auth.get_current_user()
     except Exception as exc:  # noqa: BLE001 — best-effort diagnostic
         if verbose:
-            console.print(
-                f"[dim]Could not fetch UI role-set for B3 check: {type(exc).__name__}: {exc}[/dim]"
+            typer.echo(
+                f"Could not fetch UI role-set for B3 check: {type(exc).__name__}: {exc}"
             )
         return set()
     roles = getattr(user, "roles", None) or []
     if not isinstance(roles, list):
         if verbose:
-            console.print(
-                "[dim]Could not parse UI role-set for B3 check: roles claim was not a list[/dim]"
+            typer.echo(
+                "Could not parse UI role-set for B3 check: roles claim was not a list"
             )
         return set()
     return {str(r) for r in roles if isinstance(r, str)}
@@ -201,6 +202,7 @@ def _login_with_password(
     if not verify_ssl:
         os.environ["KAMIWAZA_VERIFY_SSL"] = "false"
 
+    pat_name = f"kz-ext-{name}"
     try:
         client = KamiwazaClient(base_url=url)
         client.authenticator = UserPasswordAuthenticator(
@@ -211,13 +213,7 @@ def _login_with_password(
         # the PAT. Post-mint the only credential we have is the scoped PAT,
         # which would reflect only its own roles (§4.8 B3).
         ui_roles = _capture_ui_roles(client, verbose=verbose)
-        pat_response = client.auth.create_pat(PATCreate(name=f"kz-ext-{name}"))
-        pat_token = pat_response.token
-        if not isinstance(pat_token, str) or not pat_token:
-            raise RuntimeError(
-                "Platform did not return a token for the new PAT. Try again, "
-                "or contact the platform team."
-            )
+        pat_response = client.auth.create_pat(PATCreate(name=pat_name))
     except Exception as exc:
         console.print(f"[red]Error:[/red] Authentication failed: {exc}")
         raise typer.Exit(code=1) from exc
@@ -227,6 +223,19 @@ def _login_with_password(
             os.environ.pop("KAMIWAZA_VERIFY_SSL", None)
         else:
             os.environ["KAMIWAZA_VERIFY_SSL"] = old_verify
+
+    # Validate the PAT response *outside* the auth catch-all — this branch
+    # represents a successful authentication followed by a PAT-mint failure,
+    # not an auth failure, and the user should see a different error class.
+    pat_token = pat_response.token
+    if not isinstance(pat_token, str) or not pat_token:
+        console.print(
+            f"[red]Error:[/red] Platform accepted credentials but returned no "
+            f"token for the new PAT '{pat_name}'. The PAT may have been "
+            f"created server-side; consider deleting it via the UI before "
+            f"retrying."
+        )
+        raise typer.Exit(code=1)
 
     # Store the PAT — auth already succeeded server-side at this point
     token = StoredToken(access_token=pat_token, refresh_token=None, expires_at=0.0)
--- a/kamiwaza_extensions/commands/login.py
+++ b/kamiwaza_extensions/commands/login.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import base64
+import json
 from typing import Optional
 
 import typer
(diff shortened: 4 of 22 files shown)


jonathanleff:
Added UAC-9d failure-classes runbook at `docs/services/extensions/uac-9d-failure-classes.md` (commit 5b396f5). Covers: per-class diagnosis + fix, code examples, exit-code keying for CI.

PR body's follow-ups list is now: **TS mirror** (separate PR) + **CHANGELOG** (no convention exists yet in this repo — release notes can reference the PR directly). Runbook will grow a TS-side appendix when the mirror lands.

jxstanford (changes requested):
# Review Report: feat(extensions): UAC-9d runtime-lib exception surface + CLI plumbing (ENG-3885)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/76
**Branch**: `feature/eng-3885-uac9d-runtime-exceptions` → `release/0.12.1`
**Engines**: Claude Code, Codex, Gemini (3/4 available)

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| 1 | `kamiwaza_extensions_lib/session.py:85` | `/session` now returns `**identity.model_dump()`, which serializes the new `auth_token` field (the `X-Auth-Token` envelope value used at `client.py:127-131` as a Bearer credential) into the JSON body, exposing it to any frontend/XSS path. The previous hand-rolled response intentionally excluded it. The same change leaks `system_high` (security-classification flag) and `request_id`, which were not part of the prior public contract. Fix: `model_dump(exclude={"auth_token", "system_high", "request_id"})` or define an explicit `SessionResponse` model. | Claude Code, Codex | Sensitive Data Exposure / Broken Auth |

## High Priority Issues
| 2 | `kamiwaza_extensions/commands/login.py:56-57` | `_fetch_ui_roles()` calls `"{url}/auth/whoami"`, but the platform OpenAPI spec exposes only `/whoami` and `/auth/users/me` — there is no `/auth/whoami`. Non-200s are swallowed into `set()`, so the §4.8 B3 downgrade warning silently never fires. Point at `/auth/users/me` (which the SDK's `AuthService.get_current_user` already uses). | Codex | Correctness / Dead Feature |
| 3 | `kamiwaza_extensions/commands/login.py:199-201` | Even after fixing the endpoint, `_fetch_ui_roles(url, pat_token, ...)` authenticates the lookup with the **just-minted PAT**, so the returned roles describe the PAT itself. `pat_roles` and `ui_roles` will always match for the same credential, and a strict subset can never appear. The comparison needs to be made against the original password-login session token (or whatever credential was used before PAT mint), not the PAT being evaluated. | Codex | Logic Error |
| 4 | `kamiwaza_extensions_lib/identity.py:42` | `model_config = ConfigDict(extra="allow")` is unsound forward-compat: `Identity` is never parsed from a header dict via Pydantic — `identity_from_headers`/`extract_identity` build it with explicit kwargs — so `extra="allow"` only opens a footgun where any future `Identity(**raw_payload)` silently captures extra fields that then leak through `model_dump()` (compounds issue #1). Recommend `extra="forbid"` or `extra="ignore"`. | Claude Code | Architecture / Data Validation |
| 5 | `tests/unit/extensions_lib/test_identity.py:1224` | `test_require_auth_and_session_produce_matching_identity` only validates the `USE_AUTH=false` path (where `auth_token` is `None`), so it doesn't catch issue #1. Add a test that drives `/session` with `x-auth-token` set and asserts the JSON body does **not** contain `auth_token`/`system_high`/`request_id`. | Claude Code | Test Coverage |
| 6 | `kamiwaza_extensions/commands/login.py:84` | `if not missing or not pat_roles < ui_roles` is logically equivalent to `if not (pat_roles < ui_roles)` — when `pat_roles` is a strict subset, `missing = ui_roles - pat_roles` is necessarily non-empty by definition. The `not missing` branch is dead code. Simplify to `if not pat_roles < ui_roles: return`. | Claude Code | Code Quality |

| 7 | `tests/unit/extensions/test_doctor_hint_coverage.py:42` | `test_doctor_surfaces_hint_for_each_class` calls `DoctorChecker().run_all()`, which shells out to `docker --version`, `docker compose version`, `docker info`, makes HTTP calls, and walks cwd for `kamiwaza.json`. Slow (~30s of subprocess timeouts) and flaky on machines without Docker. Call `DoctorChecker()._uac_9d_reference_hints()` directly. | Claude Code | Test Quality |
| 8 | `kamiwaza_extensions/cli.py:87-92` | The `try/except ImportError` around imports of `kamiwaza_extensions.exit_codes` and `kamiwaza_extensions_lib.errors` adds clutter — both ship in this same package; if either import fails the whole CLI is non-functional anyway. Move imports to module top. | Claude Code | Code Quality |
| 9 | `kamiwaza_extensions/commands/login.py:86` | Bare `print()` diverges from the `console.print()` style used elsewhere in the module. Comment says it's for `capsys` — but the underlying problem is that the module-level `console = Console()` was instantiated before `capsys` started. Fix the test setup (inject `Console`, or use `Console(file=sys.stdout, force_terminal=False)`) rather than the style break. | Claude Code, Gemini | Code Consistency |
| 10 | `kamiwaza_extensions_lib/identity.py:106` | `extract_identity` doesn't strip whitespace-only header values: `lower.get(_HEADER_USER_ID) or None` keeps `"  "` truthy. Strip values: `user_id = (lower.get(_HEADER_USER_ID) or "").strip() or None`. | Claude Code | Data Validation |
| 11 | `docs/services/extensions/uac-9d-failure-classes.md:86-87` | Forward-reference to `kamiwaza-sdk/docs/extensions/non-sdk-flow.md` annotated `(when available)` — dangling pointer. Either land the doc in this PR or drop the reference. | Claude Code | Stale Reference |
| 12 | `docs/services/extensions/uac-9d-failure-classes.md:80` | "extensions no longer verify HMAC as of 2026-04-23" date assertion will rot. Replace with a link to the deprecation commit/PR or design doc §4.2.7. | Claude Code | Comment Rot |
| 13 | `docs/services/extensions/uac-9d-failure-classes.md:109-112` | Code example shows catching `OutOfEnvelopeAccessError` from `client.workrooms.get_resource(...)`, but the runtime lib does not raise this from any `get_resource` call (no HTTP 403 → `OutOfEnvelopeAccessError` mapper exists in this PR). Aspirational example — either land the mapper or mark as "future API". | Claude Code | Wrong Example |

## Test Coverage Assessment
- **Adequate**: `errors.py`, `exit_codes.py`, `identity.py` (Pydantic round-trip + strict extractor), `login.py` (PAT decode + warning logic in isolation).
- **Gaps**:
  - No test asserts the authenticated `/session` response *omits* `auth_token` (issue #1 surface).
  - `_fetch_ui_roles` not tested against a mocked HTTP layer — would have caught the wrong endpoint (issue #2) and the self-referential PAT lookup (issue #3).
  - `_handle_exception` UAC-9d branch not covered for an unknown `KamiwazaRuntimeError` subclass with an unmapped `class_name` (should fall back to `ExitCode.FAILURE = 1`).
  - No test asserts that constructing `Identity` from arbitrary kwargs (via `extra="allow"`) doesn't leak extras through `model_dump()`.

## Positive Observations
- **Single source of truth**: `exception_names.json` consumed by `exit_codes.py` and `doctor.py`, with a contract test (`test_doctor_hint_coverage.py`) failing loudly on drift — correct shape for a cross-language invariant.
- `pyproject.toml` adds `exception_names.json` to `package_data` so `importlib.resources` works after install.
- `_handle_exception` correctly orders the UAC-9d branch above the generic `KamiwazaError` fallback, preserving exit-code meaning.
- `anonymous_identity()` cleanly fixes the §4.8 P5 papercut by routing both `require_auth` and `/session` through one constructor.
- `_decode_pat_roles` is explicit that it doesn't verify the signature, with a docstring justifying why (already-authenticated context, diagnostic-only).
- Clean separation of strict (`extract_identity`) vs permissive (`identity_from_headers`) entry points, mapping cleanly to the UAC-9d contract.
- TS-8 parametrized exit-code mapping test concisely covers all four classes.

## Overall Verdict
**Risk Level**: High
**Verdict**: REQUEST_CHANGES
**Engine Agreement**: 3/4 engines contributed. Claude Code and Codex independently flagged the `/session` credential-disclosure regression (#1). Codex additionally found that the new B3 role-downgrade warning is non-functional in two independent ways (#2 wrong endpoint, #3 self-referential lookup) — both confirmed against the platform OpenAPI spec. Gemini approved without finding the security regression or the broken warning. Concur with REQUEST_CHANGES: the design is correct, but the PR ships a credential leak and a feature that cannot fire as written.

---

---
<details>
<summary>🔧 Engine Health (2/4 contributing — Moderate (2 engines contributed))</summary>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ❌ Failed (exit 0) |
| Gemini | ✅ OK |

**Engines contributing**: 2/4
</details>

jxstanford (dismissed):
# RE-REVIEW Report: feat(extensions): UAC-9d runtime-lib exception surface + CLI plumbing (ENG-3885)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/76
**Branch**: `feature/eng-3885-uac9d-runtime-exceptions` → `release/0.12.1`
**Engines**: Claude Code, Codex, Gemini (3/4 available)

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| 1 | `kamiwaza_extensions_lib/session.py:85` | Authenticated `/session` returns `{**identity.model_dump(), "expires_at": expires_at}`. Because `Identity` now carries `auth_token` (from `x-auth-token`), `system_high`, `workroom_role`, and `request_id`, the response leaks the platform's session token into JSON — readable by any frontend script and any XSS payload. The pre-PR hand-rolled response intentionally excluded these. Fix: switch to `model_dump(exclude={"auth_token", "system_high", "request_id"})` or define an explicit `SessionResponse` schema with only public fields. | Claude Code, Codex (agree) | Sensitive Data Exposure / Broken Auth |

## High Priority Issues
| 2 | `kamiwaza_extensions/commands/login.py:56-57` | `_fetch_ui_roles()` calls `f"{url}/auth/whoami"`, which the platform does not expose (real endpoints are `/whoami` and `/auth/users/me` per the OpenAPI spec). Non-200 responses are swallowed into `set()`, so the §4.8 B3 downgrade warning never fires in practice — the new diagnostic is effectively disabled. | Claude Code, Codex (agree) | Correctness / Dead Feature |
| 3 | `kamiwaza_extensions/commands/login.py:198-202` | Even with the right endpoint, `_fetch_ui_roles(url, pat_token, …)` authenticates with the **just-minted PAT**, so the response describes the PAT itself. `pat_roles` and `ui_roles` are guaranteed to match — a strict-subset can never appear and the warning is unreachable. The lookup needs to use the password-login session token, not the PAT under evaluation. | Claude Code | Logic Error |
| 4 | `kamiwaza_extensions_lib/identity.py:42` | `model_config = ConfigDict(extra="allow")` is unsound forward-compat: `Identity` is never parsed from a header dict via Pydantic — both `identity_from_headers` and `extract_identity` build it with explicit kwargs — so `extra="allow"` only opens a footgun where any future `Identity(**raw_payload)` silently captures and round-trips extras through `model_dump()` (compounds issue #1). Recommend `extra="forbid"` or `extra="ignore"`. | Claude Code | Architecture / Data Validation |
| 5 | `tests/unit/extensions_lib/test_identity.py:224` | `test_require_auth_and_session_produce_matching_identity` only validates the `USE_AUTH=false` path (where `auth_token` is `None`), so it cannot fail on issue #1. Add a test that drives `/session` with `x-auth-token` set and asserts the JSON body does NOT contain `auth_token`/`system_high`/`request_id`. | Claude Code | Test Coverage |
| 6 | `kamiwaza_extensions/commands/login.py:84` | `if not missing or not pat_roles < ui_roles:` — when `pat_roles` is a strict subset, `missing = ui_roles - pat_roles` is necessarily non-empty by definition, so the `not missing` branch is dead. Simplify to `if not pat_roles < ui_roles: return`. | Claude Code | Code Quality |

| 7 | `tests/unit/extensions/test_doctor_hint_coverage.py:42` | `test_doctor_surfaces_hint_for_each_class` calls `DoctorChecker().run_all()`, which shells out to `docker --version`/`docker compose version`/`docker info` and walks cwd for `kamiwaza.json`. Slow and flaky on Docker-less CI runners. Call `DoctorChecker()._uac_9d_reference_hints()` directly — that is the unit under test. | Claude Code | Test Quality |
| 8 | `kamiwaza_extensions/cli.py:87-92` | `try/except ImportError` around imports of `kamiwaza_extensions.exit_codes` and `kamiwaza_extensions_lib.errors` is dead defense — both ship in this distribution; if either import fails, the CLI is non-functional. Move to module top. | Claude Code | Code Quality |
| 9 | `kamiwaza_extensions/commands/login.py:86` | Bare `print()` diverges from the `console.print()` style used elsewhere. The "for capsys" comment papers over a test-setup problem (module-level `console = Console()` instantiated before `capsys` starts). Fix the test (inject Console / `Console(file=sys.stdout, force_terminal=False)`) rather than break the style. Also: warnings should go to stderr, not stdout. | Claude Code | Code Consistency |
| 10 | `kamiwaza_extensions_lib/identity.py:114-115` | `extract_identity` doesn't strip whitespace-only header values: `lower.get(_HEADER_USER_ID) or None` keeps `"  "` truthy, so a misconfigured Traefik that injects whitespace-only values bypasses the `MisboundAuthError` raise. Strip first: `(lower.get(_HEADER_USER_ID) or "").strip() or None`. | Claude Code | Data Validation |
| 11 | `docs/services/extensions/failure-classes.md:81-82` | Forward-reference to `kamiwaza-sdk/docs/extensions/non-sdk-flow.md` annotated `(when available)` — dangling pointer (file does not exist in the repo). Either land it in this PR or drop the reference. | Claude Code | Documentation |
| 12 | `docs/services/extensions/failure-classes.md:75` | `"extensions no longer verify HMAC as of 2026-04-23"` is a date-stamped assertion that will rot. Replace with a link to the deprecation commit/PR or design doc §4.2.7. | Claude Code | Documentation |
| 13 | `docs/services/extensions/failure-classes.md:111-122` | Code example shows catching `OutOfEnvelopeAccessError` from `client.workrooms.get_resource(...)`, but the runtime lib does not raise this from any `get_resource` call (no HTTP-403 → `OutOfEnvelopeAccessError` mapper exists in this PR). Aspirational — either land the mapper or mark the example "future API". | Claude Code | Documentation |

## Test Coverage Assessment
- **Adequate**: `errors.py`, `exit_codes.py`, `identity.py` (Pydantic round-trip + strict extractor), `login.py` (PAT decode + warning logic in isolation), parametrized exit-code mapping for the 4 UAC-9d classes.
- **Gaps**:
  - No assertion that the *authenticated* `/session` response omits `auth_token`/`system_high`/`request_id` (issue #1 surface — would have caught the credential leak).
  - `_fetch_ui_roles` not tested against a mocked HTTP layer — would have caught the wrong endpoint (#2) and the self-referential PAT lookup (#3).
  - `_handle_exception` UAC-9d branch not exercised for an unknown `KamiwazaRuntimeError` subclass with an unmapped `class_name` (should fall back to `ExitCode.FAILURE = 1`).
  - No test asserts that constructing `Identity` from arbitrary kwargs (via `extra="allow"`) doesn't leak extras through `model_dump()`.

## Positive Observations
- Single source of truth (`exception_names.json`) cleanly consumed by `exit_codes.py` and `doctor.py`, with a contract test that fails loudly on drift.
- `pyproject.toml` adds `exception_names.json` to `package_data` so `importlib.resources` works post-install.
- `_handle_exception` correctly orders the runtime-lib branch above the generic `KamiwazaError` fallback, preserving exit-code meaning.
- `anonymous_identity()` cleanly resolves the §4.8 P5 papercut by routing both `require_auth` and `/session` through one constructor.
- `_decode_pat_roles` is explicit (in code and docstring) about not verifying the JWT signature, justified by the already-authenticated context — appropriate diagnostic-only use.
- Clean strict (`extract_identity`) vs permissive (`identity_from_headers`) split that maps onto the UAC-9d contract.
- Parametrized exit-code mapping test concisely covers all four runtime-lib classes.
- High-quality `failure-classes.md` runbook with stable class names, exit codes, and doctor hints.

## Overall Verdict
**Risk Level**: High
**Verdict**: REQUEST_CHANGES
**Engine Agreement**: 3/4 engines contributed — Claude Code and Codex agree on the credential leak (#1) and broken `/auth/whoami` endpoint (#2); Gemini approved without flagging either, missing both. The design is correct and the diagnostics scaffolding is solid, but the PR ships a credential-disclosure regression in `/session` and a B3 warning that cannot fire as written.

---

---
<details>
<summary>🔧 Engine Health (2/4 contributing — Moderate (2 engines contributed))</summary>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ❌ Failed (exit 0) |
| Gemini | ✅ OK |

**Engines contributing**: 2/4
</details>

jonathanleff:
Thanks for the thorough review — all 13 items addressed across 4 commits. Summary:

**Critical (#1)** — `/session` credential leak: `4c9be6b`. Switched from `model_dump()` to an explicit `SESSION_PUBLIC_FIELDS` allowlist so `auth_token`/`system_high`/`request_id` can't cross the HTTP boundary. New test `test_authenticated_session_does_not_leak_sensitive_fields` pins the invariant. Also dropped `extra="allow"` from `Identity` per #4 — Pydantic default `"ignore"` is what we actually wanted.

**High (#2, #3, #6, #9)** — B3 warning was dead code: `4ee7608`. Correct analysis. The new implementation captures UI roles via `client.auth.get_current_user()` **before** `create_pat()`, using the SDK's password-session wrapper around `/auth/users/me` (which is already the right endpoint, no direct HTTP call needed). Removed `_fetch_ui_roles`; introduced `_capture_ui_roles(client)`. Simplified `_warn_if_roles_downgraded` (dropped the dead `not missing` branch). Switched from bare `print()` to `typer.echo()` so `CliRunner` captures through Click's normal stdout path.

**High (#5)** — test coverage: covered by #1's new test (authenticated `/session` asserting no leak).

**Medium (#7, #8, #10)** — cleanups: `7dc2180`. `test_doctor_surfaces_hint_for_each_class` now calls `_uac_9d_reference_hints()` directly (drops the 30s of docker subprocess timeouts). `cli.py` imports promoted to module top. `extract_identity` now treats whitespace-only values as empty.

**Medium (#11, #12, #13)** + coverage gap: `e1c0cc8`. Dropped the dangling `non-sdk-flow.md` reference; swapped the dated HMAC note for a stable pointer to design §4.2.7; re-cast the `OutOfEnvelopeAccessError` example to match what the runtime lib actually raises (with an explicit caveat that SDK 403s aren't auto-wrapped yet). Added the `_handle_exception` fallback test for unmapped `class_name` → `ExitCode.FAILURE=1`.

**Test suite:** 598 passing (up from 591, +7 new tests net; the old `_fetch_ui_roles` tests were replaced by `_capture_ui_roles` tests + the sensitive-fields guard + whitespace + fallback).

Ready for another pass when you have time.

m9e (changes requested):
# PR Review Report: feat(extensions): UAC-9d runtime-lib exception surface + CLI plumbing (ENG-3885)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/76
**Branch**: feature/eng-3885-uac9d-runtime-exceptions -> develop
**Scope**: 18 files, +1124/-39

Reviewers: Superpowers (SP), Claude (CL), Codex (CX). Kamiwaza errored (model `kimi-k2.6` not deployed) and Gemini CLI is not installed; both excluded from N. **N = 3 successful reviewers.**

---

## Critical Issues (Must Fix)

| 1 | `kamiwaza_extensions_lib/auth.py:48-54` | `require_auth()` still calls the permissive `get_identity()`, not the new strict `extract_identity()`. With `KAMIWAZA_USE_AUTH=true`, a request bearing only `X-User-Id` (missing `X-Workroom-Id`) is treated as authenticated and reaches protected handlers with `identity.workroom_id is None` — exactly the `misbound_auth` envelope this PR introduces a class for. Any endpoint depending on `Depends(require_auth)` will accept malformed platform envelopes and run downstream SDK/model calls without a workroom scope. The new exception type is defined but never raised on the auth path that the rest of the codebase actually uses. | CX | Security / Auth correctness |

| 1 | `kamiwaza_extensions_lib/auth.py:14-26` | `_FORWARD_HEADERS` was not extended for the newly-tracked envelope headers (`x-user-system-high`, `x-user-workroom-role`). `forward_auth_headers()` is the canonical helper for downstream service calls — silently dropping these means an extension cannot propagate workroom-role context or the high-side classification flag downstream. | SP, CL | Bug / API contract |
| 2 | `kamiwaza_extensions_lib/identity.py:53,102,138` | `auth_token` (a Bearer credential) now lives on the `Identity` model. The `/session` `SESSION_PUBLIC_FIELDS` allowlist correctly excludes it, but any caller that ever calls `identity.model_dump()` without `include=` (logs, metrics, exception payloads) will leak the credential. Either keep `auth_token` off `Identity` and read from headers when needed (matching `_session_expires_at`), or override `model_dump`/`__repr__` to mask `auth_token` and `system_high`. The existing leak-guard test only covers the `/session` exit. | SP | Security |
| 3 | `kamiwaza_extensions/commands/login.py:62-65` | `_capture_ui_roles` catches `Exception` and returns `set()` on any failure. With an empty UI role-set the B3 downgrade warning is silently suppressed — a transient 5xx in `auth.get_current_user()` masks the diagnostic for the very session where it would matter most. At minimum log under `--debug`/`--verbose` so operators can distinguish "no downgrade" from "we couldn't tell." | SP | Error handling |
| 4 | `kamiwaza_extensions/commands/login.py:200-205` | If `pat_response.token` is `None` (or non-string), `StoredToken(access_token=None, ...)` is silently stored and `_decode_pat_roles(None)` raises `AttributeError: 'NoneType' object has no attribute 'split'` — crashing `kz-ext login` *after* the connection has already been added. Guard with `isinstance(pat_token, str)` or short-circuit inside `_decode_pat_roles`. | CL | Robustness / NoneType crash |
| 5 | `kamiwaza_extensions_lib/identity.py:50,99,135` | `Identity.system_high` is typed `bool`, but the platform's `X-User-System-High` header carries a string classification (e.g. `"U"`, `"TS"` — see `kamiwaza_sdk/services/enclaves.py:86`). `_parse_bool("U")` returns `False`, silently dropping classification. Any extension that reads `identity.system_high` for trust decisions will misread classified requests as un-classified. Either type as `Optional[str]` or document/enforce the boolean coercion contract. | CL | Data validation / boundary semantics |
| 6 | `kamiwaza_extensions/commands/login.py:81` | `_warn_if_roles_downgraded` only fires when `pat_roles ⊊ ui_roles`. It misses overlapping-but-disjoint cases — e.g. `pat_roles={"a","b"}`, `ui_roles={"a","c"}` — where `"c"` is a real downgrade. Warn whenever `ui_roles - pat_roles` is non-empty and `ui_roles` is non-empty. | CL | Business logic correctness |

| 1 | `kamiwaza_extensions_lib/__init__.py` | New public surface (`KamiwazaRuntimeError`, `MisboundAuthError`, `UnexpectedContextError`, `OutOfEnvelopeAccessError`, `PlatformOutageError`, `extract_identity`, `anonymous_identity`) is not re-exported from the package root. Other public symbols (`Identity`, `require_auth`) are top-level imports, so this is inconsistent. Add to `__all__`. | SP, CL | API design |
| 2 | `kamiwaza_extensions/exit_codes.py:31-39` | `_exception_name_to_exit_code` does not validate that JSON `exit_code` values are members of the `ExitCode` IntEnum. JSON drift to an unmapped int (e.g. 14) → `ExitCode(code)` raises `ValueError` from inside `exit_code_for` during error handling. Add a one-time validation at module import or on first cache fill. | SP | Robustness |
| 3 | `kamiwaza_extensions/exit_codes.py:17-32` + contract test | The `ExitCode` enum hard-codes UAC-9d codes (10–13) alongside `exception_names.json`. No test fails if a new JSON entry is added without a matching `ExitCode` member. Add an assertion in the contract test that the enum is a superset of the JSON. | CL | Test coverage gap |
| 4 | `kamiwaza_extensions/doctor.py:21-29` & `exit_codes.py:32-39` | Both modules independently load + parse `exception_names.json`. Extract a single helper in `kamiwaza_extensions_lib.errors` (e.g. `iter_exception_classes()`) so future schema changes have one read site. | SP | DRY |
| 5 | `kamiwaza_extensions/doctor.py:75-91` | UAC-9d reference entries are emitted with `status="pass"` (green checkmark), so `kz-ext doctor` always shows 4 "passing" lines that are reference docs, not health checks. Use `status="info"` or render in a separate section so users don't conflate static reference with passing checks. | CL | UX / debugging ergonomics |
| 6 | `kamiwaza_extensions_lib/identity.py:34-55` | No `model_config` is set; `extra="ignore"` relies on the Pydantic v2 default. Make the contract explicit with `model_config = ConfigDict(extra="ignore")` so a future v2 default change cannot silently turn `Identity` into a leak surface. | SP, CL | Defensive programming |
| 7 | `kamiwaza_extensions/cli.py: _handle_exception` | Repeats `console.print + console.print_exception(if debug) + raise typer.Exit(code=...)` three times. Extract a small `_exit_with(exc, code)` helper. | SP | Refactoring |
| 8 | `kamiwaza_extensions/commands/login.py` | Role-warning helpers (`_decode_pat_roles`, `_capture_ui_roles`, `_warn_if_roles_downgraded`) are testable units stuck inside a CLI command file. Consider moving them into `commands/_role_check.py` (or `_pat.py`). | SP | Refactoring |
| 9 | `kamiwaza_extensions_lib/identity.py` | `identity_from_headers` and `extract_identity` share most projection logic. Extract a private `_identity_from_lower(lower)` so the strict path only adds the guard checks. | SP | Refactoring |

## Low Priority

| 1 | `kamiwaza_extensions/commands/login.py:36-49` | `_decode_pat_roles` parses an unverified JWT (clearly documented). Add a defensive cap on payload size before base64-decoding (e.g., 64 KiB) to prevent attacker-controlled `--api-key` from causing a memory blowup. | SP | Defensive programming |
| 2 | `kamiwaza_extensions_lib/session.py:60-62` | `_decode_jwt_exp` accepts string-form `exp` claims via `exp.isdigit()`. RFC 7519 requires `NumericDate` (number). More lenient than the spec — drop the string fallback or document why. (Pre-existing in touched file.) | SP | Spec compliance |
| 3 | `tests/unit/extensions_lib/test_identity.py:212-225` | `asyncio.get_event_loop().run_until_complete(...)` is deprecated in Python 3.10+ and raises `DeprecationWarning` on 3.12. Use `asyncio.run(...)` or `pytest.mark.asyncio` (which is already used in the rest of the file). | SP, CL | Future-compat / Test quality |
| 4 | `kamiwaza_extensions/commands/login.py:82-88` | The downgrade warning uses `typer.echo` (plain stdout) while surrounding success messages use `rich.Console.print` with color. Route through a styled console (`[yellow]Warning:[/yellow] ...`) for visual consistency. | CL | UX consistency |
| 5 | `kamiwaza_extensions_lib/identity.py:61-70` | `"Anonymous"` is the literal sentinel name for the anonymous identity. `is_authenticated=False` already disambiguates, but a frontend branching purely on `identity.name` would conflate the two. Document the requirement to check `is_authenticated`, or use a more obviously sentinel-y string (e.g. `"(anonymous)"`). | CL | Edge case / UX |

## Test Coverage Assessment

- **Coverage verdict**: Adequate for the new surface; gaps in CLI integration and edge cases.
- **Files with adequate coverage**: `errors.py`, `identity.py` (strict path), `exit_codes.py`, `session.py` (sensitive-field leak guard), `cli.py` (RuntimeLibException → exit code mapping), helper-level coverage in `commands/login.py`.
- **Files / scenarios missing tests**:
  - End-to-end CLI run that exercises `_login_with_password` and surfaces the role-downgrade warning (current `TestRoleSetWarning` only exercises helpers, not placement after PAT mint).
  - `_decode_pat_roles(None)` should not raise (regression for High issue #4).
  - Identity `system_high` behavior when header carries a non-boolean classification token (e.g. `"TS"`).
  - `require_auth()` raising `MisboundAuthError` when only `X-User-Id` is present (regression for the Critical issue).
  - Public export surface (`__init__.py` `__all__`) stability.
- **Suggested test cases**:
  - `test_login_emits_role_downgrade_warning` (CLI integration with mocked SDK).
  - `test_decode_pat_roles_handles_none_token`.
  - `test_require_auth_rejects_missing_workroom_id_under_use_auth`.
  - `test_identity_system_high_string_value` (or asserts the documented coercion).
  - Contract test extension: assert `ExitCode` membership covers all `exit_code` values in `exception_names.json`.

## Reviewer Agreement Matrix

| Category | Superpowers | Claude | Codex | Kamiwaza | Gemini | Consensus |
|----------|-------------|--------|-------|----------|--------|-----------|
| Security / Auth | High (auth_token leak surface, _FORWARD_HEADERS) | Lower-priority items | **P1** (require_auth bypasses extract_identity) | N/A | N/A | Strong concerns; CX surfaces a structural gap not seen by SP/CL |
| Robustness / Error handling | High (login.py exception swallowing) | High (None token AttributeError, system_high bool) | — | N/A | N/A | High agreement on login.py needing hardening |
| API contract | High (forward headers, __init__ exports) | Important (system_high typing, exports) | — | N/A | N/A | Agreement on incomplete envelope plumbing |
| Code quality | Medium (DRY, refactoring opportunities) | Medium (UX consistency) | — | N/A | N/A | Aligned suggestions |
| Tests | Low (asyncio deprecation) | Important (CLI integration gap, asyncio dep) | — | N/A | N/A | Coverage is solid; gaps noted |
| Docs / runbook | Positive | Positive (minor reference check) | — | N/A | N/A | Runbook praised |

## Reviewer Verdicts

| Superpowers | Comment (Risk Low; 3 High issues but no Critical) |
| Claude | Comment (no Critical; 5 Important issues; High confidence) |
| Codex | Request Changes (P1 finding on `require_auth` envelope enforcement) |
| Kamiwaza | N/A (model `kimi-k2.6` not deployed — error) |
| Gemini | N/A (CLI not installed) |

**Vote Tally**: Approve: 0 | Comment: 2 | Request Changes: 1 (out of 3 successful reviewers)

- **Single source of truth**: `exception_names.json` + `test_doctor_hint_coverage.py` contract test ties the JSON, `ExitCode` enum, and `kz-ext doctor` output together. The contract test calls `_uac_9d_reference_hints()` directly to dodge subprocess timeouts — thoughtful test design.
- **Allowlist over denylist**: `SESSION_PUBLIC_FIELDS` defaults new `Identity` fields to private — the safe failure mode for a public HTTP boundary. The explicit `test_authenticated_session_does_not_leak_sensitive_fields` regression guard names `auth_token`, `system_high`, `request_id` directly.
- **B3 fix correctness**: UI roles are captured against the password session **before** PAT mint; inline comments clearly call out the self-reference trap.
- **Strict vs permissive split**: `extract_identity` (strict, raises `MisboundAuthError`) vs `identity_from_headers` (permissive) is a clean dual-API; docstrings explain intent and the *why*.
- **Anonymous identity unification**: `anonymous_identity()` makes `/session` and `require_auth()` byte-identical under `USE_AUTH=false`; `test_require_auth_and_session_produce_matching_identity` enforces the property directly.
- **Defense in `_handle_exception`**: UAC-9d branch sits above the `kamiwaza_sdk` lazy import, so SDK absence cannot accidentally swallow runtime-lib exit codes — the lazy-import comment explains why.
- **Defensive `_decode_pat_roles`**: documented no-signature-verification rationale, filters non-string roles, tolerates malformed JWTs.
- **Unknown-class fallback**: `exit_code_for` returns `ExitCode.FAILURE` for unknown class names with a regression test, so third-party `KamiwazaRuntimeError` subclasses can't crash the CLI.
- **Packaging**: `pyproject.toml` correctly adds `exception_names.json` to `package_data` so `importlib.resources` works after install.
- **Excellent runbook**: `failure-classes.md` covers causes, fixes, code examples, and the related-but-distinct exit code 23.
- **Spec traceability**: inline references to §4.2.7, §4.2.8, §4.8 P5/B3 make the PR easy to audit against the design.

## Overall Verdict

**Risk Level**: Medium

**Reviewer Consensus**: 1/3 voted Request Changes; 2/3 voted Comment. The synthesized report retains a Critical finding (CX P1: `require_auth` does not enforce the strict envelope contract that this PR introduces) and 6 High-priority findings, so per the workflow's rule (1) — "Request Changes if any Critical or High-priority finding remains" — the consolidated recommendation is **Request Changes**. The PR's design and tests are solid; the gating concern is that the new strict surface is defined but not actually wired into `require_auth`, plus a `None`-token CLI crash and a `system_high` typing mismatch that risks dropping classification data.

---

Legend: SP=Superpowers, CL=Claude, CX=Codex, KZ=Kamiwaza, GM=Gemini

jonathanleff:
Critical + all 6 high issues addressed across 4 commits. Summary:

**Critical** — `require_auth` uses strict `extract_identity` (`ff889a1`)
The pre-fix path called the permissive `get_identity()` and only checked `is_authenticated`, so a request with `X-User-Id` but no `X-Workroom-Id` reached protected handlers with `workroom_id=None` — exactly the malformed envelope the new `MisboundAuthError` exists to catch. Under `USE_AUTH=true` we now route through `extract_identity`, translating `MisboundAuthError` → HTTP 401 with the canonical detail. `USE_AUTH=false` keeps the permissive flow so local dev still works without a platform-issued envelope. New tests pin both the rejection path and the dev-mode bypass.

**High #1** — `_FORWARD_HEADERS` extended (`47ac532`)
Added `x-user-system-high` and `x-user-workroom-role` to the forward set so downstream Kamiwaza service calls can re-establish the caller's classification and workroom role. Comment on the set now ties it to the Identity reads so future envelope additions get matched on both sides in one place.

**High #2** — `auth_token` removed from Identity entirely (`9954338`)
You're right that `model_dump(include=…)` covered `/session` but not the dozen other paths (logs, metrics, exception payloads). Cleanest fix is to keep the bearer credential off the model: consumers that need it (TokenRefreshMiddleware, `/session` expiry decoder) read `request.headers["x-auth-token"]` directly — same pattern as `_session_expires_at`. Updated tests assert `auth_token` is absent from `model_dump()`.

**High #3** — `_capture_ui_roles` visible diagnostic (`4a51ed5`)
`--verbose` now surfaces the exception type and message when the role-set capture fails, so operators distinguish "no downgrade" from "we couldn't tell." Default flow stays silent. Wired through `run_login` from `_state.verbose`.

**High #4** — `pat_response.token` validation + None-guard (`4a51ed5`)
Validate `pat_token` immediately after `create_pat` (raise typer.Exit(1) with a fixable hint if missing), and add a defensive type-guard in `_decode_pat_roles` so a non-string input returns `set()` instead of crashing. Also a 64 KiB payload cap on the JWT body (PR Low #1).

**High #5** — `system_high: Optional[str]` (`9954338`)
You're right that `_parse_bool("U")` returned `False` and silently dropped classification. Now stored as the stripped header string ("U", "TS", etc.), matching `kamiwaza_sdk/services/enclaves.py`. Test pins round-trip for 4 markers.

**High #6** — warning fires on overlap-but-disjoint (`4a51ed5`)
`if not pat_roles < ui_roles` was too narrow. New condition: any non-empty `ui_roles - pat_roles` triggers the warning, so `pat={a,b}, ui={a,c}` correctly flags `c` as a downgrade.

Also folded in two related Mediums while I was in the same code (Medium #6: explicit `model_config = ConfigDict(extra="ignore")` so a future Pydantic v2 default change can't silently reopen the leak surface; Medium #9: extracted `_project_identity_fields` shared between strict/permissive paths). The other Mediums (DRY between exit_codes/doctor JSON loaders, `_handle_exception` helper extraction, command-file reorganization) I'd defer to a follow-up unless you want them in this PR.

**Test suite:** 628 unit passing (+30 net new, including the 5 new tests in this round).

Branch is at `4a51ed5`. Ready for another pass.

m9e (dismissed):
# PR Re-Review Report: feat(extensions): UAC-9d runtime-lib exception surface + CLI plumbing (ENG-3885)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/76
**Branch**: `feature/eng-3885-uac9d-runtime-exceptions` → `develop`
**Scope**: 19 files, +1357/-58
**Re-review delta**: 4 commits since the previous synthesized review (2026-04-28T15:14:50Z)
- `9954338` — drop `auth_token` from Identity; `system_high` is a string
- `ff889a1` — `require_auth` uses strict `extract_identity`
- `47ac532` — forward `x-user-system-high` and `x-user-workroom-role`
- `4a51ed5` — login command robustness + visible diagnostic + broader warning

Reviewers: Superpowers (SP), Claude (CL), Codex (CX). Kamiwaza errored (model `kimi-k2.6` not deployed); Gemini CLI is not installed; both excluded from N. **N = 3 successful reviewers.**

---

## Prior Findings Verification

| # | Prior Severity | Issue (short) | Status | Evidence |
|---|----------------|---------------|--------|----------|
| C1 | Critical | `require_auth()` uses permissive `get_identity` instead of strict `extract_identity` | **Resolved** (SP, CL) | `kamiwaza_extensions_lib/auth.py:63-71` — `USE_AUTH=true` branch now calls `extract_identity(dict(request.headers))` and translates `MisboundAuthError → HTTP 401`. Regression tests at `tests/unit/extensions_lib/test_auth.py:120-146` (`test_missing_workroom_id_rejected_under_strict_auth`, whitespace variant). Local-dev bypass preserved + tested. |
| H1 | High | `_FORWARD_HEADERS` missing `x-user-system-high`, `x-user-workroom-role` | **Resolved** (SP, CL) | `auth.py:28-29` adds both headers; comment ties the set to Identity reads. New test `test_forwards_classification_and_workroom_role`. |
| H2 | High | `auth_token` on Identity is a leak vector | **Resolved** (SP, CL) | `identity.py` removes the field and `_HEADER_AUTH_TOKEN` entirely; `session.py:67-81` reads `x-auth-token` from request headers instead. `test_model_dump_round_trip` asserts `"auth_token" not in dumped`. Right call — eliminates the leak surface unconditionally. |
| H3 | High | `_capture_ui_roles` silently swallows exceptions | **Resolved** (SP, CL) | `commands/login.py:63-89` adds `verbose` plumbing; emits `[dim]…[/dim]` diagnostics for both exception path and "roles wasn't a list" path. Wired through `cli.py:143 → run_login → _login_with_password`. Tests `test_verbose_emits_diagnostic_on_sdk_exception` + quiet-mode counterpart. |
| H4 | High | PAT `None` token AttributeError after connection added | **Resolved** (SP, CL) | Two-layer defense: `_decode_pat_roles` rejects non-strings (`login.py:43-44`); `_login_with_password` validates `pat_token` immediately after `create_pat` (`login.py:215-220`) and raises a typed error **before** `mgr.add_connection`. |
| H5 | High | `system_high` typed `bool` drops classification | **Resolved** (SP, CL) | `identity.py:65` — `system_high: Optional[str] = None`; `_parse_bool` removed; round-trip test for `("U","TS","S","C")` at `test_identity.py:190-197`. |
| H6 | High | `_warn_if_roles_downgraded` misses overlap-disjoint | **Resolved** (SP, CL) | `commands/login.py:92-114` — condition is `if not ui_roles: return; missing = ui_roles - pat_roles; if not missing: return`. `test_warns_on_overlapping_disjoint` exercises `pat={a,b}, ui={a,c}` and asserts `['c']` is the diff (not full UI set). |
| M6 | Medium | No explicit `model_config` on Identity | **Resolved** (CL) | `identity.py:59` — `model_config = ConfigDict(extra="ignore")`. Test `test_unknown_kwargs_dropped_dont_leak_via_model_dump`. |
| M9 | Medium | Identity-projection DRY between strict/permissive | **Resolved** (CL) | `identity.py:101-114` — shared `_project_identity_fields(lower)` used by both `identity_from_headers` and `extract_identity`. |
| L1 | Low | JWT payload-size cap | **Resolved** (CL) | `_PAT_PAYLOAD_MAX_BYTES = 64 * 1024` enforced in `_decode_pat_roles`; new `test_returns_empty_for_oversized_payload`. |
| M1 | Medium | New public surface not re-exported | **Deferred** | `kamiwaza_extensions_lib/__init__.py` not touched; author explicitly deferred Mediums to follow-up. |
| M2-M5, M7-M8 | Medium | exit_codes JSON validation, ExitCode⊇JSON contract test, JSON-loader DRY, doctor `status="info"`, `_handle_exception` repetition, move helpers out of CLI file | **Deferred** | Not in delta; author deferred. |
| L2-L5 | Low | `_decode_jwt_exp` string `exp`, asyncio deprecation in tests, downgrade-warning Rich consistency, `"Anonymous"` literal | **Not addressed** | Carried over. Asyncio deprecation still at `tests/unit/extensions_lib/test_identity.py:241,267`. |

**Verdict on prior findings**: Critical resolved. All 6 Highs resolved with named regression tests. 3/9 Mediums resolved; remainder explicitly deferred. Author was thorough and responsive.

---

## New Issues Introduced by Delta

### High Priority

| 1 | `kamiwaza_extensions_lib/session.py:99-109` | Under `KAMIWAZA_USE_AUTH=true`, `/session` still uses the permissive `get_identity()` path, while `require_auth` now uses strict `extract_identity()`. A request with `X-User-Id` but missing `X-Workroom-Id` is reported by `/session` as `is_authenticated=true` (because `Identity.name` is set), yet any protected endpoint returns 401 from `MisboundAuthError`. The frontend sees a logged-in session; subsequent API calls all fail. This is the same logical inconsistency the prior Critical fixed in the other direction — the strict-mode contract introduced by this PR should hold uniformly across the auth surface, not just on `require_auth`. Fix: under `USE_AUTH=true`, route `/session` through `extract_identity` (mirroring `require_auth`), or have it reject malformed envelopes with the same 401 behavior. | CX (single-source) | Auth correctness / API consistency |

### Medium Priority

| 1 | `kamiwaza_extensions_lib/auth.py:71` | `HTTPException(status_code=401, detail=str(exc))` propagates the raw `MisboundAuthError` text (e.g., "Required envelope header X-Workroom-Id missing or empty") to the HTTP client. The runbook (`docs/services/extensions/failure-classes.md`) deliberately scrubs to `"Authentication required"`. Suggest `detail={"error": {"class": exc.class_name, "detail": "Authentication required"}}` (or similar canonical shape) and log the full message server-side. | SP (Low), CX (P2) | Information disclosure / API consistency |
| 2 | `kamiwaza_extensions/commands/login.py:217-220` | `RuntimeError("Platform did not return a token...")` is raised inside the surrounding `except Exception as exc` block, which re-renders it as `"Authentication failed: ..."`. The actual failure is PAT minting, not authentication — the user sees a misleading category. Consider `console.print(...)` + `raise typer.Exit(code=1)` outside the catch-all so the prefix isn't reapplied. Also: a successful `create_pat` with a missing token leaves an orphan PAT on the platform; the hint should mention the PAT name (`kz-ext-{name}`) or recommend manual cleanup. | SP, CL | Error UX / cleanup |
| 3 | `kamiwaza_extensions_lib/__init__.py:20-32` | Carryover Medium #1 — new public surface (`KamiwazaRuntimeError`, the 4 UAC-9d subclasses, `extract_identity`, `anonymous_identity`) still not in `__all__`, while `Identity` and `require_auth` are. Author deferred to a follow-up. Flagging that the deferral remains. | SP, CL | API design (deferred) |

### Low Priority

| 1 | `tests/unit/extensions_lib/test_auth.py:111-118` | `test_unauthenticated_request_raises_401` does not `monkeypatch.setenv("KAMIWAZA_USE_AUTH", "true")`. It relies on the `AuthConfig.from_env()` default — flake-prone if a prior test leaks env state. Add the explicit setenv. | CL | Test isolation |
| 2 | `kamiwaza_extensions_lib/auth.py:69` | `dict(request.headers)` materializes a full headers copy on every authenticated request to satisfy `extract_identity`'s `dict[str, str]` signature. Consider widening to `Mapping[str, str]` to skip the copy on hot paths. Micro-optimization. | SP, CL | Performance nit |
| 3 | `kamiwaza_extensions/commands/login.py:76-80` | Verbose diagnostic uses `console.print` (Rich) while the role-downgrade warning at `:110-114` uses `typer.echo`. Inconsistent output channel; tests pass via `capsys` today but this is a wart. | CL | Consistency |
| 4 | `tests/unit/extensions_lib/test_identity.py:241,267` | Carryover — `asyncio.get_event_loop().run_until_complete(...)` raises `DeprecationWarning` on Python 3.12. Use `asyncio.run(...)`. | SP, CL | Future-compat (still open) |

### Suggested New Tests

- End-to-end `_login_with_password` integration where a mocked SDK returns `pat_response.token = None`, asserting exit code 1 + the user-facing hint (covers the new RuntimeError branch end-to-end, not just `_decode_pat_roles(None)`).
- `test_require_auth_strict_path_validates_x_user_id` — symmetry with the workroom-id rejection test (currently no regression test for missing `X-User-Id` under USE_AUTH=true).
- `test_session_consistent_with_require_auth_under_strict_envelope` — pin the proposed fix for the High-priority `/session` finding above.

---

## Test Coverage Assessment (delta)

- **Coverage verdict**: Adequate for the fixes. Every prior High has a named regression test that would catch a re-occurrence.
- **Files with strong regression coverage**: `test_auth.py` (Critical + H1), `test_identity.py` (H2, H5, M6), `test_login.py` (H3, H4, H6, L1).
- **Gap**: no integration-level test that drives `_login_with_password` end-to-end with a `None`-token mock; no test pinning `/session ↔ require_auth` envelope-validation symmetry.

---

## Reviewer Agreement Matrix

| Category | Superpowers | Claude | Codex | Kamiwaza | Gemini | Consensus |
|----------|-------------|--------|-------|----------|--------|-----------|
| Prior Critical (`require_auth` strict) | Resolved | Resolved | (delta-only review didn't tick off priors but offered no objection) | N/A | N/A | Strong: resolved |
| Prior Highs (6/6) | All resolved | All resolved | (no objection) | N/A | N/A | Strong: all resolved |
| New: `/session` ↔ `require_auth` split-brain | — | — | **P2 / Must Fix** | N/A | N/A | Single-source High; structurally analogous to the prior Critical |
| New: 401 detail leakage | Low (info disclosure) | — | **P2 / Must Fix** | N/A | N/A | 2/3 flag; severity disputed (SP=Low, CX=High); synthesized as Medium |
| Test coverage of the fixes | Strong | Strong | (silent) | N/A | N/A | Strong |
| Polish (deferred Mediums) | Acknowledged | Acknowledged | — | N/A | N/A | Reasonable to defer |

## Reviewer Verdicts

| Reviewer | Individual Verdict | Basis |
|----------|--------------------|-------|
| Superpowers | Approve | "Risk Level: Low. All Critical and High findings resolved with regression tests; retained items are Low/Medium polish." |
| Claude | Approve | "Confidence: High. All Critical and High issues from prior review fixed in code AND covered by named regression tests; deferred Mediums are reasonable." |
| Codex | Request Changes | Two P2 findings ("should be fixed before considering the patch correct") — `/session` permissive vs strict-mode split-brain; raw `MisboundAuthError` text in 401 body. |
| Kamiwaza | N/A | Model `kimi-k2.6` not deployed (excluded from N) |
| Gemini | N/A | CLI not installed (excluded from N) |

**Vote Tally**: Approve: 2 | Comment: 0 | Request Changes: 1 (out of 3 successful reviewers)

---

## Positive Observations

- **Critical fix is structurally correct**: `USE_AUTH=true` now routes through `extract_identity`; `MisboundAuthError → 401` translation at the dependency boundary keeps the runtime-lib hierarchy consistent with HTTP semantics without forcing every handler to catch.
- **`auth_token` removed entirely** rather than masked — eliminates the leak surface from logs/metrics/exception payloads in one stroke. Module docstring documents the rationale (`identity.py:12-17`).
- **Symmetric defensive depth** on the PAT path: `isinstance(pat_token, str) and pat_token` at the call site, type-guarded `_decode_pat_roles`, and a 64 KiB payload cap with its own regression test. Each independently sufficient.
- **Regression tests track 1:1 with prior findings** — every High has at least one named test that would catch a regression to the prior buggy behavior.
- **`_project_identity_fields` extraction** simultaneously fixes the Medium #9 DRY concern and the High #5 `_parse_bool("U") → False` data-validation bug (both paths now share `_stripped`).
- **Explicit `model_config = ConfigDict(extra="ignore")`** addresses defense-in-depth without prompting.
- **Verbose diagnostic distinguishes "exception type" from "shape error"** with two separate paths — operators get actionable info.
- **Author replies** are precise: clear rationale for each fix, explicit deferral of polish Mediums, no hand-waving.

---

## Overall Verdict

**Risk Level**: Medium

**Reviewer Consensus**: 2/3 voted Approve; 1/3 (Codex) voted Request Changes. The synthesized report retains 1 High-priority finding (the `/session` ↔ `require_auth` strict-mode inconsistency — same logical bug class as the prior Critical, just on a different endpoint) and 1 Medium where SP and CX agree on the issue with disputed severity (raw `MisboundAuthError` text in 401 body). Per the workflow's rule (1) — *Request Changes if any High-priority finding remains* — the consolidated recommendation is **Request Changes**.

The author's response to the prior round was excellent: 1 Critical + 6 Highs resolved with named regression tests, 3 Mediums opportunistically resolved, remaining Mediums explicitly deferred. The two new findings flow naturally from the strict-mode wiring this PR introduces and should be a quick follow-up commit, not a re-architecture.

---

Legend: SP=Superpowers, CL=Claude, CX=Codex, KZ=Kamiwaza, GM=Gemini

jonathanleff:
Pushed follow-up fixes through `52f8a47`.

Addressed from the re-review:
- `/session` now mirrors strict-envelope auth behavior and avoids the authenticated/protected-call split-brain.
- `require_auth` returns a scrubbed 401 detail while logging the raw `MisboundAuthError` server-side.
- PAT mint responses with no token now surface as a distinct PAT-mint failure, do not store a bad connection, and mention the `kz-ext-<name>` cleanup target.
- Runtime-lib UAC-9d public surface is now re-exported from `kamiwaza_extensions_lib`.
- Identity parsers accept `Mapping[str, str]`, so FastAPI request headers are passed directly without copying.
- Deprecated `asyncio.get_event_loop()` test usage in the touched runtime-lib tests was replaced with `asyncio.run()`.

Verification:
- `uv run pytest tests/unit/extensions_lib/test_auth.py tests/unit/extensions_lib/test_session.py tests/unit/extensions_lib/test_identity.py tests/unit/extensions_lib/test_errors.py tests/unit/extensions_lib/test_client.py tests/unit/extensions/test_login.py` -> 105 passed
- `make test` -> 788 passed, 8 skipped, 172 deselected
- Targeted `ruff`, `black --check`, and `isort --check-only` on touched files -> passed
- `git diff --check` -> passed

Note: repo-wide `make lint` is still blocked by existing unrelated ruff issues under `kamiwaza_sdk/`; the touched files pass targeted lint/format checks.

kbot-ships (approved):
_Reviewing head `52f8a47` against `develop`._

## Bottom line

This PR introduces a well-designed runtime-lib exception hierarchy (4 canonical subclasses of `KamiwazaRuntimeError`), wires them into CLI exit codes via a single-source-of-truth JSON file, adds strict envelope validation in `extract_identity()` and `require_auth()`, fixes the "Hello, None" anonymous identity papercut, and adds a role-downgrade warning in the login command. The architecture is clean — the allowlist-based `/session` payload, the scrubbed 401 detail, the contract tests tying JSON ↔ enum ↔ doctor hints together. The security posture is improved: `require_auth` now rejects partial envelopes (user-id present, workroom-id missing) that previously sailed through. Two pre-existing tests now pass for the wrong reason and should be fixed.

**Verdict: APPROVE** — the issues below are worth fixing but none are blocking.

## Findings

### 1. Two pre-existing tests no longer test what they claim **(worth fixing)**

`tests/unit/extensions_lib/test_session.py:160` — `test_invalid_token_omits_expiry` sends `x-user-id: usr-123` with no `x-workroom-id`, under `use_auth="true"` (the `_make_app` default). Post-PR, `extract_identity` raises `MisboundAuthError` and the session endpoint returns the logged-out shape with `expires_at: None`. The test still passes but it's testing "malformed envelope → logged out" not "invalid JWT → no expiry".

Same issue at line 173: `test_anonymous_session_ignores_bearer_expiry` sends an `authorization` header with no `x-user-id`. It passes because `MisboundAuthError` fires, not because the anonymous-session logic explicitly ignores bearer expiry.

Fix: add `x-workroom-id` to the `test_invalid_token_omits_expiry` headers so it exercises the JWT-parsing path. For the anonymous test, switch to `use_auth="false"` if the intent is to test local-dev anonymous behavior, or add both identity headers if testing that an authenticated session with a malformed token omits expiry.

### 2. `_decode_pat_roles` role filtering is a no-op **(nit)**

`kamiwaza_extensions/commands/login.py:60` — `{str(r) for r in roles if isinstance(r, str)}`: the `str(r)` call on values already confirmed as `isinstance(r, str)` is redundant. The filter is doing the useful work; the conversion is dead code. Minor, no functional impact.

### 3. CI Ruff failure may be unrelated **(worth investigating)**

The PR description notes 1 CI check failing (Ruff Lint). Without the Ruff output it's hard to tell if it's a flaky base-branch issue or something introduced here. The code looks clean on inspection — no obvious lint violations in the changed files. Worth checking the CI logs to confirm it's not a blocker introduced by this PR.

## Test gaps

The test suite is thorough. The contract test (`test_doctor_hint_coverage.py`) enforcing the JSON ↔ ExitCode ↔ doctor invariant is particularly good. One specific gap:

- **`UnexpectedContextError`, `OutOfEnvelopeAccessError`, `PlatformOutageError` are defined but never raised** anywhere in the runtime-lib code in this PR. Only `MisboundAuthError` is raised by `extract_identity`. The other three exist as an API contract for extension authors and future runtime-lib code. This is fine for Phase A, but there are no integration-level tests showing how these exceptions would flow from a real handler → CLI exit code. The unit tests cover the plumbing mechanically (raising each class manually and checking exit codes), which is sufficient for now.

## What's good

- The single-source-of-truth pattern (`exception_names.json` → exit codes, doctor hints, and error classes) is well-executed and the contract test enforces it.
- The `SESSION_PUBLIC_FIELDS` allowlist in `session.py` is the right security posture — new Identity fields default to private.
- Scrubbing the `MisboundAuthError` detail from the 401 response body while logging it server-side and putting the class name in `WWW-Authenticate` is a textbook approach.
- The `Identity` conversion from dataclass to Pydantic with `extra="ignore"` is defensive and forward-compatible.
- The `_project_identity_fields` shared between permissive and strict parsers eliminates projection drift.
- Test coverage is comprehensive: 149-line error test, identity round-trip, session leak guards, PAT role decoding edge cases, CLI exit-code mapping including unknown subclasses.

## Verdict

**Verdict**: APPROVE

Solid exception hierarchy and auth-boundary tightening with strong test coverage and good security defaults. The two stale tests (finding #1) should be fixed in a follow-up to avoid false confidence, but they don't represent a regression — they pass safely.
---

PR #86, opened 2026-04-29
feat(extensions): D210 M2 Phase B — kz-ext update + runtime-lib parity + non-SDK contract (7 issues)

jonathanleff:
## Summary

Lands the SDK-side of M2 milestone *New surface — update + non-SDK + Go reference (Phase B)* for [D210 Extension Developer Experience](https://linear.app/kamiwaza/project/extension-developer-experience-d7a40baac83b). 7 issues, 9 commits — kept as separate commits for per-issue review. Three M2 issues remain deferred: ENG-3894 (Go reference + CI), ENG-3329 (Makefile deprecation, cross-repo), ENG-3896 (ZeroDowntimePATCH platform test, cross-repo) — tracked in Linear as Backlog with planned follow-up PRs.

### Issues closed (7)

| # | Issue | What | Files |
|---|-------|------|-------|
| 1 | **ENG-3891** (High) | NonSDKFlowDoc + CanonicalTestVectors. Single canonical contract for non-Py/TS extension authors (`docs/extensions/non-sdk-flow.md` 8-section spec); JSON test vectors at `docs/extensions/non-sdk-flow/test-vectors.json` are consumed bit-identically by Py + TS test suites. Header-only per §4.4.2 revised — no HMAC. | new docs + vectors; `tests/unit/extensions/test_non_sdk_flow.py` (11 tests) |
| 2 | **ENG-3892** (High) | Py runtime-lib v0.3 + canonical-vector parity. M1 already shipped `extract_identity`; M2 adds the cross-language parity contract via parametrized vector tests + version bump 0.1.0 → 0.3.0 + CHANGELOG. | `kamiwaza_extensions_lib/__init__.py`, `CHANGELOG.md`, `tests/unit/extensions_lib/test_identity_extractor_parity.py` (4 tests), `test_version.py` (2 tests) |
| 3 | **ENG-3893** (High) | TS runtime-lib v0.3 — `extractIdentityStrict` mirror of Py contract; new `errors.ts` (5 classes + className constants matching `exception_names.json`); 3 missing fields added to `Identity` (`systemHigh`, `workroomRole`, `requestId`). | `kamiwaza-ai-extensions-lib/src/server/{errors,identity,types,index}.ts`, `package.json` 0.2 → 0.3, CHANGELOG, 3 new test files (12 tests) |
| 4 | **ENG-3897** (Medium) | CompatibilityBundle. New `kamiwaza_extensions/compatibility.json` (per-CLI-version min/max ranges for Py + TS runtime libs); `doctor.py` extension parses requirements.txt + package.json declared specs, warns on out-of-range. | `compatibility.json`, `doctor.py` extension, pyproject package-data, `test_compatibility_bundle.py` (12 tests) |
| 5 | **ENG-3898** (Medium) | §4.8 CLI polish bundle: P1 (`kz-ext create --name foo` from non-empty cwd creates `foo/` and scaffolds inside), P3 (dev local hot-reload doc note), P7 (`.ai/extensions/` author-facing redirect), P8 (`kz-ext config publish-profile list` works as subcommand), B2 (Developer Guide "What runs where" diagram). | `scaffolder.py`, `commands/{create,config}.py`, new `docs/extensions/developer-guide.md`, `test_cli_polish_bundle.py` (8 tests) |
| 6 | **ENG-3890** (High, **headline**) | `kz-ext update` + `TemplateManifest` + migrations. The path by which existing scaffolds adopt new template fixes — including the M1 starter-backend hardening (ENG-3889/3919/3920) deliberately shipped without a 0.12.x point-release on the deferred-via-update plan. Five modes: default interactive, `--dry-run`, `--force`, `--non-interactive`, `--bootstrap`. JSON-merge for `kamiwaza.json`. Per-strategy file dispatch (`overwrite`, `preserve_if_modified`, `merge`). Per-shape MANIFESTS registry with invariant test enforcing every template file is classified. | new `template_manifest.py`, `commands/update.py`, `validators/metadata.py` extension, `docs/extensions/cli-reference/update.md`, 2 new test files (29 tests) |
| 7 | **ENG-3895** (High) | TokenRefreshMiddleware (Py + TS). 3-state contract: PRE_COMMIT_OK / PRE_COMMIT_401 / MID_STREAM_FAIL. Pre-commit 401 → refresh + retry once; double-401 → `PlatformOutageError`; post-commit failures translated to `StreamInterruptedError`. Per-caller refresh (no module-level Promise cache — would cross-leak users in multi-tenant Next.js). | new `kamiwaza_extensions_lib/middleware/token_refresh.py` + TS sibling, `errors.py` adds `StreamInterruptedError`, 9 new tests |

### Per-commit message bodies

Each commit's message details its own scope. From oldest to newest on this branch:
- `d5ac668` — ENG-3891
- `cb9e523` — ENG-3892
- `f450ff4` — ENG-3893
- `9670fbf` — ENG-3897
- `d6e61a0` — ENG-3898
- `765724c` — ENG-3890
- `38cc390` — ENG-3895
- `a734a20` — chore: lint cleanup
- `fcb2efb` — fix: verify-iteration-1 findings (C1+C2 critical, I1 security, I2/3/7-9/12 important)

## Acceptance Criteria Traceability

| AC | Criterion | Scenarios | Tests | Result |
|----|-----------|-----------|-------|--------|
| AC1 | `kz-ext update` all 5 modes × 3 shapes | TS-M2-1..12 | `test_update_command.py`, `test_template_manifest.py` | PASS (13 + 16 tests) |
| AC2 | Replay test green; MakefileInventoryReplay green | TS-M2-13..15 | (synthetic replay covered in unit tests; real-world Makefile replay deferred to ENG-3329) | **PARTIAL** — ENG-3329 deferred |
| AC3 | non-sdk-flow.md + vectors bit-identical Py/TS/Go | TS-M2-16..27 | `test_non_sdk_flow.py`, `identity-parity.test.ts`, `test_identity_extractor_parity.py` | PASS (Py + TS); Go follow-up via ENG-3894 |
| AC4 | IdentityExtractor parity Py↔TS; raises MisboundAuthError | TS-M2-18..27 | parity test files | PASS |
| AC5 | Go reference deploys + e2e CI green | TS-M2-28..30 | — | **DEFERRED** — ENG-3894 |
| AC6 | TokenRefreshMiddleware Py + TS | TS-M2-31..36 | `test_token_refresh.py` (Py 5), `token-refresh.test.ts` (TS 4) | PASS |
| AC7 | compatibility.json + Doctor warnings | TS-M2-37..40 | `test_compatibility_bundle.py` | PASS |
| AC8 | Makefile deprecation + replay | TS-M2-15, 41 | — | **DEFERRED** — ENG-3329 |
| AC9 | ZeroDowntimePATCH lands | TS-M2-42 | — | **DEFERRED** — ENG-3896 |
| AC10 | CLI polish P1/P3/P7/P8/B2 | TS-M2-43..46 | `test_cli_polish_bundle.py` | PASS |
| AC11 | `make test` passes; ≥40 net new; no regressions | suite-level | 866 → 1034 (+170 net new) | PASS |
| AC12 | All 10 M2 Linear issues to Done | project-level | 7/10 done; 3 deferred | **PARTIAL** |

**Coverage: 9/12 ACs fully met. 3 partial/deferred (AC2/5/8/9 — all trace to the 3 deferred Linear issues, scoped as follow-up.)**

## Documentation Status

| Category | Type | Path | Status | Action |
|----------|------|------|--------|--------|
| public | api-docs | `docs/extensions/non-sdk-flow.md` (new) | Updated | Created in PR |
| public | api-docs | `docs/extensions/non-sdk-flow/test-vectors.json` (new) | Updated | Created in PR |
| public | api-docs | `docs/extensions/cli-reference/update.md` (new) | Updated | Created in PR |
| public | api-docs | `docs/extensions/developer-guide.md` (new) | Updated | Created in PR |
| public | sdk-docs | `kamiwaza_extensions_lib/CHANGELOG.md` (new) | Updated | Py runtime lib v0.3 entry |
| public | sdk-docs | `kamiwaza-ai-extensions-lib/CHANGELOG.md` (new) | Updated | TS runtime lib v0.3 entry |
| engineering | design-doc | `kamiwaza-docs-engineering-internal/.../extension-dev-experience-system-design.md` | **Acknowledgment only** | Two minor implementation deviations — see "Design deviations" below |
| context | ai-context | `kamiwaza-sdk/CLAUDE.md` | Skipped | No new patterns; service-pattern conventions unchanged |
| context | readme | `kamiwaza-sdk/README.md` | Skipped | No top-level surface change |

### Design deviations (acknowledgment, not requiring design-doc edits)

1. **TemplateManifest hybrid model.** Design §4.2.4 specified a pure list of `TemplateOwnedFile` entries. Implementation adds `AUTHOR_OWNED_DENYLIST` per shape (for placeholder files like `frontend/src/app/page.tsx` that the template ships as examples but the author owns). The invariant test enforces partition: every template file is in MANIFESTS *or* on the denylist. Defensible adaptation; no API change.
2. **`kamiwaza.json` JSON-merge strategy.** Design §4.2.3 listed `merge` as "v1 = preserve_if_modified". Without an actual JSON merge, freshly-stamped `template_version` triggers a false-conflict on the first `kz-ext update` after `kz-ext create`. M2 implements a minimal field-merge for JSON files: rendered keys are inherited, author values win for collisions, `template_version` + `template_shape` are reset from manifest. Document in `cli-reference/update.md`.

## Test Plan

- [x] `make test` — **1034 passed, 8 skipped** (vs 864 on M1 baseline = +170 net new tests)
- [x] `npm test` in `kamiwaza-ai-extensions-lib/` — **62 passed**
- [x] No regressions vs. M1 baseline
- [x] `uv run ruff check` matches develop's pre-existing baseline; M2 net-new lint debt: 0
- [x] Verification report: PASS (iteration 2/3) — all dimension scores 8–9, +1.1 mean increase from iteration 1
- [ ] Manual: confirm `kz-ext update --dry-run` on an out-of-tree scaffold (post-merge smoke)

## Verification Report Summary

Two-iteration LocalReview cycle. Iteration 1 surfaced 2 Critical (TS post-commit translation gap + missing test) and 9 Important findings; all Critical + 7 Important fixed in `fcb2efb`. Iteration 2 verdict: **PASS**.

| Dimension | Iter 1 | Iter 2 |
|-----------|--------|--------|
| Design Adherence | 9 | 9 |
| Test Quality | 8 | 9 |
| Code Quality | 7.5 | 8.5 |
| Error Handling | 7 | 9 |
| Security | 7 | 8 |
| Refactoring | 7 | 8.5 |

Two Important findings explicitly deferred (recorded in artifact):
- **I10:** `kamiwaza_extensions_lib/middleware/` directory naming (cosmetic — actually a streaming-proxy helper, not FastAPI middleware). Follow-up rename PR.
- **I11:** Re-validate `name` from `kamiwaza.json` against `_NAME_RE` before render (defense-in-depth — same author owns input + output, not a privilege boundary).

## Defers / follow-ups (3 Linear issues remain in Backlog)

- **ENG-3894** (High) — GoReferenceExtension + `extensions/go-reference-e2e` CI. Needs Go toolchain + new CI workflow; deferred as a follow-up PR. The non-SDK flow doc + canonical test vectors that the Go reference will consume DID land in this PR.
- **ENG-3329** (Low) — Makefile deprecation. Cross-repo PR to `kamiwaza-extensions-template`. Inventory script + 14-repo replay across the M4-named extensions (kaizen, skills-library, omniparse, hello-web, workroom-manager + 9 more).
- **ENG-3896** (Urgent) — ZeroDowntimePATCHContractTest in `kamiwaza-internal/kamiwaza/tests/`. Cross-repo platform test; SDK-team-owned. Presence is the M2 milestone gate; deferred to follow-up cross-repo PR.


jxstanford (changes requested):
# Review Report: feat(extensions): D210 M2 Phase B — kz-ext update + runtime-lib parity + non-SDK contract (7 issues)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/86
**Branch**: `feature/d210-m2-extension-developer-experience` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| C1 | `kamiwaza_extensions/commands/update.py:297-315` | `_stamp_version` clobbers JSON-merge output. After `_reconcile_json_merge` writes the merged kamiwaza.json (with new template fields preserved) to disk, `_stamp_version` immediately re-writes it from the in-memory `metadata` dict, which was loaded BEFORE `_reconcile_file` ran and never updated. Any template-added fields the merge wrote are silently lost when `recorded_version != manifest.template_version`. v1 manifests don't add new fields so it doesn't bite today, but the very first M3 template change that adds a `kamiwaza.json` field will fail silently. Fix: re-read the file inside `_stamp_version` (or update `metadata` from `_reconcile_json_merge`). | Claude Code | Logic |
| C2 | `kamiwaza_extensions_lib/middleware/token_refresh.py:62,137` vs. `kamiwaza-ai-extensions-lib/src/server/token-refresh.ts:14-21,829-830` | Cross-language drift the parity contract is supposed to prevent. Python uses a module-level `asyncio.Lock` to serialize refresh calls across all users; TS deliberately has no module-level state and the TS docstring asserts "The Python sibling … takes the same approach: serializing through an `asyncio.Lock` would also mix users". The PR description likewise says "Per-caller refresh (no module-level Promise cache)". One of the two implementations (or both docstrings) is wrong. Beyond the documentation contradiction, the Python lock is a process-wide bottleneck — every concurrent 401 across every tenant serializes through it. Decide on one design and reconcile both implementations + comments. | Claude Code | Concurrency / Cross-lang contract |
| C3 | `kamiwaza_extensions/commands/update.py:449-450` | `--non-interactive` mode silently returns `skipped` for `preserve_if_modified` conflicts and exits 0. The CLI contract and docs say CI mode should fail with a non-zero exit instead of silently producing a partial update. Any repo with an author-modified template-owned file will currently pass automation even though reconciliation did not complete. | Codex | Logic / Contract |
| C4 | `kamiwaza_extensions/commands/update.py:374-378` | `preserve_if_modified` always compares to the *current* render, treating every mismatch as a conflict. The strategy is supposed to auto-update files unchanged from the scaffold's *recorded* template version. As written, when a new CLI release changes any template-owned file (e.g. `README.md`, `docker-compose.yml`, app auth routes), even untouched scaffolds will be prompted/skipped because `recorded_version` is never used to reconstruct the prior render. This neuters the core value of the strategy. | Codex | Logic |

## High Priority Issues
| H1 | `kamiwaza-ai-extensions-lib/src/server/token-refresh.ts:73` | When `streamWithRefresh` is called from Next.js/Node routes with `opts.body = request.body` (the common streaming case), `fetch` will reject because a streamed request body requires `duplex: "half"`. `createProxyHandlers` already sets this flag for the same environment, so POST/PUT/PATCH callers that forward the incoming body cannot use the new refresh helper at all. | Codex | Logic / Compatibility |
| H2 | `kamiwaza-ai-extensions-lib/src/server/token-refresh.ts:91-95` | Retry path reuses `body` for the second `fetch` after a 401, which fails when callers pass a one-shot body (`request.body` / `ReadableStream`). The stream is already consumed by the first request, so the transparent refresh-and-retry path throws instead of retrying successfully for streamed POST requests. Need to clone/buffer the body before the first attempt. | Codex | Logic |
| H3 | `kamiwaza_extensions_lib/middleware/token_refresh.py:20-21` | Module-level docstring claims the lock makes concurrent 401s "fan in to one refresh attempt." That's factually wrong — a Lock serializes; it does not deduplicate. Each waiting caller still invokes `refresh()` once. To actually fan in you'd need a shared Future/Task. Either change the implementation to single-flight (cache the in-progress Future keyed by user) or update the docstring to "serializes refreshes". | Claude Code | Documentation / Concurrency |
| H4 | `kamiwaza-ai-extensions-lib/src/server/errors.ts:5-9` | The path `kamiwaza_extensions/exception_names.json` does not exist; the actual file is `kamiwaza_extensions_lib/exception_names.json`. The comment instructs maintainers to keep classNames synchronized with this file. | Claude Code | Documentation |
| H5 | `kamiwaza_extensions_lib/exception_names.json` vs. `kamiwaza_extensions_lib/errors.py:54`, `kamiwaza-ai-extensions-lib/src/server/errors.ts:46-47` | New `stream_interrupted` class added to both runtime libs but NOT added to `exception_names.json`. The errors.ts comment explicitly says "If you rename one, rename it everywhere"; this PR added a class without "everywhere." Doctor's `_uac_9d_hints()` will surface only 4 of 5 classes. | Claude Code | Cross-lang contract |
| H6 | `docs/extensions/non-sdk-flow.md:213-214`, `tests/unit/extensions_lib/test_identity_extractor_parity.py:9-13` | Doc + test asserts `examples/extensions/go-reference/` "is a working implementation of this contract". That directory does not exist (the PR description acknowledges ENG-3894 is deferred). Either qualify the language ("planned implementation, see ENG-3894") or remove the link until the reference lands. | Claude Code | Documentation |
| H7 | `kamiwaza_extensions/doctor.py:638` | `if not line.lower().startswith("kamiwaza-extensions-lib"): continue` will match any package whose name starts with that prefix (e.g. `kamiwaza-extensions-lib-extras`), and doesn't strip PEP 508 extras `kamiwaza-extensions-lib[extras]>=0.3` cleanly. Stricter check: require the next char to be in `<>=!~ \t;[` or end-of-string, or use `packaging.requirements.Requirement`. | Claude Code | Logic |

| M1 | `kamiwaza-ai-extensions-lib/src/server/identity.ts:647-651` | v0.3 `extractIdentity` swaps `headers.get(...)` for `readStripped(...)` for email/name/workroomId/etc. v0.2 callers that previously received whitespace-only string values now get `null`. CHANGELOG says "No breaking changes" — this is a subtle behavior change worth calling out. | Claude Code | Backward compat |
| M2 | `kamiwaza_extensions/commands/config.py:32-43` | Comment says "Treat the bare-word names list/show/delete as their corresponding subcommand-style invocation" but the code only handles `name == "list"`. `kz-ext config publish-profile show` / `delete` still hit the "Profile name is required" path. Either implement show/delete the same way, or trim the comment to match. | Claude Code | Documentation |
| M3 | `kamiwaza_extensions/commands/update.py:1881` | `merged = {**rendered, **existing}` — author values "win", but if the author *deleted* a template-required field from kamiwaza.json, the merge silently re-adds it from rendered. May be intended, but isn't documented in the strategy table. Worth a one-line note in `update.md`. | Claude Code | Documentation / Logic |
| M4 | `kamiwaza_extensions/compatibility.json:2` | `cli_version` is hardcoded `"0.1.0"`. The test catches drift, but the bundle stays a static file. Consider auto-generating it at build time (or at least adding a release-prep note). | Claude Code | Maintainability |
| M5 | `kamiwaza_extensions_lib/middleware/token_refresh.py:46-49` | Comment claims "content-length and content-encoding stripped because the ASGI server recomputes them from the streamed body." ASGI does not recompute Content-Encoding (it never compresses). Behavior is correct (don't claim compression on a decoded stream) but rationale is wrong. | Claude Code | Documentation |
| M6 | `kamiwaza_extensions/doctor.py:660-670` | The Python compatibility check only validates `==`-pinned versions against `compat_range`. A range pin like `>=99.0,<100.0` (clearly outside `>=0.2,<0.4`) is silently passed. `_npm_lower_bound` already exists; consider mirroring it for Python (`SpecifierSet.filter([…])` / `Version` lower-bound regex). | Claude Code | Logic |
| M7 | `kamiwaza_extensions/commands/update.py:1909-1916` | `_prompt_conflict` prints diff via `console.print(diff)`. Rich interprets `[…]` markup inside printed strings — a diff of a file containing literal `[red]` would be rendered styled rather than literal. Low risk for templates, but consider `console.print(diff, markup=False)` for safety. | Claude Code | Robustness |
| M8 | `tests/unit/extensions/test_update_command.py:3829-3847` | `test_update_rewrites_template_version_after_success` only asserts `template_version`. It doesn't catch C1 (loss of merge-added fields). Add an assertion that fields the merge added survive `_stamp_version`. | Claude Code | Test gap |

## Test Coverage Assessment
- **Adequate coverage**: `template_manifest.py` (invariant test forces classification), `commands/update.py` (29 tests cover all 5 modes plus migrations), `kamiwaza_extensions_lib/middleware/token_refresh.py` (4 contract states tested), `kamiwaza_extensions_lib/identity.py` (parity vectors), `kamiwaza-ai-extensions-lib/src/server/{errors,identity,token-refresh}.ts`, `kamiwaza_extensions/doctor.py` runtime-lib probes.
- **Gaps**:
  - `_stamp_version` interaction with `_reconcile_json_merge` (C1, M8) — add a test where the rendered template adds a new `kamiwaza.json` field and assert it persists.
  - `--non-interactive` exit-code contract on `preserve_if_modified` conflicts (C3).
  - `preserve_if_modified` against historical template snapshots (C4) — current tests render only against the current template, masking the missing recorded-version replay.
  - `streamWithRefresh` with streamed `ReadableStream` request bodies (H1, H2) — current TS tests pass static strings/buffers.
  - `_check_python_runtime_lib` against malformed lines (`kamiwaza-extensions-lib[extras]>=0.3`, env markers, sibling-prefix package).
  - Concurrent `stream_with_refresh` calls under pytest-asyncio with function-scoped loops (validates module-level `asyncio.Lock`).
  - `kz-ext update` from a subdirectory of an extension.

## Positive Observations
- The canonical `test-vectors.json` consumed bit-identically by Py + TS suites is exactly the kind of cross-language contract that prevents drift; parity tests are well-structured.
- `template_manifest.py` invariant test (`test_every_template_file_is_classified`) guarantees future template additions can't slip past `kz-ext update` silently.
- `_reconcile_file` was refactored from a 90-line behemoth into composable per-strategy helpers — readability is good.
- `Identity` field additions (`systemHigh`, `workroomRole`, `requestId`) correctly use camelCase TS / snake_case Py, with explicit projection tests.
- `_drain` correctly uses `try/finally` so the upstream session is closed even on consumer cancellation; `MID_STREAM_FAIL` is properly mapped to `StreamInterruptedError`.
- `--bootstrap` semantics (only valid when `template_version` is missing; otherwise warn-and-skip) are sensible and documented.

## Overall Verdict
**Risk Level**: High
**Verdict**: REQUEST_CHANGES

Four critical issues span both engines: C1 (silent data loss when JSON merge meets stamp-version), C2 (cross-language token-refresh contract drift), C3 (`--non-interactive` violates documented exit-code contract), and C4 (`preserve_if_modified` is broken because it never reconstructs the recorded-version render). H1/H2 also block real-world usage of `streamWithRefresh` for streamed request bodies — the most common Next.js proxy case. These are not stylistic items; C4 in particular renders the headline "auto-update unmodified template files" feature non-functional after any template change. The remaining items (stale doc paths, missing `stream_interrupted` in `exception_names.json`, doc references) are quick fixes.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |
| Gemini | ⬜ Not available |

**Engines contributing**: 2/3
</details>

jonathanleff:
## Review round 2 — pushed 2 commits addressing 4 Critical, 7 High, 6 of 8 Medium

Thanks for the thorough review. Two fix commits pushed:
- `0fa4991` — C1, C2, C3 + all 7 Highs + 6 Mediums
- `34f9ffb` — C4 (hash-on-create + clean detection)

### Critical resolution

| # | Resolution |
|---|-----------|
| **C1** | `_stamp_version` re-reads `kamiwaza.json` from disk before stamping, preserving merge-added fields. Regression test (`test_stamp_version_preserves_merge_added_kamiwaza_fields`) covers it (M8 too). |
| **C2** | Removed the Py module-level `asyncio.Lock`. Py and TS now both do per-caller refresh — no inter-caller coordination, no module-level state. Docstrings reconciled. The "single-flight fan-in" claim was always false (Lock serializes; doesn't dedupe), thanks for catching it. |
| **C3** | `--non-interactive` raises `typer.Exit(ExitCode.VALIDATION)` when `summary.conflicts > 0`. Two new tests cover the success-path and failure-path. |
| **C4** | Implemented option (b) — hash-on-create. Scaffolder writes SHA-256 per `preserve_if_modified` file into `kamiwaza.json.template_file_hashes`. Update branches: hash-matches-recorded → auto-overwrite + refresh hash (the headline "clean since record" path); hash-diverges → existing conflict path. `--bootstrap` records on-disk hashes for old scaffolds. 4 new tests. Walked through alice/bob example in PR thread before implementing — happy to discuss further if the design isn't right. |

### High resolution

| # | Fix |
|---|-----|
| **H1+H2** | New `_materializeBody()` buffers BodyInit (`ReadableStream`, `Blob`, `FormData`, etc.) into `Uint8Array` up-front — bypasses Node's `duplex: "half"` requirement and allows the retry to replay identical bytes. Test added covering `ReadableStream` input. |
| **H3** | Py docstring rewrite — drops the "fan in" claim; matches the no-coordination implementation. |
| **H4** | `errors.ts` comment path corrected to `kamiwaza_extensions_lib/exception_names.json`. |
| **H5** | Added `stream_interrupted` to `exception_names.json` (exit_code 14) and `ExitCode.STREAM_INTERRUPTED`. Doctor's `_uac_9d_hints()` now surfaces all 5. |
| **H6** | Both `non-sdk-flow.md` and the parity-test docstring now describe the Go reference as "planned, see ENG-3894" instead of asserting it exists. |
| **H7** | Switched `doctor.py` requirements parsing to `packaging.requirements.Requirement` — exact name match, PEP 508 extras handled. Two new tests (prefix-alias rejection, `[extras]` handling). |

### Medium resolution

| # | Fix |
|---|-----|
| **M1** | CHANGELOG note added to `kamiwaza-ai-extensions-lib/CHANGELOG.md` documenting the whitespace-only header behavior change. |
| **M2** | Comment in `commands/config.py` clarified — only `list` is routed (show/delete need explicit `--show <name>`/`--delete <name>` because their argument shape differs). |
| **M3** | `cli-reference/update.md` strategy table now documents JSON-merge "deletion-resurrection" semantics. |
| **M5** | `token_refresh.py` Content-Encoding comment corrected (we strip because we forward decoded bytes; ASGI doesn't compress). |
| **M6** | New `_python_spec_outside_supported()` does range-vs-range overlap (probes `==`-pins and lower bounds). Catches `>=99.0,<100.0` as a warn. Test added. |
| **M7** | `_prompt_conflict` uses `console.print(diff, markup=False)` so a diff containing literal `[red]`-shaped substrings doesn't render as Rich markup. |
| **M8** | Covered alongside C1 — `test_stamp_version_preserves_merge_added_kamiwaza_fields` asserts merge-added fields survive the post-loop persist. |

### Deferred (out of scope for this revision)

- **M2 partial** — only `list` routed; `show`/`delete` still require explicit flags. Argument-shape ambiguity (bare `show` could be a profile literally named "show") — wanted alignment before generalizing.
- **M4** — `compatibility.json.cli_version` auto-generation at build time. Coherence test catches drift; auto-gen is build-tooling work tracked for M3.

### Tests

- **Py:** 866 → 877 (+11 net new from review fixes)
- **TS:** 62 → 63 (+1 mid-stream/streamed-body parity)
- **Total: 940 tests pass.** Lint at develop baseline.

Ready for re-review whenever you have a moment.

jxstanford (commented):
# RE-REVIEW Report: feat(extensions): D210 M2 Phase B — kz-ext update + runtime-lib parity + non-SDK contract (7 issues)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/86
**Branch**: `feature/d210-m2-extension-developer-experience` → `develop`
**Engines**: Codex (1/3 available)

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| 1 | `kamiwaza_extensions/doctor.py:57-68` | Compatibility bundle (`compatibility.json`) is loaded via `importlib.resources` but `pyproject.toml` only declares `templates/**/*` as package data. In wheel/sdist installs the file is absent, so `kz-ext doctor` raises `FileNotFoundError` whenever it reaches the runtime-lib checks. | Codex | Packaging / Runtime crash |
| 2 | `kamiwaza_extensions/commands/update.py:144-153` | The `--non-interactive` conflict gate runs *after* `_reconcile()` has already written clean-file updates and bumped `kamiwaza.json.template_version`. A failing non-interactive update exits non-zero but leaves a dirty checkout marked as current while conflicting template-owned files were skipped — silent state corruption. | Codex | Correctness / State integrity |

## High Priority Issues
| 3 | `kamiwaza_extensions/scaffolder.py:41-46` | Fresh Python app/tool scaffolds render `kamiwaza-extensions-lib>=0.1.0`, but the new compatibility bundle treats `<0.2` as unsupported. Newly created extensions immediately trigger `kz-ext doctor` warnings before the author changes anything. | Codex | Consistency / DX |

| — | — | None raised by available engines. | — | — |

## Test Coverage Assessment
- Coverage looks reasonable for the happy paths: `test_update_command.py` exercises non-interactive conflict handling, `test_compatibility_bundle.py` covers TS-M2-37..40, and `test_token_refresh.py` covers the mid-stream refresh contract on both Python and TS sides.
- **Gap (Critical #1)**: No test installs the package as a wheel/sdist and runs `doctor`, so the missing `compatibility.json` in `package-data` is not caught. Add an installed-package smoke test or extend `test_compatibility_bundle.py` to use `importlib.resources.files()` against the installed distribution.
- **Gap (Critical #2)**: `test_non_interactive_exits_validation_on_conflict` asserts the exit, but does not assert that `kamiwaza.json.template_version` and clean template-owned files were left untouched. Add an assertion that the scaffold is byte-identical to its pre-call state on non-interactive failure.
- **Gap (High #3)**: No test verifies that a freshly scaffolded project passes `kz-ext doctor` with zero warnings — the most basic DX invariant.

## Positive Observations
- Clean separation of pure-data manifests (`template_manifest.py`) from behavior (`commands/update.py`) — easy to extend per shape.
- Cross-language error-class parity via `exception_names.json` registry is a thoughtful approach to keeping Python/TS runtime libs in sync, with a code comment that explicitly calls out the rename-everywhere requirement.
- Identity model deliberately excludes `X-Auth-Token` to prevent credential leakage through `model_dump()` — correct security posture, well documented.
- Compatibility drift surfaces as **warn**, not fail, per design §4.2.10 — author-friendly default.
- Manifest-driven update with per-file strategies (`overwrite` / `preserve_if_modified`) and `.orig` backups is the right shape for a reconcile command.

## Overall Verdict
**Risk Level**: High
**Engine Agreement**: 1/3 engines contributed (Claude Code hit max-turns; third engine unavailable)

The two critical issues are both shipping-blockers: a packaged `kz-ext doctor` will crash, and a failed non-interactive update leaves the scaffold in an inconsistent state that the user will not notice until a later `update` run silently skips files. Issue #3 compounds #1 by guaranteeing every new scaffold trips a warning the moment it's created. Recommend addressing all three before merge; the fixes are local (declare package data, move conflict gate before `_reconcile()`, bump scaffolder pin to `>=0.2.0`).

Note: Only Codex completed; Claude Code timed out at 40 turns. Consider re-running before merge if you want a second independent pass on the update-command state machine.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ❌ Failed (exit 1) |
| Codex | ✅ OK |
| Gemini | ⬜ Not available |

**Engines contributing**: 1/3
</details>


jonathanleff:
## Round 2 review — pushed `0aa3413` addressing all 3 findings

Thanks Codex. All three were genuine; the verdict-WITHHELD note was useful but the findings stand on their own.

### Resolution

| # | Status | Approach |
|---|--------|----------|
| **C-Pkg** | Fixed | Added `compatibility.json` to `[tool.setuptools.package-data]` for `kamiwaza_extensions`. Verified `uv build` now ships the file in the wheel. New regression test grepps `pyproject.toml` so a future contributor can't silently drop it. |
| **C-Order** | Fixed | `--non-interactive` now runs a `dry_run=True, quiet=True` plan-only pre-pass. If `summary.conflicts > 0`, exit before any files are written. Otherwise proceed with the normal reconcile. Regression test asserts post-failure invariants: `template_version` unchanged AND clean `preserve_if_modified` files were not rewritten. |
| **H-FreshPin** | Fixed | `build_render_context` now reads runtime-lib pins from `compatibility.json` (single source of truth — drift-proof by construction, cached via `lru_cache`). Bumped the committed `examples/chatbot-app/{backend/requirements.txt,frontend/package.json}` to match so the chatbot↔scaffold byte-equality test stays green. New test asserts a fresh-scaffold pin passes its own doctor check. |

### Test impact

- **Py:** 877 → **880** (+3 from this round: package-data invariant, `--non-interactive` no-partial-write, fresh-scaffold pin)
- **TS:** unchanged at 63
- **Total: 943 tests pass.** Lint at develop baseline (36 errors, all pre-existing in untouched files).

### Notes on the existing `>=0.1.0` pins

The docs/example update from `>=0.1.0` to `>=0.2,<0.4` is a soft-incompatibility for old scaffolds in the wild — anyone with `kamiwaza-extensions-lib>=0.1.0` will see a doctor warning until their next `kz-ext update`. The C4 hash-on-create path closes this naturally: existing scaffolds bootstrap → record on-disk hash for `requirements.txt` → next update re-renders requirements.txt with the new pin → on-disk hash matches recorded → silent overwrite to the new pin. So the warning is self-healing on the next update cycle.

Ready for re-review. Confidence on this PR is now: 4 reviews, 11 fixes across 4 critical findings (originally) + 7 high + 6 medium + this round's 3.

jxstanford (commented):
# RE-REVIEW Report: feat(extensions): D210 M2 Phase B — kz-ext update + runtime-lib parity + non-SDK contract (7 issues)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/86
**Branch**: `feature/d210-m2-extension-developer-experience` → `develop`
**Engines**: Codex (1/3 available)

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| 1 | `kamiwaza-ai-extensions-lib/src/server/proxy.ts:4-15` | `FORWARD_REQUEST_HEADERS` allowlist drops `X-User-Workroom-Role` and `X-User-System-High` on the Next.js→backend hop. Generated `/session` and `/api/[...path]` routes lose workroom role and classification fields, so backend `extract_identity()` sees `workroom_role=None` / `system_high=None` — breaks the standard scaffolded proxy path. | Codex | Correctness / Regression |

## High Priority Issues
| 2 | `kamiwaza_extensions/doctor.py:94-103` | `_python_spec_outside_supported` only inspects `==`, `>=`, `>`, `~=` operators. An upper-bound-only pin like `kamiwaza-extensions-lib<0.2` returns `None` (compatible), even though every allowed version is outside the bundled support window. False-negative in the new compatibility check. | Codex | Correctness |
| 3 | `kamiwaza-ai-extensions-lib/src/server/token-refresh.ts:299-316` | `streamWithRefresh` throws synchronously for `FormData` / `URLSearchParams` bodies even though `ProxyOpts.body` is typed `BodyInit`. Either accept these bodies (serialize and propagate the generated `Content-Type` with multipart boundary) or narrow the type so the API doesn't advertise unsupported shapes. Multipart/form-urlencoded requests can't use the refresh helper. | Codex | API Contract |

| — | — | None beyond the above. | — | — |

## Test Coverage Assessment
- Strong unit + integration coverage for the new surfaces: `UpdateCommand` modes (TS-M2-1..11), template manifest invariants, compatibility bundle (TS-M2-37..40), identity parity/strict (Py↔TS), token-refresh three-state contract (PRE_COMMIT_OK / PRE_COMMIT_401 / MID_STREAM_FAIL) including a 5-way concurrency test, non-SDK flow contract (TS-M2-16/17), and replay tests against historical template snapshots.
- **Gap (related to Issue #1)**: no test asserts the proxy forwards the full envelope (specifically `X-User-Workroom-Role` and `X-User-System-High`) end-to-end. Adding that assertion in `kamiwaza-ai-extensions-lib/tests/proxy.test.ts` would have caught the regression and would prevent recurrence after the fix.
- **Gap (Issue #2)**: `doctor` compatibility tests don't cover upper-bound-only specifiers (`<X`, `<=X` with no lower bound). Worth adding a parametrized case.
- **Gap (Issue #3)**: no `streamWithRefresh` test exercises `FormData` / `URLSearchParams` bodies — current behavior (sync throw) is therefore unverified by intent.

## Positive Observations
- Three-state token-refresh contract is mirrored cleanly across Py and TS, with a deliberate, well-commented rejection of cross-request promise caching (avoids the multi-tenant credential-mixup that earlier shape would have caused).
- Per-file strategy model in `template_manifest.py` + the "every templated file classified exactly once" CI invariant is exactly the right shape for a long-lived `kz-ext update` contract — explicit-by-design, fails loudly on drift.
- `template_file_hashes` ("clean since record") flow makes `preserve_if_modified` silently sweep unmodified files instead of prompting — good UX without sacrificing safety; `--bootstrap` cleanly handles pre-M2 scaffolds.
- `Identity` deliberately omits `X-Auth-Token` to prevent credential leakage via `model_dump()` in logs/metrics/error payloads — comment captures the why; `/session` regression test enforces it.
- `exit_code_for(class_name)` driven from `exception_names.json` keeps CLI exit codes, runtime-lib classes, and Doctor hints in lockstep across Py/TS; addition of `STREAM_INTERRUPTED = 14` is consistent.
- CHANGELOGs, doc additions (`update.md`, `non-sdk-flow.md`, `developer-guide.md`), and canonical `test-vectors.json` give external (Go) implementers a concrete contract to build against.

## Overall Verdict
**Risk Level**: High
**Engine Agreement**: 1/3 engines contributed (Claude Code hit max-turns; only Codex produced findings — confidence on completeness is reduced, but Issue #1 is a concrete shipped-template regression and is sufficient on its own.)

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ❌ Failed (exit 1) |
| Codex | ✅ OK |
| Gemini | ⬜ Not available |

**Engines contributing**: 1/3
</details>


jxstanford (commented):
# RE-REVIEW Report: feat(extensions): D210 M2 Phase B — kz-ext update + runtime-lib parity + non-SDK contract (7 issues)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/86
**Branch**: `feature/d210-m2-extension-developer-experience` → `develop`
**Engines**: Codex (1/3 available)

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| 1 | `kamiwaza_extensions/commands/update.py:315-318` | `_reconcile()` only forwards `name` and `type_` into `build_render_context()`, so `{{version}}` and `{{description}}` fall back to the scaffold defaults (`0.1.0` / `A Kamiwaza … extension`). Any update that re-renders templated files (e.g. `README.md`, `frontend/package.json`, `frontend/src/app/layout.tsx`) will use stale values, producing spurious conflicts or silently overwriting the project's own metadata with original scaffold text. | Codex | Correctness / Data integrity |

## High Priority Issues
| 2 | `kamiwaza_extensions/doctor.py:142-143` | `_python_spec_outside_supported()` treats any overlap with the supported window as a pass. Open-ended specs like `kamiwaza-extensions-lib>=0.2` or `~=0.2` are reported as compatible with `>=0.2,<0.4`, even though a fresh install can legally resolve `0.4+` (outside the CLI's advertised window). Defeats the new compatibility-bundle drift detection on the Python side. | Codex | Compatibility / False negative |
| 3 | `kamiwaza_extensions/doctor.py:779-785` | TypeScript runtime-lib check only validates the first version-like token of a declared range, so `"@kamiwaza-ai/extensions-lib": ">=0.2"` passes against `>=0.2,<0.4` while npm can resolve `0.4+`. Same false-negative gap as the Python checker, on the TS side. | Codex | Compatibility / False negative |

| 4 | `kamiwaza_extensions/scaffolder.py:164-167` | New "scaffold into `./<name>/`" branch assumes `cwd/<name>` is a directory and calls `iterdir()` directly. If a regular file or symlink with that name already exists, `kz-ext create --name …` crashes with `NotADirectoryError` instead of returning the intended `FileExistsError` message — a regression in the new non-empty-directory flow. | Codex | UX / Error handling |

## Test Coverage Assessment
- Codex did not flag missing tests for the touched paths, but the issues above suggest specific gaps worth closing:
  - No `update` test exercises a project whose `kamiwaza.json` `version`/`description` differs from scaffold defaults — would have surfaced issue #1.
  - No `doctor` test exercises open-ended specifiers (`>=0.2`, `~=0.2`, bare `>=X`) for either Python or TS runtime-lib pins — would have surfaced #2 and #3.
  - No `scaffolder` test for "target name exists as file/symlink" in the new non-empty-cwd branch — would have surfaced #4.
- Existing additions (`test_compatibility_bundle.py`, `identity-parity.test.ts`, `errors.test.ts`, `token-refresh.test.ts`) cover the new contracts at the happy-path level but do not catch these edge cases.

## Positive Observations
- Three-state mid-stream-safe token-refresh contract (`PRE_COMMIT_OK` / `PRE_COMMIT_401` / `MID_STREAM_FAIL`) is implemented in parity across Python and TS with matching tests.
- Canonical error hierarchy is centralized via `exception_names.json` and mirrored in both runtime libs, with a clear cross-language "rename in both places" comment.
- New `BodyTooLargeError` cap on `streamWithRefresh` buffering prevents resource exhaustion on retried bodies.
- `kz-ext update` adds a per-file strategy (`overwrite` / `preserve_if_modified`) with `.orig` backups, dry-run, and `--non-interactive` for CI — solid scaffold-evolution model.
- Round-3 fix to forward `x-user-system-high`, `x-user-workroom-role`, and the `x-user-workroom-id` alias closes a real identity-envelope hole on proxied requests.

## Overall Verdict
**Risk Level**: Medium
**Engine Agreement**: 1/3 engines contributed (Claude Code hit max-turns; Kamiwaza n/a)

Note: Verdict driven primarily by issue #1 (silent metadata regression on `kz-ext update`), with #2/#3 weakening the headline compatibility-bundle feature this PR ships. Single-engine coverage limits cross-validation — a follow-up review pass once Claude Code is unblocked is advisable before merge.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ❌ Failed (exit 1) |
| Codex | ✅ OK |
| Gemini | ⬜ Not available |

**Engines contributing**: 1/3
</details>


jxstanford (commented):
# RE-REVIEW Report: feat(extensions): D210 M2 Phase B — kz-ext update + runtime-lib parity + non-SDK contract (7 issues)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/86
**Branch**: `feature/d210-m2-extension-developer-experience` → `develop`
**Engines**: Codex (1/3 available)

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| 1 | `kamiwaza_extensions/compatibility.json:8` | TypeScript runtime pin uses Python-style range `">=0.2,<0.4"`, which npm/pnpm cannot parse as valid semver. This value is rendered into scaffolded `frontend/package.json` files (and is already in `examples/chatbot-app/frontend/package.json`), so fresh scaffolds break at `npm install`. Use npm syntax like `">=0.2 <0.4"` or `"^0.3.0"`. | Codex | Correctness / Build break |

## High Priority Issues
| 2 | `kamiwaza_extensions/doctor.py:97-99` | `~=` is treated as a lower-bound-only operator in `_spec_bounds`, so PEP 440 compatible-release pins like `kamiwaza-extensions-lib~=0.3.0` (which expands to `>=0.3.0,<0.4.0` and is fully inside the supported `>=0.2,<0.4` window) are falsely warned by `kz-ext doctor`. Derive the implied upper bound for `~=` (e.g., `~=X.Y.Z` → `<X.(Y+1)`). | Codex | Correctness |
| 3 | `kamiwaza_extensions/commands/update.py:501-502` | Binary template-owned files bypass the per-strategy update path, so overwrite-mode binary assets (e.g., `frontend/public/kmza-icon.png` in the app manifest) are replaced in place with no `.orig` backup. Author customizations to template-owned binaries are silently lost on `kz-ext update --force`, violating the documented overwrite-with-backup contract. | Codex | Data loss |

## Test Coverage Assessment
- Test suite could not be executed by Codex (collection error in environment); coverage cannot be empirically verified here.
- Test files for the new surfaces are present and substantive: `tests/unit/extensions/test_update_command.py`, `test_template_manifest.py`, `test_compatibility_bundle.py`, `test_doctor.py`; `tests/unit/extensions_lib/test_token_refresh.py`, `test_identity_extractor_parity.py`; TS mirrors at `kamiwaza-ai-extensions-lib/tests/token-refresh.test.ts` and `identity-parity.test.ts`.
- Gaps implied by findings: no test exercises `~=` requirements through `_check_python_runtime_lib` (Issue #2); no test asserts `.orig` backup creation for modified **binary** template-owned files under `overwrite` (Issue #3); no test renders `compatibility.json`'s TS pin into a scaffolded `package.json` and runs `npm install`/parse against it (Issue #1).

## Positive Observations
- Three-state token-refresh contract (`PRE_COMMIT_OK` / `PRE_COMMIT_401` / `MID_STREAM_FAIL`) implemented in parity across Python and TypeScript with shared semantics and matching error types.
- Range-vs-range compatibility checking (M6) is a real correctness improvement over the prior `==`-pin-only path; round-3 H1 upper-bound-only edge case is explicitly tested.
- Identity extraction now has canonical JSON test vectors driving both Python and TS implementations, closing prior parity drift.
- `kz-ext update` design is conservative: per-file strategies, recorded SHA hashes for `preserve_if_modified`, `--dry-run`/`--non-interactive` for CI, and `.orig` backups (for text paths).
- Proxy header allowlist gap (`x-user-system-high`, `x-user-workroom-role`, `x-user-workroom-id`) is fixed and rationale is documented inline.

## Overall Verdict
**Risk Level**: High
**Engine Agreement**: 1/3 engines contributed (Claude Code hit max-turns; Kamiwaza unavailable). Single-engine signal — Issue #1 is a hard build break for new scaffolds and is independently verifiable from the diff; Issues #2–#3 are concrete correctness gaps that warrant a fix or explicit acknowledgment before merge.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ❌ Failed (exit 1) |
| Codex | ✅ OK |
| Gemini | ⬜ Not available |

**Engines contributing**: 1/3
</details>


kbot-ships (dismissed):
_Reviewing head `c9d8c46` against `develop`._

## Bottom line

This PR delivers seven issues worth of SDK-side extension developer tooling: a canonical non-SDK contract doc with shared test vectors, cross-language runtime-lib parity (Py + TS 0.3.0), a compatibility-bundle doctor probe, CLI polish, `kz-ext update` with template manifests and hash-based clean detection, and a token-refresh middleware with a clear three-state contract. The direction is solid — the design is thoughtful, the multi-round review history shows genuine hardening, and the test coverage is substantive (~120 tests across the 7 issues). I have one worth-fixing finding and several nits; nothing blocking.

## Findings

### 1. `_open` uses raw `__aenter__`/`__aexit__` on httpx stream context manager **(worth fixing)**

`kamiwaza_extensions_lib/middleware/token_refresh.py:91,104` — `_UpstreamSession` holds a reference to the context manager object and calls `__aenter__`/`__aexit__` directly. This works but is fragile: if `__aenter__` raises (network error during the TLS handshake, DNS failure, connection pool timeout), the `_UpstreamSession` is never constructed, so `aclose()` is never called — and the half-open context manager leaks. The `stream_with_refresh` caller does not wrap the `_open` call in a try/except that would close the context manager on failure.

Suggestion: wrap the `ctx.__aenter__()` in a try/except inside `_open` and call `ctx.__aexit__` on failure, or restructure so `_UpstreamSession` is itself an `async with` target. Low probability in practice (httpx handles most of this internally), but worth hardening for a middleware that runs on every proxied request.

### 2. `--force` + `--non-interactive` flags are not mutually exclusive **(nit)**

`kamiwaza_extensions/commands/update.py:116-195` — both flags are accepted simultaneously without validation. If a user passes `--force --non-interactive`, the `non_interactive` plan-only pass runs first (which checks for conflicts), then `_reconcile` runs with `force=True` (which overwrites conflicts). In practice the plan pass would report 0 conflicts (because the apply pass would force them), but the plan pass runs with `dry_run=True, force=False` — so it *would* see conflicts, then error out before the force pass ever runs. This makes `--force --non-interactive` equivalent to `--non-interactive` (force is silently ignored). Consider either rejecting the combination or documenting the precedence.

### 3. `stream_interrupted` failure class missing from non-SDK flow doc **(nit)**

`docs/extensions/non-sdk-flow.md:149-161` — §5 "Failure semantics" lists 5 classes but omits `stream_interrupted` (exit code 14), which was added in this PR. Non-SDK authors implementing the token-refresh pattern won't know to surface this class for mid-stream failures. The doc and `exception_names.json` are out of sync.

### 4. `_reconcile_json_merge` additive-only semantics may surprise on `--bootstrap` → `update` round-trip **(nit)**

`kamiwaza_extensions/commands/update.py:652-695` — `merged = {**rendered, **existing}` means if the template *adds* a new field in a future CLI version, it's picked up (good). But if the template *removes* a field, `update` silently preserves the stale value from the user's file. This is documented in `update.md` and is by-design, but the `_reconcile_json_merge` function itself has no inline note about it — the doc is the only place the semantics are explained. A single-line comment on the merge line would help future readers.

### 5. `compatibility.json` `cli_version` is hard-coded to `"0.1.0"` **(nit)**

`kamiwaza_extensions/compatibility.json:2` — the `cli_version` field is `"0.1.0"`, and a test (`test_cli_version_matches_package_version`) asserts it equals `kamiwaza_extensions.__version__`. This works today only because `__version__` is also `"0.1.0"`. On the first CLI version bump, someone must remember to update this file. A build-time generation step (or at least a prominent comment in the JSON) would reduce the foot-gun surface. The existing test is the safety net, but the fix hint in the test suggests auto-generation as a follow-up — fair enough, but this is the kind of thing that bites on a release day.

## Test gaps

- **`--force --non-interactive` combination** — no test covers what happens when both are passed. The current behavior (non-interactive wins, force is silently ignored) should have a test pinning that contract so a future refactor doesn't accidentally let force bypass the plan-pass guard.

- **`_open` exception during `__aenter__`** — no test simulates a connection-level failure (e.g., DNS timeout) during the `_open` call in `stream_with_refresh`. The mid-stream and pre-commit paths are well-tested, but the "can't even open" path relies on httpx's internal cleanup.

- **`_reconcile_json_merge` with a template that removes a field** — the additive-only behavior is documented but not tested. A test showing that a field present in the user's `kamiwaza.json` but absent from the rendered template survives the merge would pin the design decision.

## What's good

The hash-based clean detection for `preserve_if_modified` files (PR-86 C4 option b) is well-designed — it solves the "every CLI bump prompts on every file" problem cleanly, and the bootstrap path for old scaffolds is a thoughtful migration story. The canonical test-vectors approach for cross-language parity (Py, TS, and future Go all consuming the same JSON) is excellent — it's the right way to enforce contract parity without duplicating assertions.

The `--non-interactive` plan-then-apply design (with the documented TOCTOU caveat) is the right call for CI use. The per-caller refresh decision in the token middleware (no module-level Promise cache) correctly avoids the cross-user credential mixup in multi-tenant Next.js — that's a subtle correctness call that would be easy to get wrong.

Test coverage is thorough: 120+ tests across the 7 issues, with invariant tests (every template file is classified), coherence tests (compatibility.json matches __version__), and parity tests (same vectors, Py and TS). The conftest LRU-cache clearing fixture is a nice detail that prevents cross-test pollution.

## Verdict

**Verdict**: APPROVE

Solid delivery across 7 issues with thorough tests, careful cross-language parity, and well-documented design decisions. The findings are all nits or worth-fixing items — nothing blocking.
---

PR #94, opened 2026-05-07
feat(extensions): --digest flag on kz-ext publish for digest-pinned catalog refs (ENG-4370)

JohnStrick:
## Summary

Adds `--digest sha256:<64-hex>` to `kz-ext publish` and auto-resolves the digest from the registry when omitted, so catalog entries reference images as `image:tag@sha256:digest`. Buildable images only — pass-through external refs (postgres etc.) keep whatever pinning their source repo applied.

## Motivation

ENG-4369 versioning policy makes catalog identity digest-pinned so it survives tag mutability. CI auto-publish (ENG-3593) needs to record the exact image it just built. Pulling `image:tag@sha256:<d>` always returns the same content even if the tag is later re-pointed.

## Contract

- `--digest <sha256:...>` requires exactly one buildable service in compose; multi-buildable runs must rely on auto-resolve.
- When `--digest` is omitted, every buildable ref is resolved post-push via `docker buildx imagetools inspect`. With `--no-push`, the same call runs against the existing tag (the registry already has the image from a prior run).
- `--revision` and `--digest` are orthogonal: revision drives the tag and the `(name, semver, revision)` dedup guard; digest pins identity. Catalog can carry both — `<reg>/<ext>-<svc>:<rev>@sha256:<d>`.
- `RegistryBuilder.transform_image_tags()` now preserves an existing `@sha256:...` suffix when rewriting the tag (was silently dropping it).
- Dry-run skips digest resolution.

## Verification

- 1430 unit tests passing locally; 31 new across image_pusher, registry_builder, publish command.
- `uv run ruff check` clean on all changed files.

## Notes

- Consumer-side parsing of combined `repo:tag@digest` refs is ENG-4592 (blocking), tracked separately. Producer-side ships forward-compatible: older readers that ignore the digest portion will still see a valid `repo:tag`.
- `--revision` kept as-is. Its dedup guard is unique behavior `--digest` does not replace.
- Docs reference `--revision` in `docs/extensions/non-sdk-flow.md` and `docs/extensions/developer-guide.md`. Holding doc updates until ENG-3593 (CI auto-publish) lands so the recommended flow can be documented end-to-end.

### DIFF
--- a/kamiwaza_extensions/cli.py
+++ b/kamiwaza_extensions/cli.py
@@ -345,6 +345,16 @@ def publish(
             "idempotent CI re-publishes; --force overrides."
         ),
     ),
+    digest: Optional[str] = typer.Option(
+        None,
+        "--digest",
+        help=(
+            "Pin the catalog entry to a specific manifest digest "
+            "(sha256:<64-hex>). Catalog refs become 'image:tag@sha256:...' "
+            "for immutable identity. Requires exactly one buildable "
+            "service in compose; omit to auto-resolve per-service digests."
+        ),
+    ),
 ) -> None:
     """Publish extension to catalog."""
     from kamiwaza_extensions.commands.publish import run_publish
@@ -356,6 +366,7 @@ def publish(
         no_push=no_push,
         verbose=_state.verbose,
         revision=revision,
+        digest=digest,
     )
 
 
--- a/kamiwaza_extensions/commands/publish.py
+++ b/kamiwaza_extensions/commands/publish.py
@@ -81,6 +81,7 @@ def run_publish(
     no_push: bool = False,
     verbose: bool = False,
     revision: Optional[str] = None,
+    digest: Optional[str] = None,
 ) -> None:
     """Build, push, and publish extension to catalog."""
     from kamiwaza_extensions.catalog_publisher import (
@@ -93,7 +94,11 @@ def run_publish(
     from kamiwaza_extensions.exit_codes import ExitCode
     from kamiwaza_extensions.extension_detector import ExtensionDetector
     from kamiwaza_extensions.image_builder import ImageBuilder, ImageBuildError
-    from kamiwaza_extensions.image_pusher import ImagePusher, ImagePushError
+    from kamiwaza_extensions.image_pusher import (
+        ImagePushError,
+        ImagePusher,
+        validate_digest,
+    )
     from kamiwaza_extensions.profile_manager import ProfileManager
     from kamiwaza_extensions.registry_builder import RegistryBuilder
     from kamiwaza_extensions.validators.compose import ComposeValidator
@@ -111,6 +116,16 @@ def run_publish(
             console.print(f"[red]Error:[/red] {exc}")
             raise typer.Exit(code=int(ExitCode.VALIDATION)) from exc
 
+    # Same fail-fast intent for --digest: reject malformed input before any
+    # build/push side effects. Format guard only — buildable-count check
+    # happens after compose data is loaded.
+    if digest is not None:
+        try:
+            validate_digest(digest)
+        except ValueError as exc:
+            console.print(f"[red]Error:[/red] {exc}")
+            raise typer.Exit(code=int(ExitCode.VALIDATION)) from exc
+
     # 1. Detect extension
     detector = ExtensionDetector()
     info = detector.detect()
@@ -118,6 +133,23 @@ def run_publish(
     if info.compose_data is None:
         console.print("[red]Error:[/red] No docker-compose.yml found.")
         raise typer.Exit(code=1)
+
+    # ENG-4370: --digest pins identity for one image. Multi-image extensions
+    # must rely on auto-resolve (the per-service push digest), so reject the
+    # ambiguous case before doing any work.
+    if digest is not None:
+        buildable_services = [
+            name for name, svc in (info.compose_data.get("services") or {}).items()
+            if "build" in svc
+        ]
+        if len(buildable_services) != 1:
+            found = ", ".join(buildable_services) if buildable_services else "none"
+            console.print(
+                f"[red]Error:[/red] --digest requires exactly one buildable "
+                f"service in docker-compose.yml; found {len(buildable_services)} "
+                f"({found}). Omit --digest to auto-resolve per-service digests."
+            )
+            raise typer.Exit(code=int(ExitCode.VALIDATION))
 
     version = info.version
     ext_type = _infer_extension_type(info.metadata)
@@ -301,6 +333,23 @@ def run_publish(
     elif no_push:
         console.print("  [dim]Skipping push (--no-push)[/dim]")
 
+    # 6.5 Resolve digests for buildable images (ENG-4370). Catalog refs
+    # are pinned `image:tag@sha256:...` so they're immutable regardless
+    # of tag mutability. Pass-through external/prebuilt-internal refs
+    # (postgres, etc.) keep whatever pinning their source repo applied.
+    digest_map: Dict[str, str] = {}
+    if image_refs:
+        if digest is not None:
+            # Buildable-count validation above guarantees a single ref.
+            digest_map[image_refs[0]] = digest
+        else:
+            try:
+                for ref in image_refs:
+                    digest_map[ref] = ImagePusher.resolve_digest(ref)
+            except ImagePushError as exc:
+                console.print(f"\n[red]Error:[/red] {exc}")
+                raise typer.Exit(code=1) from exc
+
     # 7. Build catalog entry
     reg_builder = RegistryBuilder()
     entry = reg_builder.build_entry(
@@ -310,6 +359,7 @@ def run_publish(
         version=version,
         stage=stage,
         revision=revision,
+        digest_map=digest_map or None,
     )
 
     # 8. Publish to catalog
--- a/kamiwaza_extensions/commands/publish.py
+++ b/kamiwaza_extensions/commands/publish.py
@@ -110,10 +110,6 @@ def _auto_resolve_digests(
     --no-push``): pre-PR behavior was tag-only with no docker dependency,
     so a resolve failure becomes a warning and the ref is omitted from
     the result rather than aborting the publish.
-
-    For the partial shape ``--no-push`` without ``--no-build``, a
-    ``manifest unknown`` registry error is rewrapped with an actionable
-    hint (push first, or pass ``--digest``).
     """
     from kamiwaza_extensions.image_pusher import ImagePushError, ImagePusher
 
@@ -129,13 +125,7 @@ def _auto_resolve_digests(
                     "this ref"
                 )
                 continue
-            if no_push and "manifest unknown" in str(exc).lower():
-                console.print(
-                    f"\n[red]Error:[/red] registry has no image at "
-                    f"{ref}. Push it first or pass --digest explicitly."
-                )
-            else:
-                console.print(f"\n[red]Error:[/red] {exc}")
+            console.print(f"\n[red]Error:[/red] {exc}")
             raise typer.Exit(code=1) from exc
     return digest_map
 
@@ -417,6 +407,19 @@ def run_publish(
         image_refs = _collect_image_refs(
             info.compose_data, info.name, image_tag, registry
         )
+
+    # 5.5 Preflight: digest resolution post-push needs `docker buildx
+    # imagetools`. If buildx is missing on this host, fail before mutating
+    # the registry rather than push-then-fail-on-resolve. Only required
+    # when push will happen and there's at least one published service to
+    # pin (the catalog-only-republish path soft-falls in
+    # `_auto_resolve_digests` and doesn't need this guard).
+    if not no_push and buildable_services:
+        try:
+            ImagePusher.check_buildx_available()
+        except ImagePushError as exc:
+            console.print(f"\n[red]Error:[/red] {exc}")
+            raise typer.Exit(code=1) from exc
 
     # 6. Push images (same stage-aware tag)
     if not no_push and image_refs:
--- a/kamiwaza_extensions/commands/publish.py
+++ b/kamiwaza_extensions/commands/publish.py
@@ -72,6 +72,74 @@ def _collect_image_refs(
     return refs
 
 
+def _verify_supplied_digest(ref: str, supplied: str) -> None:
+    """Resolve *ref* in the registry and abort if it disagrees with *supplied*.
+
+    Catches CI typos, stale digests, and the TOCTOU window where a
+    parallel run re-pointed the tag between our push and our publish.
+    Caller must guarantee the image was just pushed (i.e. ``no_push`` is
+    False) — otherwise the registry isn't authoritative for what we
+    intended to publish.
+    """
+    from kamiwaza_extensions.exit_codes import ExitCode
+    from kamiwaza_extensions.image_pusher import ImagePushError, ImagePusher
+
+    try:
+        actual = ImagePusher.resolve_digest(ref)
+    except ImagePushError as exc:
+        console.print(f"\n[red]Error:[/red] {exc}")
+        raise typer.Exit(code=1) from exc
+    if actual != supplied:
+        console.print(
+            f"\n[red]Error:[/red] supplied --digest does not match "
+            f"the registry manifest for {ref}.\n"
+            f"  supplied: {supplied}\n"
+            f"  registry: {actual}\n"
+            "Re-run with the correct digest, or omit --digest to "
+            "auto-resolve."
+        )
+        raise typer.Exit(code=int(ExitCode.VALIDATION))
+
+
+def _auto_resolve_digests(
+    refs: List[str], *, no_build: bool, no_push: bool,
+) -> Dict[str, str]:
+    """Resolve registry digests for each *ref* and return ``{ref: digest}``.
+
+    Soft-fails for the catalog-only-republish shape (``--no-build
+    --no-push``): pre-PR behavior was tag-only with no docker dependency,
+    so a resolve failure becomes a warning and the ref is omitted from
+    the result rather than aborting the publish.
+
+    For the partial shape ``--no-push`` without ``--no-build``, a
+    ``manifest unknown`` registry error is rewrapped with an actionable
+    hint (push first, or pass ``--digest``).
+    """
+    from kamiwaza_extensions.image_pusher import ImagePushError, ImagePusher
+
+    digest_map: Dict[str, str] = {}
+    for ref in refs:
+        try:
+            digest_map[ref] = ImagePusher.resolve_digest(ref)
+        except ImagePushError as exc:
+            if no_build and no_push:
+                console.print(
+                    f"  [yellow]warn[/yellow] could not resolve digest "
+                    f"for {ref}: {exc} — catalog will use tag-only for "
+                    "this ref"
+                )
+                continue
+            if no_push and "manifest unknown" in str(exc).lower():
+                console.print(
+                    f"\n[red]Error:[/red] registry has no image at "
+                    f"{ref}. Push it first or pass --digest explicitly."
+                )
+            else:
+                console.print(f"\n[red]Error:[/red] {exc}")
+            raise typer.Exit(code=1) from exc
+    return digest_map
+
+
 def run_publish(
     *,
     stage: str,
@@ -123,7 +191,11 @@ def run_publish(
         try:
             validate_digest(digest)
         except ValueError as exc:
-            console.print(f"[red]Error:[/red] {exc}")
+            # The exception text contains the user-supplied digest verbatim;
+            # rich console treats `[…]` as markup, so disable interpretation
+            # to avoid silent stripping or a markup-injection vector.
+            console.print("[red]Error:[/red] ", end="")
+            console.print(str(exc), markup=False)
             raise typer.Exit(code=int(ExitCode.VALIDATION)) from exc
 
     # 1. Detect extension
@@ -156,9 +228,9 @@ def run_publish(
         )
         raise typer.Exit(code=int(ExitCode.VALIDATION))
 
-    # ENG-4370 H1: auto-resolve queries the registry post-build. Built-but-
-    # not-pushed leaves the registry stale (or empty), so the resolved
-    # digest would either error or silently pin the *previous* push.
+    # Auto-resolve queries the registry post-build. Built-but-not-pushed
+    # leaves the registry stale (or empty), so the resolved digest would
+    # either error or silently pin the *previous* push.
     # Force the user to either push, skip the build, or supply --digest.
     # Skipped under: dry-run (no build/push happens), and external-only
     # extensions (no buildable services means no hazard).
@@ -271,13 +343,15 @@ def run_publish(
 
         # Dry-run preview reflects an explicit --digest; auto-resolve
         # is skipped because a dry-run shouldn't talk to the registry.
+        # Mirrors the live path's published_refs derivation: pin against
+        # the buildable_services-derived ref so a profiled helper that
+        # appears first in dict order can't redirect the user's digest.
         dry_digest_map: Dict[str, str] = {}
-        if digest is not None:
-            dry_refs = _collect_image_refs(
-                info.compose_data, info.name, image_tag, registry
+        if digest is not None and buildable_services:
+            dry_canonical_ref = (
+                f"{registry}/{info.name}-{buildable_services[0]}:{image_tag}"
             )
-            if dry_refs:
-                dry_digest_map[dry_refs[0]] = digest
+            dry_digest_map[dry_canonical_ref] = digest
 
         # Run merge check so dry-run detects version conflicts
         reg_builder = RegistryBuilder()
@@ -370,75 +444,27 @@ def run_publish(
     elif no_push:
         console.print("  [dim]Skipping push (--no-push)[/dim]")
 
-    # 6.5 Resolve digests for published buildable images (ENG-4370). Catalog
-    # refs are pinned `image:tag@sha256:...` so they're immutable regardless
-    # of tag mutability. Pass-through external/prebuilt-internal refs
-    # (postgres, etc.) keep whatever pinning their source repo applied.
-    #
-    # Iterate over `published_refs` derived from `buildable_services` (which
-    # already excludes profile-only services) instead of `image_refs`.
-    # `image_refs` reflects what ImageBuilder built and ImagePusher pushed,
-    # which can include profiled helpers — those won't appear in the
-    # transformed catalog compose, so resolving / pinning their digest is at
-    # best wasted work and at worst (with --digest) misdirects the user's
-    # supplied digest at the wrong ref.
+    # Catalog refs are pinned `image:tag@sha256:...` so they're immutable
+    # regardless of tag mutability. published_refs is derived from
+    # buildable_services (which excludes profile-only services per
+    # ComposeTransformer's filter), so a profiled helper appearing first
+    # in image_refs cannot misdirect the digest map. Pass-through external
+    # refs keep whatever pinning their source repo applied.
     published_refs: List[str] = [
         f"{registry}/{info.name}-{name}:{image_tag}" for name in buildable_services
     ]
     digest_map: Dict[str, str] = {}
     if published_refs:
         if digest is not None:
-            # Buildable-count validation above guarantees a single ref.
+            # Single-buildable invariant guaranteed by validation above.
             ref = published_refs[0]
             digest_map[ref] = digest
-            # H2: when push happened, verify the supplied digest matches
-            # the registry's manifest. Catches CI typos, stale digests,
-            # and the TOCTOU window where a parallel run re-pointed the
-            # tag between our push and our publish.
             if not no_push:
-                try:
-                    actual = ImagePusher.resolve_digest(ref)
-                except ImagePushError as exc:
-                    console.print(f"\n[red]Error:[/red] {exc}")
-                    raise typer.Exit(code=1) from exc
-                if actual != digest:
-                    console.print(
-                        f"\n[red]Error:[/red] supplied --digest does not match "
-                        f"the registry manifest for {ref}.\n"
-                        f"  supplied: {digest}\n"
-                        f"  registry: {actual}\n"
-                        "Re-run with the correct digest, or omit --digest to "
-                        "auto-resolve."
-                    )
-                    raise typer.Exit(code=int(ExitCode.VALIDATION))
+                _verify_supplied_digest(ref, digest)
         else:
-            for ref in published_refs:
-                try:
-                    digest_map[ref] = ImagePusher.resolve_digest(ref)
-                except ImagePushError as exc:
-                    # H3 (Codex re-review): under --no-build --no-push the
-                    # caller is doing a catalog-only republish; pre-PR
-                    # behavior was tag-only with no docker dependency.
-                    # Soft-fail to preserve that: warn, leave this ref out
-                    # of digest_map → catalog gets tag-only for it.
-                    if no_build and no_push:
-                        console.print(
-                            f"  [yellow]warn[/yellow] could not resolve "
-                            f"digest for {ref}: {exc} — catalog will use "
-                            "tag-only for this ref"
-                        )
-                        continue
-                    # M4: actionable hint for partial-republish flows
-                    # where the tag isn't in the registry yet.
-                    if no_push and "manifest unknown" in str(exc).lower():
-                        console.print(
-                            f"\n[red]Error:[/red] registry has no image at "
-                            f"{ref}. Push it first or pass --digest "
-                            "explicitly."
-                        )
-                    else:
-                        console.print(f"\n[red]Error:[/red] {exc}")
-                    raise typer.Exit(code=1) from exc
+            digest_map = _auto_resolve_digests(
+                published_refs, no_build=no_build, no_push=no_push,
+            )
 
     # 7. Build catalog entry
     reg_builder = RegistryBuilder()
--- a/kamiwaza_extensions/commands/publish.py
+++ b/kamiwaza_extensions/commands/publish.py
@@ -155,7 +155,10 @@ def run_publish(
     # not-pushed leaves the registry stale (or empty), so the resolved
     # digest would either error or silently pin the *previous* push.
     # Force the user to either push, skip the build, or supply --digest.
-    if not no_build and no_push and digest is None:
+    # Dry-run is exempt: that branch returns before build/push/auto-resolve,
+    # so no hazard exists — preview the catalog without forcing the user
+    # to add ceremony flags.
+    if not dry_run and not no_build and no_push and digest is None:
         console.print(
             "[red]Error:[/red] --no-push without --no-build cannot pin a "
             "catalog digest — the just-built image is only in your local "
--- a/kamiwaza_extensions/commands/publish.py
+++ b/kamiwaza_extensions/commands/publish.py
@@ -151,6 +151,19 @@ def run_publish(
             )
             raise typer.Exit(code=int(ExitCode.VALIDATION))
 
+    # ENG-4370 H1: auto-resolve queries the registry post-build. Built-but-
+    # not-pushed leaves the registry stale (or empty), so the resolved
+    # digest would either error or silently pin the *previous* push.
+    # Force the user to either push, skip the build, or supply --digest.
+    if not no_build and no_push and digest is None:
+        console.print(
+            "[red]Error:[/red] --no-push without --no-build cannot pin a "
+            "catalog digest — the just-built image is only in your local "
+            "daemon. Push first, pass --no-build to publish-only against "
+            "the existing registry tag, or supply --digest explicitly."
+        )
+        raise typer.Exit(code=int(ExitCode.VALIDATION))
+
     version = info.version
     ext_type = _infer_extension_type(info.metadata)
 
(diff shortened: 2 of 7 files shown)


jxstanford (changes requested):
# Review Report: feat(extensions): --digest flag on kz-ext publish for digest-pinned catalog refs (ENG-4370)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/94
**Branch**: `feature/ENG-4370` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| _none_ | | | | |

## High Priority Issues
| 1 | `kamiwaza_extensions/commands/publish.py:341-348` | `--no-push` regression: after a normal build with `--no-push`, `image_refs` exist only in the local daemon but the new digest block still calls `ImagePusher.resolve_digest(ref)`, which inspects the remote registry. If the tag was never pushed → command fails with `manifest unknown`. If the tag already existed remotely → catalog entry is pinned to the **old** remote manifest instead of the image just built. The `test_no_push_still_resolves_digest` test mocks past this. Either gate `resolve_digest` on `not no_push`, fail loudly with an actionable message, or capture the digest from the build/push output instead of a separate registry round-trip. | Codex (P1), Claude (related #5) | Logic / Regression |
| 2 | `kamiwaza_extensions/commands/publish.py:341-351` | Auto-resolve has no integrity check between push and inspect. With user-supplied `--digest`, the value is recorded verbatim (`digest_map[image_refs[0]] = digest`) without verifying it matches the manifest now in the registry — a typo or stale CI digest silently lands in the catalog as immutable identity. Fix: when `--digest` is set, also call `resolve_digest()` and reject on mismatch (at least when `--no-push` is false). | Claude | Logic / Data integrity |
| 3 | `kamiwaza_extensions/commands/publish.py:336-351` | TOCTOU window between `pusher.push()` and `ImagePusher.resolve_digest()`. If a parallel CI run re-points the tag between push and inspect, this run records the *other* run's digest. Mitigation: have `ImagePusher.push()` capture the manifest digest from buildx output (`--metadata-file` or parsing `digest: sha256:...`) instead of a separate inspect call. | Claude | Concurrency / Race |

| 4 | `kamiwaza_extensions/registry_builder.py:94-102` | If `metadata["extra_docker_images"]` repeats one of the extension's own service images, `build_entry()` rewrites only the compose-derived copy to `ref@sha256:...` and leaves the metadata copy as the original mutable `ref`. String-based dedup misses the duplicate, so `docker_images` ends up with both pinned and unpinned entries — regresses dedup and leaks an unpinned tag into the catalog. | Codex (P2) | Logic / Data integrity |
| 5 | `kamiwaza_extensions/image_pusher.py:107-131` | `resolve_digest` shells out to `docker buildx imagetools inspect` while `ImagePusher.push()` may have used Podman (`_has_podman()` / `insecure` path). A Podman-only host gets a misleading "docker not found" error after a successful Podman push. Either gate digest resolution to docker-only paths with a clearer message, or add a `podman manifest inspect` fallback. | Claude | Code quality / UX |
| 6 | `kamiwaza_extensions/commands/publish.py:344` | `image_refs[0]` is positional and assumes order matches the single buildable. Safe today (the `len(buildable_services) != 1` guard ensures one ref) but brittle if `_collect_image_refs` / `ImageBuilder.build` ever return extras (multi-arch, future aliases). Prefer keying by service-name → ref → digest end-to-end. | Claude | Logic / Maintainability |
| 7 | `kamiwaza_extensions/commands/publish.py:336-351` | When `--no-build --no-push` is combined with auto-resolve and the image isn't in the registry, `resolve_digest` errors with a generic "manifest unknown". Wrap with an actionable message: "registry has no image at <ref> — push it first or pass --digest". | Claude | UX / Error messages |
| 8 | `kamiwaza_extensions/registry_builder.py:264` | New regex enforces digest hex grammar (`@sha256:[a-f0-9]{64}`). A malformed digest already on the input (`@sha256:short`) leaves the unmatched tail in place, producing `…:newtag@sha256:short`. Probably the right behavior (preserve, don't corrupt), but worth a regression test. | Claude | Test gap |
| 9 | `kamiwaza_extensions/commands/publish.py:401` | Final summary prints `image_refs` (tag-only) rather than the digest-pinned refs that landed in the catalog. Operators don't see what was actually pinned. Consider `f"{ref}@{digest_map[ref]}"` when populated. | Claude | UX |
| 10 | `kamiwaza_extensions/registry_builder.py:264` | Pre-existing: fallback regex in `transform_image_tags` is unanchored and could match `image:` mid-line in env values/comments. Not introduced by this PR; the new digest-suffix capture inherits it. Worth a follow-up to anchor on `^\s+image:`. | Claude | Defensive (pre-existing) |

## Test Coverage Assessment
- **Strong on unit-level pure functions**: `_apply_digests`, `validate_digest`, `resolve_digest` (success, inspect failure, malformed output, missing docker), `transform_image_tags` (digest preservation in both extension-name and fallback branches), `build_entry` with `digest_map` (omitted/empty/partial maps, already-pinned, no-mutation, revision orthogonality). 31 new tests, broadly parametrized.
- **Gaps**:
  - Orchestration **failure path** when `ImagePusher.resolve_digest` raises `ImagePushError` — should exit 1 and skip catalog publish; not covered through `run_publish`.
  - Explicit `--digest` combined with `--no-push` (likely CI shape: re-pin without re-push) — only auto-resolve `--no-push` is tested.
  - Issue #1 (`--no-push` after build, image only local) — `test_no_push_still_resolves_digest` mocks `resolve_digest` to succeed, hiding the registry-vs-local discrepancy.
  - Issue #2 (digest-mismatch detection on user-supplied `--digest`) — uncovered because behavior doesn't exist yet.
  - Issue #4 (`extra_docker_images` repeating a buildable service) — uncovered.

## Positive Observations
- Validation is genuinely fail-fast: digest format checked before detection; buildable-count checked before any build/push side effects. Tests assert no `detect`/`build`/`push`/`resolve` calls on invalid inputs.
- `_apply_digests` is a pure function on a deepcopy — caller's compose dict isn't mutated, with explicit test.
- `transform_image_tags` regex change correctly preserves an already-attached digest in both branches; previous silent-strip was a real latent bug.
- Centralised `DIGEST_PATTERN` reused for both user input and tool output — sanity-checks `imagetools inspect` against malformed output.
- Already-pinned compose entries preserved (`if "@" not in img`), preventing double-pin from stale `digest_map`.
- `revision` and `digest` orthogonality documented and tested end-to-end.

## Overall Verdict
**Risk Level**: Medium
**Verdict**: REQUEST_CHANGES
**Engine Agreement**: 2/3 engines contributed (Claude APPROVE, Codex REQUEST_CHANGES — taking higher severity)

The `--no-push` regression (#1) is the load-bearing concern: existing flows that built locally and skipped push now either fail loudly or — worse — silently pin the catalog to a stale remote manifest. Combined with #2 (no integrity check on user-supplied `--digest`) and #4 (`extra_docker_images` duplicate-with-pin), the digest-pinning contract has correctness gaps in supported configurations. The implementation quality and test coverage on the happy paths are strong; addressing #1, #2, and #4 (or explicitly documenting/erroring on the unsupported shapes) should land this cleanly.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |
| Gemini | ⬜ Not available |

**Engines contributing**: 2/3
</details>

JohnStrick:
Addressed in 7a4dbe0 (and earlier commits since the review at 165cb7e). Per-finding:

**H1 — fixed.** Built-but-no-push without `--digest` now errors with an actionable message. Auto-resolve only allowed when there's a fresh push or in the catalog-only-republish shape. The test that encoded the broken behavior was replaced with one that asserts the rejection.

**H2 — fixed.** When `--digest` is supplied and push happened, `resolve_digest` is called and a mismatch aborts publish before catalog write. Skipped under `--no-push` (consistent with H1's escape hatch).

**H3 — deferred to ENG-4608.** Real concern but scoped out: H2's mismatch detection catches the race when `--digest` is supplied, which is the CI shape. The proper fix (`--metadata-file` digest capture from buildx) is an `ImageBuilder`/`ImagePusher` refactor and belongs in its own ticket.

**M1 — fixed.** `extra_docker_images` entries that duplicate a service ref are now pinned through the same `digest_map`, so dedup collapses the redundant copy. Regression test added.

**M2 — dismissed.** `publish` always passes `insecure=False`, which forces docker on both push and digest-resolution paths regardless of podman availability. The mismatch surfaces only if a future caller flips that flag, and that's a coordinated change. No code change.

**M3 — dismissed.** The single-buildable guard makes `image_refs[0]` safe today; threading service-name through end-to-end is a refactor for an unrealized risk. Not introducing churn here.

**M4 — fixed.** `manifest unknown` is now wrapped with `"registry has no image at <ref> — push it first or pass --digest"` under `--no-push`.

**M5 — test added.** Confirms `@sha256:short` (and other malformed suffixes) pass through verbatim.

**M6 — fixed.** Final summary prints `image:tag@sha256:...` so operators see what was pinned.

**N1 — deferred.** Pre-existing in `transform_image_tags`'s fallback regex, not introduced by this PR. Worth anchoring on `^\s+image:` separately.

Test gaps from the assessment are covered: orchestration failure path (`resolve_digest` raising), explicit `--digest` + `--no-push` combo, M1 collision, H2 mismatch detection, H1 build+no-push rejection. 6 new tests; suite at 1438 passing.

Thanks for the review.

JohnStrick:
Both addressed in 1455ce7.

**F1 — fixed.** `transform_image_tags` was corrupting digest-only refs (`image: <repo>@sha256:<64>`) — the repo-path char class allowed `@`, so the regex captured `service@sha256` as the path and the hex as the tag, rewriting to `service@sha256:<newtag>`. Added `@` to the path exclusion (`[^:\s@]+`) in both the extension and fallback branches. Digest-only refs now fail to match and pass through verbatim. Live-confirmed via REPL.

**F2 — fixed.** H1 condition now also requires `not dry_run`. The dry-run branch returns before any build/push/auto-resolve, so the stale-registry hazard cannot occur — there's no reason to force `--no-build` or `--digest` ceremony. Live-confirmed: `kz-ext publish --stage dev --dry-run --no-push` now proceeds past my checks.

3 new unit tests for F1 (digest-only in extension branch, fallback branch, mixed siblings) + 1 for F2 (dry-run + no-push allowed). 1442 passing.

jxstanford (changes requested):
# RE-REVIEW Report: feat(extensions): --digest flag on kz-ext publish for digest-pinned catalog refs (ENG-4370)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/94
**Branch**: `feature/ENG-4370` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)

## High Priority Issues
| 1 | publish.py:161-168 | The `--no-push` guard fires before discovering whether anything is buildable. Extensions whose compose only references external/prebuilt images (no `build:` services) — and CI wrappers that always pass `--no-push` — now exit with an error even though there is no local-only image to pin. Regresses valid publish-only flows. | Codex | Logic / Regression |
| 2 | publish.py:140-145 | The single-buildable enforcement is computed against raw `info.compose_data`, but `ComposeTransformer.transform()` strips services that declare a `profiles` key before catalog entry construction. An extension with one real buildable service plus a profiled dev-only helper is rejected as "found 2", blocking digest-pinned publishes for extensions that use local-only profile services. Count should be over the post-transform/published service set. | Codex | Logic / Regression |
| 3 | image_pusher.py:120-137 | `resolve_digest()` shells out to `docker buildx imagetools inspect`, and `run_publish(..., no_build=True, no_push=True)` calls it before publishing. Catalog-only republishes now hard-require Docker/buildx on the host even though those workflows neither build nor push. Multi-image extensions cannot work around this via `--digest` (rejected by issue #2's path as well). Consider an HTTP/registry-API fallback or skipping resolve when the user already supplied digests for every ref. | Codex | Dependency / Regression |

| 1 | image_pusher.py:124-138 | `subprocess.run` for `imagetools inspect` sets `timeout=60` but only catches `FileNotFoundError`; a wedged registry surfaces as an unhandled `subprocess.TimeoutExpired` rather than an `ImagePushError`. Consistent with sibling `_push`/`_login_registry` patterns, so not a regression — worth fixing across the module. | Claude Code | Error handling |
| 2 | publish.py:377 | H2 verification is skipped when `--no-push` is combined with explicit `--digest`. A CI typo in the digest against an existing tag goes undetected. Consider an opportunistic verify-with-warning when the ref is reachable in the registry. | Claude Code | Logic |

## Test Coverage Assessment
- New unit coverage is broad and thoughtful for the happy paths and explicitly-flagged hazards (H1/H2/M1/F1/F2): digest validation, resolve success/failure, registry_builder digest preservation across both regex branches, dedup with digest collision, dry-run pass-through, etc. (31 new tests).
- **Gaps surfaced by Codex's findings**:
  - No test for `publish --no-push` against a compose with no `build:` services (external-only extensions).
  - No test for `--digest` on an extension that has a profiled local-only helper alongside a single buildable service (post-transform vs raw service count).
  - No test exercising `--no-build --no-push` on a host without Docker/buildx (or otherwise verifying that publish-only does not require buildx).

## Positive Observations
- Author addressed every prior-round review concern with code, intent comments, and targeted regression tests across 6 follow-up commits (H1, H2, M1, F1, F2, error message safety).
- `_apply_digests` uses `copy.deepcopy` and skips already-pinned refs (`"@" not in img`) — preserves caller compose, idempotent under re-pin.
- `transform_image_tags` regex extended to include `@` in the path/tag class, so `image@sha256:<64hex>` no longer collides; a regression test pins this.
- `resolve_digest` correctly uses `{{json .Manifest}}` (not `.Manifest.Digest`, which would Stringer-dump the manifest) — gotcha called out in the docstring.
- `image_refs` keys produced by `ImageBuilder.build`, `_collect_image_refs`, and `ComposeTransformer` output are aligned, so `digest_map` keys reliably correlate with rewritten compose `image` values.
- All subprocess calls use list form (no shell injection surface).

## Overall Verdict
**Risk Level**: Medium
**Verdict**: REQUEST_CHANGES
**Engine Agreement**: 2/3 engines contributed (Claude Code: APPROVE; Codex: 3 P2 functional regressions). Per dedup rule, higher severity wins. The three Codex findings are concrete regressions in previously-supported publish flows (external-only extensions with `--no-push`, profile-filtered services with `--digest`, and publish-only republishes on hosts without Docker/buildx) and warrant fixes plus regression tests before merge.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |
| Gemini | ⬜ Not available |

**Engines contributing**: 2/3
</details>

JohnStrick:
Addressed in 7305ae1.

**HIGH 1 — fixed.** H1 now skipped when the compose has no buildable services. External-only extensions with `--no-push` publish without error.

**HIGH 2 — fixed.** Buildable count now mirrors `ComposeTransformer`'s filter (excludes services with `profiles:`). Profiled dev-only helpers no longer inflate the count and reject `--digest`.

**HIGH 3 — soft fallback (option A).** Under `--no-build --no-push`, `resolve_digest` failures are caught and the catalog falls back to tag-only for that ref with a warning. Build/push paths still hard-fail. The proper HTTP-registry-API fallback isn't being filed as a separate ticket right now per `@johnstrick`'s call — happy to revisit if the docker-less republish flow becomes a real CI shape.

**MEDIUM 1 — fixed (with sibling sweep).** `subprocess.TimeoutExpired` now caught and re-raised as `ImagePushError` in all three subprocess sites: `_login_registry`, `_push`, and `resolve_digest`.

**MEDIUM 2 — dismissed.** Designed: under `--no-push` the contract is "trust the supplied digest." An opportunistic verify-with-warning would add a network call and a half-trust mode without preventing the bad publish (warning ≠ rejection). The CI shape this matters for — push + `--digest` — already runs full H2.

**Note on commit message:** 7305ae1 references "ENG-4609" for an HTTP-fallback follow-up ticket that wasn't filed. The commit text overcommits — should read "deferred, no follow-up ticket filed."

Test coverage added for: external-only extension with `--no-push`, profiled-helper buildable count, `--no-build --no-push` resolve failure → tag-only fallback, subprocess timeout in all three sites. 1448 unit tests passing.

jxstanford (changes requested):
# RE-REVIEW Report: feat(extensions): --digest flag on kz-ext publish for digest-pinned catalog refs (ENG-4370)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/94
**Branch**: `feature/ENG-4370` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| _none_ | | | | |

## High Priority Issues
| 1 | `kamiwaza_extensions/commands/publish.py:381` (and `:141-145` guard) | Profiled-service filter inconsistency. `buildable_services` excludes services with `profiles:`, but `ImageBuilder.build()` (image_builder.py:62-66) and `_collect_image_refs()` (publish.py:61-72) do not. With one non-profiled buildable + one profiled buildable, `image_refs` has 2 entries and `image_refs[0]` may resolve to the profiled service depending on declaration order. Consequences: in the `--digest --no-push` path the user's digest is mapped onto the profiled ref, which `_apply_digests` then can't find in `transformed_compose` (profiled services are stripped) — catalog publishes the intended service tag-only, silently un-pinned. In the `--digest` + push path the H2 verify usually catches it but emits an error referencing the dev-helper ref. **Both engines agree.** Fix: filter `image_refs` (and the dry-run `_collect_image_refs`) by `not svc.get("profiles")`, or look up the ref by intersecting with `buildable_services`. | Claude, Codex | Logic / consistency |
| 2 | `tests/unit/extensions/test_publish_cmd.py:915-961` | `test_digest_with_profiled_helper_allowed` mocks `ImageBuilder.build` to return only the backend ref — the real builder would also return the profiled `dev-helper` ref. The test does not exercise the actual code path that motivates the fix and would not catch issue #1. Use a builder mock that returns both refs so the test matches production behaviour. | Claude | Test coverage |
| 3 | `kamiwaza_extensions/commands/publish.py:141-145` | The new pre-validation buildable-services scan runs before `ComposeValidator.validate()` and assumes `info.compose_data["services"]` is a mapping of name→dict. For YAML that parses but is structurally invalid (e.g. `services: []`, or a service whose body is a list), `.items()` / `svc.get(...)` will raise `AttributeError` and the command exits with an unhandled traceback instead of the normal validation error. Move this scan after validation, or guard the types. | Codex | Robustness / error reporting |

| 4 | `kamiwaza_extensions/image_pusher.py:122` | `import json as _json` is performed inside `resolve_digest`. Module-level import would be simpler — nothing here justifies a deferred import. | Claude | Code quality |
| 5 | `kamiwaza_extensions/image_pusher.py:92-95` | `FileNotFoundError` clauses in `_login_registry` and `_push` re-raise without `from`, while the new `TimeoutExpired` clauses use `from exc`. Inconsistent exception chaining for what is otherwise the same pattern (the `FileNotFoundError` form is pre-existing). | Claude | Code quality |
| 6 | `kamiwaza_extensions/commands/publish.py:402` | The H2 mismatch path raises `ExitCode.VALIDATION` after a successful push — registry already mutated, catalog rejected. The right call, but consider including the just-pushed tag in the error so the operator can decide whether to roll back. | Claude | UX / error message |

## Test Coverage Assessment
- **Adequate**: `image_pusher.py` (validate_digest happy/sad, resolve_digest output paths incl. timeout, JSON parse failure, missing/malformed digest field, missing docker), `registry_builder.py` (digest-suffix preservation in both regex branches, digest-only ref pass-through, malformed digest pass-through, deepcopy non-mutation, dedup with extra_docker_images, revision×digest orthogonality), `commands/publish.py` (auto-resolve happy path, multi-buildable rejection, format validation, H1/H2 hazards, external-only allowed, dry-run with/without digest, no-build no-push resolution, soft-fallback on resolve failure, revision×digest).
- **Gaps**:
  - Profiled-buildable + `--digest` interaction with a builder mock that returns refs for the profiled service too (issues #1/#2).
  - Malformed-but-parseable compose shapes (e.g. `services: []`) hitting the new pre-validation scan (issue #3).
  - Whether `resolve_digest` is contractually expected to rely on prior `docker login` (no test asserts it does not attempt its own auth).

## Positive Observations
- Re-review feedback was systematically addressed: H1/H2/H3 hazards, M1 timeouts, M4 partial-republish hint, M5 malformed-suffix pass-through, F1 digest-only refs, F2 dry-run-honors-digest, rich-markup escaping in error messages — each has a corresponding test.
- `_apply_digests` deep-copies the compose dict so caller data isn't mutated; verified by `test_does_not_mutate_caller_compose`.
- TOCTOU window between push and publish is explicitly handled by H2 verification — good defensive thinking.
- `extra_docker_images` dedup-collision case catches a subtle catalog-leak that would otherwise be invisible.
- The `transform_image_tags` regex change correctly excludes `@` from path/tag character classes so digest-only refs can't be corrupted by greedy capture; both branches and both directions are tested.
- Use of `{{json .Manifest}}` (rather than `.Manifest.Digest`) is documented in code with the rationale — saves the next maintainer from debugging the Stringer-vs-JSON pitfall.
- Subprocess timeouts on every external call are appropriate for CI use.

## Overall Verdict
**Risk Level**: Low

Both engines independently identified the same residual logic gap (profiled-service filtering inconsistency between `buildable_services` and `image_refs`). The blast radius is narrow: it requires an uncommon compose shape (a profiled buildable alongside a real buildable), and is mostly caught by the H2 verify in the push path; CI flows that motivated this PR use auto-resolve rather than `--digest`. Codex additionally surfaced a robustness gap where the pre-validation scan can crash on structurally invalid compose files. Neither is a release blocker, but both warrant a follow-up tightening (filter `image_refs` to non-profiled services, move/guard the pre-validation scan, and a non-mocked test for the profiled-helper case).

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |
| Gemini | ⬜ Not available |

**Engines contributing**: 2/3
</details>

JohnStrick:
Two more rounds of fixes landed since 7305ae1 — point by point:

**5238e40 — `--digest` pinning was still using `image_refs[0]`** (caught by local Codex review)

The earlier HIGH 2 fix filtered the *count* against `buildable_services`, but `digest_map[image_refs[0]] = digest` still walked `_collect_image_refs` output (unfiltered). With a profiled helper first in compose dict order, the user's digest could be pinned against / H2-verified against the wrong (local-only) ref. Fixed by deriving `published_refs` from the already-filtered `buildable_services` and using it for both the pin target and the auto-resolve loop. Final summary line also switched to `published_refs` for catalog-accurate output. New regression tests cover dict-order traps in both auto-resolve and explicit-`--digest` flows.

**665f66a — local-review (3 Claude reviewers) findings**

- **Critical: dry-run `--digest` had the same dict-order trap as the live path.** Same root cause as 5238e40, in the dry-run branch. Replaced `_collect_image_refs(...)` with the canonical-ref derivation. Regression test added with `dry_run=True` and profiled-helper-first compose.
- **`resolve_digest` `AttributeError` on non-dict JSON.** Valid JSON whose top level wasn't a dict (e.g. `[]`) hit `manifest.get(...)` and raised `AttributeError`, not `ImagePushError`. Added `isinstance(manifest, dict)` guard. Test added.
- **Markup-injection in `--digest` format error.** The error message echoed user input through rich's markup parser; a `[bracket]` in the bad digest got silently stripped. Switched to `markup=False` for the user-input portion. Live-verified with `sha256:[bracket]injection`.
- **Auto-resolve block was 6-deep, exceeding `.ai/rules/style.md`'s 4-level cap.** Extracted `_verify_supplied_digest` and `_auto_resolve_digests` module helpers; the call site is now flat. Also separates the verify-vs-resolve concerns at the type/function boundary.
- **Stripped review-round labels (`H1:`, `H2:`, `M4:`, `P2`, `Re-review HIGH N`, `F1`/`F2`) from comments.** Comments document invariants for future maintainers, not the PR's review history.

Deferred (not in this PR): rebuilding `published_refs` from `transformed_compose["services"][svc].image` (currently safe — `ComposeTransformer` always rewrites to `<reg>/<ext>-<svc>:<tag>`); auto-resolve TOCTOU documentation (covered by ENG-4608); podman parity for `resolve_digest` (theoretical — `publish` always sets `insecure=False`); test-decorator boilerplate refactor.

1452 unit tests passing, ruff clean, live smokes verified the markup fix, dry-run digest pin, and the canonical-ref pin in CLI invocation.

JohnStrick:
Consolidated status across both review rounds:

**First review (at 165cb7e)**
- H1 (--no-push hazard) — fixed in 7a4dbe0; strengthened in 7305ae1 (skip when no buildable services) and 1455ce7 (exempt --dry-run)
- H2 (--digest integrity check) — fixed in 7a4dbe0; deeper layer landed in 5238e40 (was still using `image_refs[0]` even after filtering count) and 665f66a (dry-run path had the same gap)
- H3 (TOCTOU between push and inspect) — deferred to ENG-4608; H2's mismatch detection is the partial mitigation
- M1 (extras dedup) — fixed in 7a4dbe0
- M2 (podman parity) — dismissed: `publish` always passes `insecure=False`, so docker is forced for both push and digest resolution
- M3 (image_refs[0] positional) — dropped per author judgment: the single-buildable guard makes positional indexing safe today
- M4 (actionable manifest-unknown) — fixed in 7a4dbe0
- M5 (test gap on malformed digest suffix) — fixed in 165cb7e
- M6 (summary shows pinned refs) — fixed in 7a4dbe0; summary now derives from `published_refs` (665f66a)
- N1 (pre-existing fallback regex anchoring) — deferred (pre-existing, separate from ENG-4370)

**Re-review (at 1455ce7)**
- HIGH 1 (external-only extension + --no-push) — fixed in 7305ae1
- HIGH 2 (profile-filter buildable count) — fixed in 7305ae1; deeper layer in 5238e40 + 665f66a (above)
- HIGH 3 (catalog-only republish required docker) — fixed in 7305ae1 with soft fallback (option A)
- MEDIUM 1 (subprocess.TimeoutExpired) — fixed in 7305ae1, swept across all three subprocess sites
- MEDIUM 2 (opportunistic verify under --digest + --no-push) — dismissed: under `--no-push` the contract is "trust the supplied digest"; verify-with-warning adds a network call without preventing the bad publish

**Additional (from local review, 665f66a)** — not flagged by either JohnBot round but worth noting:
- `resolve_digest` AttributeError on non-dict JSON → now `ImagePushError`
- `--digest` format error was leaking user input through rich markup → `markup=False`
- Auto-resolve block was 6-deep, over the project's 4-level cap → extracted `_verify_supplied_digest` and `_auto_resolve_digests`
- Review-round labels stripped from comments (per project guidance: comments document invariants for future maintainers, not the PR's review history)

1452 unit tests passing, ruff clean.

jxstanford (changes requested):
# RE-REVIEW Report: feat(extensions): --digest flag on kz-ext publish for digest-pinned catalog refs (ENG-4370)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/94
**Branch**: `feature/ENG-4370` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| – | – | None | – | – |

## High Priority Issues
| 1 | `kamiwaza_extensions/image_pusher.py:124-137` | `resolve_digest()` unconditionally requires `docker buildx imagetools inspect`, which is now reached by both auto-resolve and `--digest` verification on every push-enabled publish. Hosts that previously supported `docker build`/`docker push` but lack the buildx plugin (older Docker packages, some Docker-compatible CLIs) will now fail *after* the image has already been pushed — registry mutated, catalog publish aborted. Add a preflight check (fail before push if buildx is missing) or a fallback (e.g., `docker manifest inspect`) so this isn't a silent compatibility regression. | Codex | Compatibility / regression |

| 1 | `kamiwaza_extensions/commands/publish.py:132-136` | The `if no_push and "manifest unknown" in str(exc).lower():` branch in `_auto_resolve_digests` is unreachable. By the time this function runs, `digest is None`; H1 (line 237-250) blocks `not no_build and no_push and digest is None and buildable_services`, so when `no_push=True`, only `no_build=True` can reach this loop, and the `if no_build and no_push:` branch above (line 125) `continue`s before line 132 is evaluated. The hint message also duplicates H1's guidance. Either drop the branch or generalize the hint for an actually-reachable failure mode (e.g., parallel-actor TOCTOU after push). | Claude Code | Dead code |
| 2 | `kamiwaza_extensions/image_pusher.py:175-177` | Pre-existing, but worth noting since timeouts were touched: the `verbose=True` push uses `check=True` and only catches `FileNotFoundError`/`TimeoutExpired`. A non-zero exit in verbose mode raises an uncaught `CalledProcessError`. Out of scope for this PR; consider a follow-up. | Claude Code | Error handling |
| 3 | `kamiwaza_extensions/registry_builder.py:106-112` | `extra_docker_images` digest-pinning matches against post-rewrite refs in `digest_map`. If a user lists a service ref without the stage suffix (e.g. `kamiwazaai/my-app-backend:1.0.0` while compose carries `1.0.0-dev`), the lookup misses and the unpinned ref leaks into the catalog. Tests cover the matching case only. Either tighten with a normalize pass or document the constraint explicitly. | Claude Code | Logic / contract |

## Test Coverage Assessment
- **Adequate**: `image_pusher.py` (`validate_digest` parametrized accept/reject; `resolve_digest` success, inspect failure, non-dict JSON, invalid JSON, missing/malformed digest field, missing docker, timeout; push/login timeout handlers). `registry_builder.py` (digest preservation in extension and fallback regex branches, digest-only refs untouched, malformed digest pass-through, `_apply_digests` non-mutation, partial map, already-pinned skip, extras dedup collision, revision×digest orthogonality). `commands/publish.py` (auto-resolve, multi-buildable rejection, build+no-push hazard, external-only allowance, profiled-helper exclusion, canonical-ref pinning when profiled helper appears first in dict order, soft-fallback under `--no-build --no-push`, dry-run preview honoring digest, mismatch verification, registry-resolve failure aborts, dry-run profiled-first canonical pin).
- **Gaps**: No coverage for the buildx-missing scenario flagged in High #1 (no test that publish fails-fast vs. fails-after-push when buildx is absent). `_auto_resolve_digests` "manifest unknown" branch has no test (pending decision on dead code). No test that `markup=False` neutralizes a malicious-looking `--digest` payload at console render time (low priority).

## Positive Observations
- Strong fail-fast discipline: digest grammar validation runs before extension detection / image build / push, so bad input never produces side effects.
- `_verify_supplied_digest` closes a real TOCTOU window; testing it against the canonical buildable ref (not `image_refs[0]`) avoids the dict-order trap when a profiled helper builds first — the test `test_digest_pins_canonical_ref_when_profiled_helper_appears_first` explicitly guards the regression.
- Mirroring `ComposeTransformer`'s `profiles:` filter when computing `buildable_services` correctly aligns digest-pinning with what's actually published.
- Regex update in `transform_image_tags` is conservative and well-tested: digest suffixes survive rewrite; digest-only refs (no tag) pass through; malformed `@sha256:short` suffixes pass through verbatim instead of corrupting; the `[^:\s@]+` exclusion of `@` is the correct fix for the original ambiguity.
- `_apply_digests` deep-copies, doesn't double-pin, and skips non-buildable refs (verified by `test_does_not_mutate_caller_compose` and `test_already_pinned_ref_not_double_pinned`).
- JSON-template approach (`{{json .Manifest}}`) with shape, type, and grammar checks is robust against buildx output drift.
- Markup-injection risk in error messages is explicitly addressed (`markup=False`).
- 31 new tests, well-targeted at trap cases (profiled-first, dict order, dry-run preview, soft-fallback, mismatch).

## Overall Verdict
**Risk Level**: Medium
**Verdict**: REQUEST_CHANGES

Rationale: Code quality is strong and the contract for digest pinning is well-designed and well-tested. The blocker is High #1 — `resolve_digest` introduces a hard `docker buildx` dependency on the post-push path, so environments without buildx will now mutate the registry and then fail the catalog write. A preflight (fail before push when buildx is unavailable) or a `docker manifest inspect` fallback would resolve this; both options are small. The Medium items are non-blocking polish.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |
| Gemini | ⬜ Not available |

**Engines contributing**: 2/3
</details>

JohnStrick:
Addressed in 6a72c00.

**HIGH 1 (`resolve_digest` requires buildx on post-push path)** — fixed. Added `ImagePusher.check_buildx_available()` and called it before the push step, gated on `not no_push and buildable_services` (i.e., when downstream digest resolution will actually be required). If buildx is unavailable the publish aborts before pushing, so the registry-catalog divergence Codex flagged can't happen. Skipped under `--no-build --no-push` because that path already soft-falls in `_auto_resolve_digests`. The check is one cheap subprocess call (`docker buildx imagetools --help`, 5s timeout); modern docker users see no behavior change.

**MEDIUM 1 (dead `manifest unknown` branch)** — fixed. The reachability analysis is correct: by the time `_auto_resolve_digests` runs, `digest is None`; H1's earlier guard blocks the only state shape that could reach the branch without `no_build` also being set; and the `if no_build and no_push:` soft-fall above `continue`s before the dead branch is evaluated. Dropped the branch.

**MEDIUM 2 (`_push(verbose=True)` doesn't catch `CalledProcessError`)** — deferred. Pre-existing on develop; Codex's own framing was "out of scope for this PR." Noting for a follow-up sweep across the subprocess call sites in `image_pusher.py`.

**MEDIUM 3 (`extra_docker_images` stage-suffix mismatch)** — addressed by docstring rather than code. The match is intentionally exact-string against the post-stage-suffix ref (e.g. `<reg>/<ext>-<svc>:<version>-dev`), and authoring guidance lives in the `build_entry` docstring now. Normalizing in `RegistryBuilder` would require knowing the stage suffix at the metadata layer, which is a layer inversion. The realistic incidence (user redundantly listing their own buildable service in `extra_docker_images` with the pre-suffix tag) is narrow enough that explicit-author guidance feels right.

7 new unit tests for the buildx preflight (4 for `check_buildx_available` directly, 3 for the wiring in `run_publish`). 1459 unit tests passing, ruff clean.

jxstanford (approved):
# RE-REVIEW Report: feat(extensions): --digest flag on kz-ext publish for digest-pinned catalog refs (ENG-4370)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/94
**Branch**: `feature/ENG-4370` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)

(none)

## High Priority Issues

(none)

| 1 | commands/publish.py:203–206 | `info.compose_data.get("services")` is dereferenced before `ComposeValidator.validate()` runs. If compose parses to a non-mapping root (YAML list or scalar), this raises `AttributeError` instead of surfacing the existing "Compose file must be a YAML mapping" validation error — a regression versus prior behavior. Reorder so structural validation runs first, or add a `isinstance(compose_data, dict)` guard before iterating buildable services. | Codex | Robustness |
| 2 | commands/publish.py:227–240 | The new H1 guard turns `kz-ext publish --no-push` (no `--no-build`, no `--digest`, has buildable services) into a hard error. This is a behavioral change for users/CI scripts who previously used `--no-push` to publish catalog-only after a separate push step. Error message lists alternatives (`--no-build`, supply `--digest`, push first). Worth calling out in CHANGELOG/release notes. | Claude | Backward compat |
| 3 | image_pusher.py:115–135 | `check_buildx_available` runs `docker buildx imagetools --help` with `capture_output=True`, discarding stderr. On odd installs (buildx without imagetools subcommand), the failure message would be more diagnostic if it included `result.stderr`. Polish only. | Claude | Error msg quality |
| 4 | commands/publish.py:75–101 | `_verify_supplied_digest` calls `resolve_digest` (uses buildx). The preflight at L417–422 only runs `if not no_push and buildable_services` — which is exactly when verification triggers, so the dependency holds. Add a short inline note tying these two together so a future refactor doesn't move one without the other. | Claude | Maintainability |

## Test Coverage Assessment
- **Adequate coverage**: `commands/publish.py` (digest format, multi-buildable rejection, H1 hazard, external-only allow, profiled-helper edge cases, dry-run pinning, mismatch detection, TOCTOU verify, soft-fallback under `--no-build --no-push`, revision×digest orthogonality), `image_pusher.py` (validate_digest grammar, resolve_digest happy/error/timeout/non-dict-JSON/missing-field/malformed-digest, check_buildx_available all branches, push/login timeout), `registry_builder.py` (digest preservation, digest-only refs untouched, partial map, already-pinned refs not double-pinned, extras dedup, revision+digest combo, no caller mutation).
- **Gap**: No test exercising the malformed-compose path (M1) — non-mapping root YAML would currently crash before validation.

## Positive Observations
- Re-review feedback fully addressed: buildx preflight (6a72c00), canonical-ref pinning instead of `image_refs[0]` (5238e40), dry-run digest map honoring `buildable_services[0]` (665f66a), markup-safe error message (665f66a), JSON guard on imagetools output (665f66a).
- TOCTOU verify on supplied digests catches CI typos, stale digests, and parallel-publish races before catalog write.
- `_apply_digests` deep-copies before mutation; `"@" not in img` prevents double-pinning.
- Profile filtering applied consistently — `buildable_services` mirrors `ComposeTransformer`'s `profiles:` strip, aligning `published_refs` and pinning loop with what reaches the catalog.
- Subprocess timeouts uniform (login 30s, buildx check 5s, resolve 60s, push 600s) — wedged commands surface as `ImagePushError` rather than hangs.
- `transform_image_tags` regex correctly excludes `@` from path/tag char classes so digest-only refs can't be re-tagged into corruption.

## Overall Verdict
**Risk Level**: Low
**Verdict**: APPROVE

Rationale: All previous review concerns addressed; comprehensive test coverage including tricky edge cases. M1 (Codex) is a real but narrow robustness gap on malformed YAML — worth a follow-up, not blocking. M2 (`--no-push` rejection) is a deliberate fix for a pre-existing latent bug with clear alternatives — flag in release notes.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |
| Gemini | ⬜ Not available |

**Engines contributing**: 2/3
</details>

PR #100, opened 2026-05-11
fix(extensions): kz-ext publish reads docker-compose.appgarden.yml when present (ENG-4907)

JohnStrick:
## Summary

- `kz-ext publish` now treats `docker-compose.appgarden.yml` as the source of truth for catalog publishing when present, replacing today's behavior of reading `docker-compose.yml` and applying its own `ComposeTransformer` (which bypasses each extension's `sync-compose.py` transformations).
- Only image tags for services with a `build:` block in source `docker-compose.yml` are rewritten, so `--stage`/`--revision` still pin the CI-built image. External refs (e.g. `ghcr.io/.../neo4j`) and any service the extension's `sync-compose.py` invented pass through verbatim.
- Falls back to today's `ComposeTransformer` behavior against `docker-compose.yml` when no appgarden file is present (or when it's malformed); a warning is surfaced in the malformed case.

## Why

Per [ENG-4907](https://linear.app/kamiwaza/issue/ENG-4907): each extension's `sync-compose.py` carries extension-specific transformations (env-shape tweaks, hostname rewrites, network policy hints, etc.) that kz-ext doesn't and can't replicate. For graphiti, three transformations were silently being dropped by kz-ext's generic transform; defensive shims in graphiti's runtime saved it, but other extensions in the auto-publish rollout queue (ENG-4612/4613/4615/4616/4638) wouldn't be so lucky. This unblocks fleet adoption of `extension-build.yml@main`.

## Design choice

`kz-ext publish` is a thin layer over the extension's authored deployment compose; the `build:`-context lookup against source compose is the line between "extension's domain (any shape in appgarden.yml)" and "publish's domain (tag for what we just built)." Sibling commands (`kz-ext dev`, `validate`, `dev_local`, `port_forward`, `shell`, `logs`, `bump`, `status`, `config`) keep reading `docker-compose.yml` — they need build contexts and host port mappings that only live there.

## Test plan

- [x] `tests/unit/extensions/test_publish_cmd.py` — 11 new tests across 3 classes:
  - `TestLoadAppgardenCompose`: file present / missing / malformed YAML / non-mapping top-level
  - `TestRetagAppgardenCompose`: retag build-context service / passthrough external image / graphiti-shaped mixed services with env-placeholder preservation / sync-compose-only services / `--revision`-shaped tag / no source mutation
  - `TestPublishWithAppgarden`: end-to-end — appgarden bypasses `ComposeTransformer`, validator called against appgarden path; fallback path unchanged when no appgarden file
[...]

PR #101, opened 2026-05-11
fix(extensions): kz-ext publish honors compose image namespace + fails loudly on digest miss (ENG-4909)

JohnStrick:
## Summary

Multiple sites in `kz-ext` independently synthesized the canonical image ref for build-context services. They needed to agree, otherwise the catalog ships referencing one path while the image was pushed to another — and any consumer pulling the catalog ref gets an unpullable image. After this PR, a single shared helper (`compute_canonical_refs`, layered on `_canonical_build_ref`) is the source of truth across every site (publish live + dry-run + dev), digest-resolution failures abort the publish instead of silently degrading to tag-only catalog entries, and unqualified declared images correctly fall back to the cluster-registry rewrite instead of leaking pushes to Docker Hub.

Linear: [ENG-4909](https://linear.app/kamiwaza/issue/ENG-4909)

## What this does

1. **`_retag_appgarden_compose` preserves declared namespace.** Reads compose's `image:` and rewrites only the tag.
2. **`_auto_resolve_digests` aborts on resolution failure.** Previously soft-failed to tag-only entries in `--no-build --no-push` mode; now errors with an actionable message naming the failed ref and the common causes. Soft-fall hid the omniparse-style mismatch this PR is designed to surface.
3. **`_canonical_build_ref` centralizes the namespace logic** in `compose_transformer.py`. Reads the service's declared `image:` ONLY when registry-qualified (first slash-separated component contains `.` or `:`, or is exactly `localhost` — docker's standard rule). Unqualified refs (`api:latest`, `my-org/api:1.0`) and missing/empty declarations fall back to legacy `{registry}/{ext}-{svc}:{tag}`, preventing `docker push` from silently routing to Docker Hub while the cluster registry expects the rewrite. Strips whitespace.
4. **`compute_canonical_refs` is the single shared derivation** used by publish (live + dry-run) and dev. Same source of truth means none of the three paths can drift on namespace, profile filtering, or appgarden precedence. Lives in `compose_transformer.py`; thin wrappers in `commands/publish.py` and `commands/dev.py`.
5. **Per-service appgarden precedence is presence-based.** When `docker-compose.appgarden.yml` declares a service entry — even an empty mapping — that's the source of truth (matches `_retag_appgarden_compose`'s behavior). Source-compose entry is the fallback for services the appgarden file omits. Switching to `name in appgarden` instead of `appgarden.get(name) or svc` keeps the two paths in lockstep on the empty-mapping case.
6. **Profile-gated services are excluded from `canonical_refs`** so they don't leak into the `--no-build` push list or the dev push path. Mirrors `buildable_services`' filter.
7. **`ImageBuilder.build` gains an `image_refs` kwarg** so callers pass the canonical map. Both `run_publish` and `run_dev_remote` compute via `compute_canonical_refs` and pass it through.

## Commits

| Commit | Scope |
|---|---|
| `fix(extensions): preserve appgarden compose image namespace on retag` | Primary catalog-namespace fix. `_retag_appgarden_compose` reads declared image and rewrites only the tag. Adds the original `_replace_image_tag` helper. |
| `fix(extensions): kz-ext publish aborts on digest resolution failure` | Hard-fail in `_auto_resolve_digests` (was soft-fall in `--no-build --no-push`). Actionable error message naming the failed ref. `rich.markup.escape` on dynamic parts. |
| `refactor(extensions): move _replace_image_tag to compose_transformer module` | Pure refactor — helper relocated so `ComposeTransformer` and `ImageBuilder` can import without depending on `commands/`. Behavior unchanged. |
| `fix(extensions): derive build refs and published_refs from canonical helper` | Threads the canonical helper through `ComposeTransformer.transform_service`, `ImageBuilder.build`, `_collect_image_refs`, `published_refs`, and `dry_canonical_ref`. `run_publish` computes a `canonical_refs` map once and passes it everywhere downstream. End-to-end test asserts the namespace-divergence flow through build+push+resolve. |
| `chore(extensions): update stale soft-fall references + generic registry wording` | Drive-by: stale "soft-fall" comments updated to reflect the new hard-fail; error message "GHCR path" → "registry image path". |
| `fix(extensions): kz-ext dev builds at canonical image refs` | Aligns `run_dev_remote` with `run_publish`. Without canonical_refs threaded through `ImageBuilder.build`, the builder synthesized the legacy `{ext}-{svc}` form — pods hit `ImagePullBackOff` on extensions with non-conventional paths. Adds `test_dev_canonical_refs.py`. |
| `fix(extensions): publish canonical_refs honors appgarden image namespace` | Appgarden takes precedence for the live `canonical_refs` map. Also: strip whitespace in `_canonical_build_ref`; drop `_collect_image_refs` in favor of `list(canonical_refs.values())`; rename two same-named tests across files. |
| `fix(extensions): only honor declared image namespace when registry-qualified` | **(H1 from re-review)** Adds `_looks_registry_qualified`. Unqualified short refs (`api:latest`, `my-org/api:1.0`) now fall back to legacy `{registry}/{ext}-{svc}:{tag}` instead of being honored verbatim — previously they would have leaked `docker push` to Docker Hub. |
| `fix(extensions): share canonical_refs derivation across publish + dev paths` | **(M1+M2+M3 from re-review)** Extracts `compute_canonical_refs` shared by all three sites. Presence-based appgarden lookup (M1), appgarden-aware dry-run (M2), profile-gated services filtered out (M3). |

## Behavioral changes (for review attention)

1. **`_auto_resolve_digests` hard-fails by default** (no opt-in flag). Previously soft-failed in `--no-build --no-push` mode. Escape hatches remain: `--digest sha256:...`, `--dry-run`, legacy `make publish-registry` (retiring).

2. **`ComposeTransformer.transform_service` now respects registry-qualified declared `image:` namespace.** Previously always synthesized `{registry}/{ext}-{svc}:{tag}` regardless of compose; now honors qualified declarations but still rewrites unqualified ones to the cluster registry. Same fix applies to `_retag_appgarden_compose` and the dev path.

3. **`kz-ext dev` now builds at the declared compose namespace.** Previously dev.py's `ImageBuilder` call defaulted to the legacy `{ext}-{svc}` form, while the K8s payload (built from the transformed compose) referenced the declared namespace — guaranteed `ImagePullBackOff` on any extension that publishes under a non-conventional path.

4. **Unqualified declared images fall back to legacy rewrite, not Docker Hub.** Compose shapes like `image: api:latest` or `image: my-org/api:1.0` would have been honored verbatim by `_canonical_build_ref` before the H1 fix from the re-review — `docker push` resolves those to Docker Hub. They now fall back to `{cluster_registry}/{ext}-{svc}:{tag}` so the rewrite stays internally consistent.

### Unit tests
- [x] `uv run pytest tests/unit/extensions/` — 1151 pass, 1 skipped (BSD-sed-only; unrelated), 1 deselected (`test_available_port` is brittle to which ports are taken on the host; unrelated to this PR).
- [x] `uv run pytest tests/unit/extensions/test_publish_cmd.py tests/unit/extensions/test_compose_transformer.py tests/unit/extensions/test_dev_canonical_refs.py` — 124 pass. New tests since the original PR:
  - `TestLooksRegistryQualified` (5 cases) + `TestCanonicalBuildRef` (8 cases) — H1 predicate and the canonical-build-ref behavior across qualified/unqualified shapes
  - `TestComputeCanonicalRefs` (8 cases) — shared helper: empty source, only-buildable, profile filter, appgarden override, empty appgarden entry, missing appgarden entry, None source, legacy fallback
  - `TestBuildContextRemoval::test_unqualified_short_form_falls_back_to_legacy` — H1 regression-shape end-to-end through `ComposeTransformer.transform`
  - `TestPublishDigest::test_dry_run_digest_uses_appgarden_namespace` — M2 dry-run + `--digest` + appgarden namespace divergence
  - `TestPublishWithBuild::test_no_build_push_excludes_profile_gated_services` — M3 `--no-build` push list excludes profile-gated refs
  - `TestPublishWithAppgarden::test_appgarden_image_namespace_drives_build_and_push` — appgarden namespace flows to ImageBuilder + digest resolve
  - `TestDevRemoteBuildsAtCanonicalRefs::test_divergent_image_namespace_flows_through_to_builder` — `run_dev_remote` passes canonical refs to `ImageBuilder.build`
- [x] `uv run ruff check` on changed files — clean.

### E2E validation (on omniparse, the real-world divergent-namespace case)

omniparse declares `image: ghcr.io/kamiwaza-internal/kamiwaza-extensions-omniparse/images/omniparse:2.0.14` in both source and appgarden compose. Ran against the local kamiwaza.test cluster and the R2 dev catalog using `kz-ext` built from this branch:

| Step | Result |
|---|---|
| **1. `kz-ext publish --stage dev --dry-run --force`** | Dry-run completes; catalog write would replace v2.0.14. canonical_refs computed correctly. (Minor cosmetic finding: dry-run "Would build images:" display still uses legacy short form — actual build/push uses canonical_refs. Captured as a follow-up.) |
| **2. K8s payload assertion** | Test deploy with `--revision eng-4909-validate`. Pod spec confirmed: `image: ghcr.io/kamiwaza-internal/kamiwaza-extensions-omniparse/images/omniparse:eng-4909-validate` — canonical declared namespace. Existing 41h-old omniparse pod at same canonical ref is Running, proving cluster pulls successfully when image exists. Full build skipped (mechanical; unit test covers `image_refs=` kwarg wire-up). |
| **3. `kz-ext publish --stage dev --no-build --no-push --force`** | Published omniparse v2.0.14 to R2 dev catalog. Downloaded `dev-info/garden/v3/tools.json`: `compose_yml.services.omniparse-server.image` and top-level `docker_images[0]` both contain `ghcr.io/kamiwaza-internal/kamiwaza-extensions-omniparse/images/omniparse:2.0.14-dev@sha256:a2f1c7466baf3f6161701e9adaaeb6e9e94fa04f10248f3435bed8397b1fc01b`. |
| **4. Direct Python invocation post-H1 fix** | `_looks_registry_qualified` on real omniparse compose returns True for `ghcr.io/...`, False for `api:latest`/`my-org/api:1.0`, True for `localhost:5000/foo`/`localhost/foo`. `compute_canonical_refs` against omniparse's real compose produces `ghcr.io/.../images/omniparse:{tag}` for both publish (with appgarden) and dev (without). |

## Related

- ENG-4638 — wired up auto-publish; surfaced this gap
- ENG-4872 — fixed `:latest` race in container-build (related infra)
- ENG-4910 — closed as won't-do; was based on the misread that omniparse needed to rename
- PR #100 (ENG-4907) — landed `_retag_appgarden_compose` with the catalog-write-side bug fixed in commit 1 here

## Follow-ups (not blocking this PR)

- **Dry-run display synthesizes legacy short form.** `_collect_buildable_image_names` (used only for the "Would build images:" message in dry-run output) still produces `{ext}-{svc}:tag`. The actual build path uses `canonical_refs` correctly; the display is just misleading for extensions with declared namespaces. Cosmetic-only.

### DIFF
--- a/kamiwaza_extensions/commands/dev.py
+++ b/kamiwaza_extensions/commands/dev.py
@@ -314,7 +314,6 @@ def run_dev_remote(
         DeploymentTimeoutError,
     )
     from kamiwaza_extensions.dev_state import (
-        DevState,
         mark_step,
         read_state,
         resume_message,
--- a/kamiwaza_extensions/commands/dev.py
+++ b/kamiwaza_extensions/commands/dev.py
@@ -305,7 +305,7 @@ def run_dev_remote(
 
     from kamiwaza_extensions.compose_transformer import (
         ComposeTransformer,
-        _canonical_build_ref,
+        compute_canonical_refs,
     )
     from kamiwaza_extensions.connections import ConnectionManager
     from kamiwaza_extensions.deployment_poller import (
@@ -435,22 +435,18 @@ def run_dev_remote(
     )
     transformed = transformer.resolve_env_placeholders(transformed)
 
-    # Canonical image refs for every build-context service. Single source
-    # of truth shared with the transformed compose so the image we build
-    # and push matches the ref the K8s payload will pull (ENG-4909). The
+    # Canonical image refs for every build-context service. Single
+    # source of truth shared with the transformed compose so the image
+    # we build and push matches the ref the K8s payload will pull. The
     # transformer honors a service's declared image namespace; without
-    # this map ImageBuilder would synthesize the legacy {ext}-{svc} form
-    # and ship a pod referencing an image that was never pushed.
-    canonical_refs: Dict[str, str] = {
-        name: _canonical_build_ref(
-            svc, name,
-            fallback_registry=registry,
-            fallback_extension_name=info.name,
-            revision_tag=rev_tag,
-        )
-        for name, svc in (info.compose_data.get("services") or {}).items()
-        if "build" in svc
-    }
+    # this map ImageBuilder would synthesize the legacy {ext}-{svc}
+    # form and ship a pod referencing an image that was never pushed.
+    canonical_refs: Dict[str, str] = compute_canonical_refs(
+        info.compose_data.get("services") or {},
+        registry=registry,
+        extension_name=info.name,
+        revision_tag=rev_tag,
+    )
 
     # 5b. Resolve SDK override for build
     build_overrides = None
--- a/kamiwaza_extensions/commands/dev.py
+++ b/kamiwaza_extensions/commands/dev.py
@@ -69,26 +69,29 @@ def _build_patch_service_specs(payload: Any) -> List[Any]:
     ``CreateExtension`` payload, forwarding the new ``x-kamiwaza``
     per-service overrides via ``extra="allow"``.
 
-    jxstanford PR #97 review H2: without forwarding ``healthCheck`` /
-    ``automountServiceAccountToken`` / ``containerSecurityContext`` on
-    PATCH, iterative ``kz-ext dev`` redeploys silently drop the very
-    overrides this PR added.
+    The PATCH carries the full ``(registry, repository, tag)`` triple
+    from the canonical image ref. The operator reconstructs the CR's
+    image field from those three, so a repository change between
+    deploys — common when an extension's declared image namespace
+    differs from the legacy ``{ext}-{svc}`` form that pre-fix kz-ext
+    would have written — flows through. Sending only ``tag`` would
+    leave the CR's image field at the original repository and produce
+    ``ImagePullBackOff`` on the next pull.
     """
+    from kamiwaza_extensions.compose_transformer import _split_image_ref
     from kamiwaza_sdk.schemas.extensions import ImagePatch, PatchServiceSpec
 
     patch_services: List[Any] = []
     for svc in payload.services:
-        # Split tag after last '/' to avoid confusing registry port with tag
-        image = svc.image
-        slash_pos = image.rfind("/")
-        after_slash = image[slash_pos + 1:] if slash_pos >= 0 else image
-        if ":" in after_slash:
-            tag = after_slash.rsplit(":", 1)[1]
-        else:
-            tag = "latest"
+        registry, repository, tag = _split_image_ref(svc.image)
+        image_patch = ImagePatch(
+            tag=tag,
+            registry=registry,
+            repository=repository,
+        )
         spec = PatchServiceSpec(
             name=svc.name,
-            image=ImagePatch(tag=tag),
+            image=image_patch,
         )
         if svc.env:
             spec.env = svc.env
--- a/kamiwaza_extensions/commands/dev.py
+++ b/kamiwaza_extensions/commands/dev.py
@@ -303,7 +303,10 @@ def run_dev_remote(
     from kamiwaza_sdk import KamiwazaClient
     from kamiwaza_sdk.exceptions import APIError
 
-    from kamiwaza_extensions.compose_transformer import ComposeTransformer
+    from kamiwaza_extensions.compose_transformer import (
+        ComposeTransformer,
+        _canonical_build_ref,
+    )
     from kamiwaza_extensions.connections import ConnectionManager
     from kamiwaza_extensions.deployment_poller import (
         DeploymentFailedError,
@@ -433,6 +436,23 @@ def run_dev_remote(
     )
     transformed = transformer.resolve_env_placeholders(transformed)
 
+    # Canonical image refs for every build-context service. Single source
+    # of truth shared with the transformed compose so the image we build
+    # and push matches the ref the K8s payload will pull (ENG-4909). The
+    # transformer honors a service's declared image namespace; without
+    # this map ImageBuilder would synthesize the legacy {ext}-{svc} form
+    # and ship a pod referencing an image that was never pushed.
+    canonical_refs: Dict[str, str] = {
+        name: _canonical_build_ref(
+            svc, name,
+            fallback_registry=registry,
+            fallback_extension_name=info.name,
+            revision_tag=rev_tag,
+        )
+        for name, svc in (info.compose_data.get("services") or {}).items()
+        if "build" in svc
+    }
+
     # 5b. Resolve SDK override for build
     build_overrides = None
     if sdk_repo and not no_build:
@@ -498,6 +518,7 @@ def run_dev_remote(
                 service_filter=service,
                 verbose=verbose,
                 build_overrides=build_overrides,
+                image_refs=canonical_refs,
             )
         except ImageBuildError as exc:
             console.print(f"\n[red]Error:[/red] {exc}")
@@ -508,13 +529,15 @@ def run_dev_remote(
         console.print()
     else:
         console.print("[dim]Skipping build (--no-build)[/dim]")
-        # Collect expected image refs for push
-        image_refs = []
-        for svc_name, svc in (info.compose_data.get("services") or {}).items():
-            if service and svc_name != service:
-                continue
-            if "build" in svc:
-                image_refs.append(f"{registry}/{info.name}-{svc_name}:{rev_tag}")
+        # Collect expected image refs for push from the canonical map so
+        # --no-build pushes hit the same registry path the deployment
+        # payload will reference.
+        if service:
+            image_refs = (
+                [canonical_refs[service]] if service in canonical_refs else []
+            )
+        else:
+            image_refs = list(canonical_refs.values())
         console.print()
 
     # 7. Push images
--- a/kamiwaza_extensions/commands/publish.py
+++ b/kamiwaza_extensions/commands/publish.py
@@ -167,32 +167,6 @@ def _collect_buildable_image_names(
     return names
 
 
-def _collect_image_refs(
-    compose_data: Dict[str, Any],
-    extension_name: str,
-    version: str,
-    registry: str,
-) -> List[str]:
-    """Return canonical image refs for services with build contexts.
-
-    Honors each service's declared ``image:`` field (namespace preserved,
-    tag rewritten to *version*) so the refs returned here match what
-    ``_retag_appgarden_compose`` writes to the catalog. Falls back to
-    the legacy ``{registry}/{extension_name}-{svc_name}:{version}`` form
-    only when no image is declared.
-    """
-    refs: List[str] = []
-    for svc_name, svc in (compose_data.get("services") or {}).items():
-        if "build" in svc:
-            refs.append(_canonical_build_ref(
-                svc, svc_name,
-                fallback_registry=registry,
-                fallback_extension_name=extension_name,
-                revision_tag=version,
-            ))
-    return refs
-
-
 def _verify_supplied_digest(ref: str, supplied: str) -> None:
     """Resolve *ref* in the registry and abort if it disagrees with *supplied*.
 
@@ -553,9 +527,24 @@ def run_publish(
     # (legacy {ext}-{svc} vs the namespace declared in compose), and
     # the catalog ships referencing an image that may not exist at the
     # synthesized path. (ENG-4909.)
+    #
+    # Per service, prefer the appgarden compose's declaration when
+    # present — that's what ``_retag_appgarden_compose`` writes into the
+    # catalog. The source ``docker-compose.yml`` declaration is the
+    # fallback for services the appgarden file omits, and the legacy
+    # ``{ext}-{svc}`` form remains the final fallback (handled by
+    # ``_canonical_build_ref``). Reading from source when an appgarden
+    # entry overrides the namespace would build/push/digest at a
+    # different path than the catalog references.
+    appgarden_services = (
+        (appgarden_data or {}).get("services") or {}
+        if isinstance(appgarden_data, dict)
+        else {}
+    )
     canonical_refs: Dict[str, str] = {
         name: _canonical_build_ref(
-            svc, name,
+            appgarden_services.get(name) or svc,
+            name,
             fallback_registry=registry,
             fallback_extension_name=info.name,
             revision_tag=image_tag,
@@ -592,9 +581,7 @@ def run_publish(
             console.print("    [yellow]![/yellow] No images to build")
     else:
         console.print("  [dim]Skipping build (--no-build)[/dim]")
-        image_refs = _collect_image_refs(
-            info.compose_data, info.name, image_tag, registry
-        )
+        image_refs = list(canonical_refs.values())
 
     # 5.5 Preflight: digest resolution post-push needs `docker buildx
     # imagetools`. If buildx is missing on this host, fail before mutating
--- a/kamiwaza_extensions/commands/publish.py
+++ b/kamiwaza_extensions/commands/publish.py
@@ -231,15 +231,18 @@ def _verify_supplied_digest(ref: str, supplied: str) -> None:
 
 
 def _auto_resolve_digests(
-    refs: List[str], *, no_build: bool, no_push: bool,
+    refs: List[str],
 ) -> Dict[str, str]:
     """Resolve registry digests for each *ref* and return ``{ref: digest}``.
 
-    Soft-fails for the catalog-only-republish shape (``--no-build
-    --no-push``): pre-PR behavior was tag-only with no docker dependency,
-    so a resolve failure becomes a warning and the ref is omitted from
-    the result rather than aborting the publish.
+    Aborts the publish on any resolution failure. Silent degradation to
+    tag-only entries (the previous behavior in ``--no-build --no-push``
+    mode) hid upstream image-name mismatches that produced unpullable
+    catalog refs (ENG-4909), and a transient registry hiccup was
+    indistinguishable from a legitimate "docker not on host" fall-through.
     """
+    from rich.markup import escape as escape_markup
+
     from kamiwaza_extensions.image_pusher import ImagePushError, ImagePusher
 
     digest_map: Dict[str, str] = {}
@@ -247,14 +250,24 @@ def _auto_resolve_digests(
         try:
             digest_map[ref] = ImagePusher.resolve_digest(ref)
         except ImagePushError as exc:
-            if no_build and no_push:
-                console.print(
-                    f"  [yellow]warn[/yellow] could not resolve digest "
-                    f"for {ref}: {exc} — catalog will use tag-only for "
-                    "this ref"
-                )
-                continue
-            console.print(f"\n[red]Error:[/red] {exc}")
+            # Markup-escape the dynamic parts so a registry error message
+            # containing literal ``[`` (rare but possible) can't garble the
+            # error output via rich's tag parser.
+            safe_ref = escape_markup(ref)
+            safe_exc = escape_markup(str(exc))
+            console.print(
+                f"\n[red]Error:[/red] could not resolve digest for "
+                f"[bold]{safe_ref}[/bold]: {safe_exc}\n\n"
+                "The image must exist in the registry before catalog publish.\n"
+                "Common causes:\n"
+                "  • Image name in docker-compose.appgarden.yml doesn't "
+                "match the actual GHCR path\n"
+                "  • Image hasn't been pushed yet (run with build+push, "
+                "or push manually first)\n"
+                "  • Transient registry outage (retry the publish)\n"
+                "  • To pin a known digest without registry lookup, pass "
+                "--digest sha256:..."
+            )
             raise typer.Exit(code=1) from exc
     return digest_map
 
@@ -625,9 +638,7 @@ def run_publish(
             if not no_push:
                 _verify_supplied_digest(ref, digest)
         else:
-            digest_map = _auto_resolve_digests(
-                published_refs, no_build=no_build, no_push=no_push,
-            )
+            digest_map = _auto_resolve_digests(published_refs)
 
     # 7. Build catalog entry
     reg_builder = RegistryBuilder()
--- a/kamiwaza_extensions/commands/publish.py
+++ b/kamiwaza_extensions/commands/publish.py
@@ -11,7 +11,10 @@ import yaml
 from rich.console import Console
 
 from kamiwaza_extensions.catalog_publisher import DEFAULT_CATALOG_SCHEMA
-from kamiwaza_extensions.compose_transformer import _canonical_build_ref
+from kamiwaza_extensions.compose_transformer import (
+    _canonical_build_ref,
+    compute_canonical_refs,
+)
 from kamiwaza_extensions.extension_detector import infer_extension_type
 
 console = Console(stderr=True)
@@ -456,6 +459,25 @@ def run_publish(
             registry=registry,
         )
 
+    # Canonical image refs for every build-context service. Single
+    # source of truth shared with the build, push, digest resolution,
+    # and catalog-write paths so they can't drift. Derived once here
+    # before the dry-run branch so the preview reflects exactly what
+    # the live path would produce — appgarden namespace precedence and
+    # profile-gating included.
+    appgarden_services = (
+        (appgarden_data or {}).get("services") or {}
+        if isinstance(appgarden_data, dict)
+        else {}
+    )
+    canonical_refs: Dict[str, str] = compute_canonical_refs(
+        info.compose_data.get("services") or {},
+        registry=registry,
+        extension_name=info.name,
+        revision_tag=image_tag,
+        appgarden_services=appgarden_services,
+    )
+
     # -- Dry-run path (still runs merge check to detect conflicts) --
     if dry_run:
         short_names = _collect_buildable_image_names(
@@ -473,14 +495,7 @@ def run_publish(
         # appears first in dict order can't redirect the user's digest.
         dry_digest_map: Dict[str, str] = {}
         if digest is not None and buildable_services:
-            dry_canonical_ref = _canonical_build_ref(
-                (info.compose_data.get("services") or {}).get(buildable_services[0]),
-                buildable_services[0],
-                fallback_registry=registry,
-                fallback_extension_name=info.name,
-                revision_tag=image_tag,
-            )
-            dry_digest_map[dry_canonical_ref] = digest
+            dry_digest_map[canonical_refs[buildable_services[0]]] = digest
 
         # Run merge check so dry-run detects version conflicts
         reg_builder = RegistryBuilder()
@@ -519,39 +534,6 @@ def run_publish(
         console.print()
         console.print("[dim]No changes made (dry-run mode).[/dim]")
         return
-
-    # Canonical image refs for every build-context service. Single
-    # source of truth — seeds the build, the push, digest resolution,
-    # and the catalog entry's image refs. Without sharing this map
-    # across all four sites they can each synthesize divergent forms
-    # (legacy {ext}-{svc} vs the namespace declared in compose), and
-    # the catalog ships referencing an image that may not exist at the
-    # synthesized path. (ENG-4909.)
-    #
-    # Per service, prefer the appgarden compose's declaration when
-    # present — that's what ``_retag_appgarden_compose`` writes into the
-    # catalog. The source ``docker-compose.yml`` declaration is the
-    # fallback for services the appgarden file omits, and the legacy
-    # ``{ext}-{svc}`` form remains the final fallback (handled by
-    # ``_canonical_build_ref``). Reading from source when an appgarden
-    # entry overrides the namespace would build/push/digest at a
-    # different path than the catalog references.
-    appgarden_services = (
-        (appgarden_data or {}).get("services") or {}
-        if isinstance(appgarden_data, dict)
-        else {}
-    )
-    canonical_refs: Dict[str, str] = {
-        name: _canonical_build_ref(
-            appgarden_services.get(name) or svc,
-            name,
-            fallback_registry=registry,
-            fallback_extension_name=info.name,
-            revision_tag=image_tag,
-        )
-        for name, svc in (info.compose_data.get("services") or {}).items()
-        if "build" in svc
-    }
 
     # 5. Build images (using stage-aware tag)
     image_refs: List[str] = []
--- a/kamiwaza_extensions/commands/publish.py
+++ b/kamiwaza_extensions/commands/publish.py
@@ -56,6 +56,23 @@ def _load_appgarden_compose(
     return candidate, data
 
 
+def _replace_image_tag(image_ref: str, new_tag: str) -> str:
+    """Return *image_ref* with its tag (and any digest) replaced by *new_tag*.
+
+    The namespace (registry + repo path) is preserved verbatim. Handles
+    refs that include a registry port (``localhost:5000/foo:tag``) by
+    using the position of the last ``/`` to disambiguate the port colon
+    from the tag colon, and strips any ``@sha256:...`` suffix before
+    re-tagging.
+    """
+    ref = image_ref.split("@", 1)[0]
+    last_slash = ref.rfind("/")
+    last_colon = ref.rfind(":")
+    if last_colon > last_slash:
+        ref = ref[:last_colon]
+    return f"{ref}:{new_tag}"
+
+
 def _retag_appgarden_compose(
     appgarden_data: Dict[str, Any],
     source_compose_data: Optional[Dict[str, Any]],
@@ -106,7 +123,22 @@ def _retag_appgarden_compose(
         if not isinstance(svc, dict):
             continue
         if svc_name in build_services:
-            svc["image"] = f"{registry}/{extension_name}-{svc_name}:{image_tag}"
+            existing = svc.get("image")
+            if isinstance(existing, str) and existing.strip():
+                # The appgarden compose's image field is the canonical
+                # declaration of where this build's image lives in the
+                # registry — set by the extension's sync-compose.py from
+                # its docker-compose.yml. We only own the *tag* (stage
+                # suffix or --revision SHA); the namespace stays what
+                # the extension authored. Without this, extensions whose
+                # GHCR path doesn't match the {ext}-{svc} convention
+                # silently get the wrong namespace in the catalog.
+                svc["image"] = _replace_image_tag(existing, image_tag)
+            else:
+                # No declared image — fall back to the {ext}-{svc}
+                # convention so extensions relying on auto-generated
+                # image fields keep working.
+                svc["image"] = f"{registry}/{extension_name}-{svc_name}:{image_tag}"
     return out
 
 
--- a/kamiwaza_extensions/commands/publish.py
+++ b/kamiwaza_extensions/commands/publish.py
@@ -11,7 +11,7 @@ import yaml
 from rich.console import Console
 
 from kamiwaza_extensions.catalog_publisher import DEFAULT_CATALOG_SCHEMA
-from kamiwaza_extensions.compose_transformer import _replace_image_tag
+from kamiwaza_extensions.compose_transformer import _canonical_build_ref
 from kamiwaza_extensions.extension_detector import infer_extension_type
 
 console = Console(stderr=True)
@@ -107,22 +107,18 @@ def _retag_appgarden_compose(
         if not isinstance(svc, dict):
             continue
         if svc_name in build_services:
-            existing = svc.get("image")
-            if isinstance(existing, str) and existing.strip():
-                # The appgarden compose's image field is the canonical
-                # declaration of where this build's image lives in the
-                # registry — set by the extension's sync-compose.py from
-                # its docker-compose.yml. We only own the *tag* (stage
-                # suffix or --revision SHA); the namespace stays what
-                # the extension authored. Without this, extensions whose
-                # GHCR path doesn't match the {ext}-{svc} convention
-                # silently get the wrong namespace in the catalog.
-                svc["image"] = _replace_image_tag(existing, image_tag)
-            else:
-                # No declared image — fall back to the {ext}-{svc}
-                # convention so extensions relying on auto-generated
-                # image fields keep working.
-                svc["image"] = f"{registry}/{extension_name}-{svc_name}:{image_tag}"
+            # The appgarden compose's image field is the canonical
+            # declaration of where this build's image lives in the
+            # registry — set by the extension's sync-compose.py from
+            # its docker-compose.yml. We only own the *tag* (stage
+            # suffix or --revision SHA); the namespace stays what the
+            # extension authored.
+            svc["image"] = _canonical_build_ref(
+                svc, svc_name,
+                fallback_registry=registry,
+                fallback_extension_name=extension_name,
+                revision_tag=image_tag,
+            )
     return out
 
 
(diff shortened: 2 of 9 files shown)


jxstanford (changes requested):
# Review Report: fix(extensions): kz-ext publish honors compose image namespace + fails loudly on digest miss (ENG-4909)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/101
**Branch**: `feature/eng-4909-kz-ext-honor-compose-image` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)

## High Priority Issues
| 1 | `kamiwaza_extensions/commands/publish.py:629-631` (also affects `image_builder.py:65`) | The fix only goes half-way: `_retag_appgarden_compose` now preserves the declared namespace in the catalog compose, but `published_refs` (and `ImageBuilder.build`) still synthesize the legacy `{registry}/{ext}-{svc}:{tag}` convention. For the exact non-conventional case this PR targets (omniparse: `images/omniparse` declared, `images/tool-omniparse-omniparse-server` legacy), the consequences are: **(a) build+push path** — kz-ext builds/pushes to the legacy ref, `_auto_resolve_digests` keys `digest_map` with legacy refs, but `transformed_compose["services"][*]["image"]` carries the declared ref, so `_apply_digests` (registry_builder.py:421) fails exact-string match and the catalog ships with the declared image **unpinned** (and possibly never pushed by this run); **(b) `--digest` path** — same mismatch; **(c) `--no-build --no-push` path** — the new hard-fail catches it loudly, which is the only path this PR fully fixes. Derive `published_refs` (and `ImageBuilder.build`'s `image_ref`) from the same declared-namespace image fields that `_retag_appgarden_compose` writes. | Codex (P1), Claude (Medium) — higher severity wins | Logic — partial fix |

| 2 | `kamiwaza_extensions/commands/publish.py:588-589` | Stale comment: "the catalog-only-republish path soft-falls in `_auto_resolve_digests` and doesn't need this guard." This PR removed the soft-fall; the guard-skipping is still correct (no push → no buildx required) but the cited reason no longer holds. | Claude | Doc rot |
| 3 | `tests/unit/extensions/test_publish_cmd.py:799-801` | Same stale comment in `test_buildx_preflight_skipped_when_no_push`. | Claude | Doc rot |
| 4 | `kamiwaza_extensions/commands/publish.py:263` | Error message says "GHCR path" but the publish flow supports arbitrary registries (e.g., `localhost:5000`). Generic wording ("registry image path") would age better. | Claude | Polish |
| 5 | `kamiwaza_extensions/commands/publish.py:258-270` | Optional: add a smoke test pinning the new error-message contract (failed ref name + at least one "common cause" line). The actionable error string is the value of the change but nothing currently asserts it. | Claude | Test gap |

## Test Coverage Assessment
- `_replace_image_tag`: well covered (simple, namespace-divergent, registry-port, digest-strip, no-existing-tag).
- `_retag_appgarden_compose`: namespace-preservation + legacy fallback + external/mixed/profiled-build/sync-invented cases all present.
- `test_no_build_no_push_resolve_failure_aborts_by_default`: correctly inverts the prior (buggy) soft-fall assertion.
- **Gap**: No test covers the build+push or `--digest` paths for non-conventional namespaces — i.e., the precise scenario Issue #1 identifies. A test wiring a non-conventional appgarden image through the full `run_publish` path with auto-resolve would have caught the digest_map mismatch.

## Positive Observations
- `_replace_image_tag` correctly handles registry ports (`last_colon > last_slash` disambiguation), strips `@sha256:...` digests, and degrades when no tag exists — all with dedicated tests.
- `rich.markup.escape` neutralises markup-injection from upstream registry error strings.
- Falling back to `{ext}-{svc}` when no `image:` is declared preserves backward compatibility.
- The replaced test (`..._aborts_by_default`) properly inverts an assertion that had been codifying buggy semantics from PR #100.
- The new `_auto_resolve_digests` error message names the failed ref and lists actionable causes including the `--digest` escape hatch.
- `buildable_services` profile-filter is correctly mirrored inside `_retag_appgarden_compose`, preventing a profiled build service from being retagged into the catalog without a corresponding push.

## Overall Verdict
**Risk Level**: Medium
**Verdict**: REQUEST_CHANGES
**Engine Agreement**: 2/2 engines contributed (Codex P1, Claude Medium — higher severity applied per dedup rules)

**Rationale**: The PR title claims `kz-ext publish honors compose image namespace`, but only the catalog compose honors it — the build/push pipeline and `published_refs` derivation still synthesize the legacy `{ext}-{svc}` convention. For omniparse-style extensions, this means the build+push and `--digest` paths produce a catalog entry with the declared namespace image **unpinned** and possibly never pushed by this publish. The hard-fail in `_auto_resolve_digests` does catch the `--no-build --no-push` symptom loudly (a real improvement), but the other two paths regress to a different shade of silent-broken. Issue #1 should be fixed in this PR — derive `published_refs` and `ImageBuilder.build`'s `image_ref` from the same declared-namespace fields `_retag_appgarden_compose` rewrites. Stale comments (#2, #3) are non-blocking cleanups.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |
| Gemini | ⬜ Not available |

**Engines contributing**: 2/3
</details>

JohnStrick:
@jxstanford Thanks for the careful review. Walking through each finding:

## H1 — partial fix (publish.py:629-631 + image_builder.py:65) → Fixed

You were right that the prior commits only fixed the catalog-write side. `published_refs`, `ImageBuilder.build`, and `_collect_image_refs` were still synthesizing the legacy `{ext}-{svc}` form, so for omniparse-style extensions the build+push and `--digest` paths would ship a catalog entry with the declared namespace **unpinned** (digest_map keyed by legacy form, `_apply_digests` exact-string match fails against the compose's declared namespace).

Three new commits land the full fix:

- **`refactor(extensions): move _replace_image_tag to compose_transformer module`** — pure refactor that puts the helper in a library module so `ComposeTransformer` and `ImageBuilder` can use it without creating a `lib → commands` import dependency.
- **`fix(extensions): derive build refs and published_refs from canonical helper`** — the H1 fix. New `_canonical_build_ref` helper in `compose_transformer.py`. `run_publish` now computes a `canonical_refs` map once (per build-context service → canonical ref, namespace from declared `image:` with tag rewritten) and routes it through every site:
  - `ImageBuilder.build` (new `image_refs` kwarg; `dev.py` caller keeps backward-compat via the optional param defaulting to None)
  - `_collect_image_refs` (`--no-build` push path)
  - `published_refs` (digest resolution + summary print)
  - `dry_canonical_ref` (`--dry-run --digest`)
  - `ComposeTransformer.transform_service` (the fallback path when no `docker-compose.appgarden.yml` exists — would've been the next foot-gun once an extension without appgarden compose hit auto-publish)
- New test `test_non_conventional_namespace_flows_through_build_and_resolve` wires an omniparse-style declared namespace through the full `run_publish` path with build+push mocked. Asserts:
  1. `ImageBuilder.build` receives the declared namespace via the `image_refs` kwarg (not the synthesized form)
  2. `resolve_digest` queries the declared namespace
  3. `digest_map` keys match the declared namespace → `_apply_digests` pins the catalog entry's compose `image:` correctly

That closes the test-coverage gap you identified ("no test covers the build+push or `--digest` paths for non-conventional namespaces"). Also added a fallback-path divergence test in `TestBuildContextRemoval` for the `ComposeTransformer` change, replacing the existing `test_updates_existing_image_tag` whose assertion codified the same buggy synthesis you flagged.

## M2 (`publish.py:603` stale soft-fall comment) → Fixed

Updated in the `chore` commit. Now reads "the catalog-only-republish path (`--no-build --no-push`) doesn't push, so the buildx preflight is moot there" — guard-skipping is still correct and the cited reason now matches reality.

## M3 (`test_publish_cmd.py:800` same stale comment) → Fixed

Same wording update in the `chore` commit.

## M4 (`publish.py:263` "GHCR path" wording) → Fixed

Changed to "registry image path" in the `chore` commit. Generic wording matches the publish flow's support for arbitrary registries (`localhost:5000`, vendor mirrors).

## M5 (test for error-message contract) → Deferred

Not added in this PR. Asserting exact error-message strings locks in wording that's likely to evolve, while the existing `test_no_build_no_push_resolve_failure_aborts_by_default` already pins the meaningful contract (non-zero exit on resolve failure). Happy to add a fragment-only assertion (e.g., that the failed ref appears in the error) as a follow-up if you think the value warrants it — but per your own Optional classification I left it for now.

Ready for re-review. Five commits total, +234 LOC / -71 across publish, image_builder, compose_transformer, and their tests.

jxstanford (changes requested):
# RE-REVIEW Report: fix(extensions): kz-ext publish honors compose image namespace + fails loudly on digest miss (ENG-4909)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/101
**Branch**: `feature/eng-4909-kz-ext-honor-compose-image` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)

## High Priority Issues
| 1 | kamiwaza_extensions/compose_transformer.py:177-183 (interacts with kamiwaza_extensions/commands/dev.py:427,491) | `ComposeTransformer.transform_service` now emits the declared-namespace canonical ref, and `dev.py` feeds the resulting `transformed` dict to `PayloadBuilder` — but `dev.py`'s call to `ImageBuilder.build()` at line 491 does NOT pass `image_refs=`, so it still tags/pushes the legacy `{registry}/{ext}-{svc}:{rev}` form. For any dev compose service with `build:` plus a declared `image:` whose namespace diverges from the legacy convention, the K8s payload tells the cluster to pull the canonical ref while only the legacy ref was pushed → ImagePullBackOff (or, more insidiously, a successful pull of a stale image from a prior run). The publish path threads `canonical_refs` through `ImageBuilder.build()`; the dev path needs the same wiring or the shared transformer needs a publish-only escape hatch. | Codex | Correctness / regression |
| 2 | kamiwaza_extensions/commands/publish.py:556-564 (vs. line 469's `_retag_appgarden_compose`) | When `docker-compose.appgarden.yml` is present, the catalog's `transformed` is built from `appgarden_data` (whose `image:` fields are the canonical record per `_retag_appgarden_compose`'s docstring at line 110), but `canonical_refs` — which seeds build, push, and `digest_map` keys — is built from `info.compose_data` (the source `docker-compose.yml`). If sync-compose produces an appgarden file whose service image namespace differs from the source compose's (or source has no `image:` while appgarden does — the very omniparse-style case this PR targets), publish pushes one ref and the catalog ships pointing at a different one, AND `_apply_digests` (registry_builder.py:421) matches by exact string so the digest_map entry silently no-ops → catalog gets an unpinned, possibly-unpullable ref. `canonical_refs` should be derived from `appgarden_data` when present so all four sites agree. | Codex | Correctness |

| 1 | kamiwaza_extensions/commands/publish.py:595-597 | In the `no_build` branch, `_collect_image_refs(info.compose_data, info.name, image_tag, registry)` recomputes what `canonical_refs` already holds via the same `_canonical_build_ref` call. Replace with `image_refs = list(canonical_refs.values())` to remove the duplicate traversal and a second site that could drift if canonical-ref derivation changes. | Claude Code | Code quality / duplication |
| 2 | kamiwaza_extensions/compose_transformer.py:50 | `isinstance(declared, str) and declared.strip()` checks non-empty but passes the un-stripped value to `_replace_image_tag`. A quoted YAML value like `"  ghcr.io/foo  "` would propagate surrounding whitespace into the rewritten ref. Strip at the boundary. | Claude Code | Defensive validation |
| 3 | tests/unit/extensions/test_publish_cmd.py:2027 vs test_compose_transformer.py:113 | Two distinct tests both named `test_preserves_declared_namespace_when_diverges_from_convention` live in different classes — not a collision, but `pytest -k` matches both. Disambiguate (e.g. `..._in_retag` / `..._in_transform_service`). | Claude Code | Test naming |

## Test Coverage Assessment
- Strong direct coverage for the helpers: `_replace_image_tag` (port/digest/no-tag edge cases) and `_canonical_build_ref` (via `transform_service` and `_retag_appgarden_compose` paths).
- `test_non_conventional_namespace_flows_through_build_and_resolve` is well-structured and asserts the omniparse-style ref propagates through `image_refs=`, `resolve_digest()`, and `digest_map=` — exactly the regression shape this PR prevents.
- `test_no_build_no_push_resolve_failure_aborts_by_default` correctly locks in the fail-loud contract.
- **Gaps tied to the High-priority findings above**:
  - No test exercises `kz-ext dev` with a service whose declared image diverges from `{ext}-{svc}` — the dev-path inconsistency in H1 would have been caught by an end-to-end assertion that ImageBuilder is tagged with the same ref the PayloadBuilder sends to the cluster.
  - No test covers the appgarden-vs-source image divergence in H2 — `TestPublishWithAppgarden.test_appgarden_file_bypasses_generic_transform` uses an appgarden image that happens to match the legacy convention, so the bug is invisible.
- Minor: no direct unit test for `ImageBuilder.build`'s new `image_refs` kwarg (asserting fallback when a service is absent from the map). The e2e coverage is probably sufficient.

## Positive Observations
- Single-source-of-truth refactor is the right shape — five divergent call sites collapse into `_canonical_build_ref` consumed everywhere.
- `_replace_image_tag` correctly disambiguates the registry-port colon from the tag colon via `rfind("/")` and strips `@sha256:...` suffixes; edge cases are well-covered.
- The behavioral change from soft-fall to loud abort in `_auto_resolve_digests` is intentional, well-justified in the docstring with ENG-4909 rationale, and the prior test was updated to assert the new contract.
- Actionable error message with `escape_markup` defense against rich's tag parser.
- Stale comments referencing the old soft-fall behavior were updated in both publish.py and the test file.
- Removal of unused `import sys` in `image_builder.py` is welcome hygiene.

## Overall Verdict
**Risk Level**: High
**Verdict**: REQUEST_CHANGES
**Engine Agreement**: 2/3 engines contributed. Claude Code rated APPROVE based on the publish path being internally consistent; Codex's cross-file analysis surfaced two correctness issues — the dev path was left out of the canonical-ref plumbing (H1), and the appgarden vs. source-compose divergence in publish (H2) defeats digest pinning silently for exactly the non-conventional namespace case ENG-4909 is meant to fix. Both warrant changes before merge; H2 in particular undermines the PR's stated goal.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |
| Gemini | ⬜ Not available |

**Engines contributing**: 2/3
</details>

jxstanford (changes requested):
# RE-REVIEW Report: fix(extensions): kz-ext publish honors compose image namespace + fails loudly on digest miss (ENG-4909)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/101
**Branch**: `feature/eng-4909-kz-ext-honor-compose-image` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| _(none)_ | | | | |

## High Priority Issues
| 1 | `kamiwaza_extensions/compose_transformer.py:49-53` | `_canonical_build_ref` returns the declared `image:` verbatim (only retagging) for ANY string, so a buildable service with a non-registry-qualified image like `image: api:latest` or `image: my-org/api:latest` now bypasses the derived/profile registry. Pre-PR `transform_service` would have rewritten these to `{registry}/{ext}-{svc}:{tag}`; post-PR both `kz-ext dev` and `kz-ext publish` will build and push at the bare ref (which `docker push` resolves to Docker Hub), while the K8s payload / catalog reference the same bare ref the cluster registry doesn't carry — net result: ImagePullBackOff or unpullable catalog entries for compose shapes that worked before. Recommended fix: only honor `declared` when it is already qualified with a registry host (contains a `.`/`:` before the first `/`), otherwise fall back to the legacy `{fallback_registry}/{fallback_extension_name}-{svc_name}:{revision_tag}` form. | Codex | Logic / Regression |

| 1 | `kamiwaza_extensions/commands/publish.py:544-554` | The `appgarden_services.get(name) or svc` derivation diverges from `_retag_appgarden_compose` when an appgarden entry is present but its mapping is empty/missing `image:` (Python truthiness: `{} or svc` → `svc`). In that case `canonical_refs` reads from the source compose, but `_retag_appgarden_compose` calls `_canonical_build_ref(svc=appgarden_entry, ...)` and writes the legacy fallback ref into the catalog. Build/push/digest happen at the source ref while the catalog references a different ref — recreating the very mismatch this PR fixes. Use `appgarden_services[name] if name in appgarden_services else svc` (presence-based, not truthiness-based) so the two sites agree. | Codex | Logic / Consistency |
| 2 | `kamiwaza_extensions/commands/publish.py:476-483` | Dry-run `dry_canonical_ref` reads from source `info.compose_data` only, but the live path (lines 539-554) prefers `appgarden_services.get(name) or svc`. When an appgarden compose declares a different namespace than the source compose **and** the user passes `--digest`, the dry-run keys `dry_digest_map` against the source-compose namespace while `transformed` (built via `_retag_appgarden_compose`) carries the appgarden namespace; `_apply_digests` does exact-string match and silently leaves the entry unpinned. No production impact (dry-run doesn't write a catalog) but masks a successful-looking dry-run for a config that would pin differently live. Mirror the live derivation. | Claude | Logic / Consistency |
| 3 | `kamiwaza_extensions/commands/dev.py:444-453` and `publish.py:544-554` | `canonical_refs` includes services that have `"build" in svc` regardless of `profiles:` — under `--no-build` paths `image_refs = list(canonical_refs.values())` then becomes the push list, which can include profile-gated images. Pre-existing behavior (old `_collect_image_refs` had the same gap), but the comment block calls this "single source of truth" — worth filtering by `not svc.get("profiles")` for consistency with `buildable_services`. | Claude | Consistency |

## Test Coverage Assessment
- **Adequate coverage**: `_replace_image_tag` (simple tag, divergent namespace, registry-with-port, digest-stripping, tag-less); `transform_service` namespace-preservation and divergent-namespace; `_retag_appgarden_compose` divergent namespace + legacy fallback; `_auto_resolve_digests` abort behavior; end-to-end `test_appgarden_image_namespace_drives_build_and_push` and `test_dev_canonical_refs.py`.
- **Gaps**:
  - **High #1**: no test asserts behavior for a buildable service with an unqualified image like `image: api:latest` or `image: my-org/api:latest` — the regression is invisible to the suite.
  - **Medium #1**: no test for an appgarden service entry that exists but is an empty mapping (or missing `image:`) — the truthiness-vs-presence divergence isn't covered.
  - **Medium #2**: dry-run + `--digest` + appgarden-namespace-divergence path is not asserted.
  - `_auto_resolve_digests` markup-escape behavior is implemented but not asserted.

## Positive Observations
- Centralizing canonical-ref derivation in `_canonical_build_ref` removes five independent sites of synthesis; the helper handles realistic edge cases (port-in-registry, existing digest, missing tag, empty `image:`) and is well-tested for the registry-qualified case.
- Deliberate trade-off in `_auto_resolve_digests` (drop soft-fall, abort with actionable error) is well-justified and matches the goal of surfacing the upstream mismatch the soft-fall was hiding. Error message clearly lists causes and the `--digest` escape hatch.
- Markup-escape on dynamic error content (`escape_markup(ref)` / `escape_markup(str(exc))`) is a thoughtful defensive touch.
- Test fixtures use realistic divergent-namespace shapes (omniparse: `ghcr.io/.../tool-omniparse/omniparse`), so they would have caught the original bug.
- The `image_refs` kwarg on `ImageBuilder.build` is additive and backward-compatible.
- Comment density is high; `_canonical_build_ref` docstring explicitly names the call sites it unifies.

## Overall Verdict
**Risk Level**: Medium
**Verdict**: REQUEST_CHANGES

The fix is conceptually right and well-tested for the omniparse-style case it targets, but the `_canonical_build_ref` rule "if `image:` declared, retag verbatim" is too permissive: it regresses compose shapes where `image:` is present but not registry-qualified (a common dev convention) — those previously got rewritten to the cluster registry and now silently route to Docker Hub. The appgarden empty-mapping case is a smaller-but-real divergence between the canonical-refs derivation and `_retag_appgarden_compose`. Both deserve narrowing the helper's "honor declared" predicate and switching to presence-based fallback before merge; tests for the unqualified-image and empty-appgarden-entry shapes should land alongside.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |
| Gemini | ⬜ Not available |

**Engines contributing**: 2/3
</details>

JohnStrick:
@jxstanford addressed all 4 findings from the re-review.

**H1 — `_canonical_build_ref` registry-qualified predicate** (commit e67e59f)

Added `_looks_registry_qualified` and gated the "honor declared image" branch on it. Predicate matches docker's standard rule: first slash-separated component must contain `.` or `:`, or be exactly `localhost`. Bare repos (`api:latest`) and Docker Hub short forms (`my-org/api:1.0`) fall back to the legacy `{registry}/{ext}-{svc}:{tag}` form — those would have leaked to Docker Hub under `docker push` while the cluster registry expected the cluster-side rewrite.

Tests: 5 cases in `TestLooksRegistryQualified`, 8 in `TestCanonicalBuildRef`, and `test_unqualified_short_form_falls_back_to_legacy` in `TestBuildContextRemoval` exercising the regression shape through `ComposeTransformer.transform`.

**M1 — appgarden lookup presence-based, not truthiness-based** (commit e905241)

Switched `appgarden_services.get(name) or svc` to `appgarden_services[name] if name in appgarden_services else svc`. Empty appgarden mappings now correctly fall through to `_canonical_build_ref({}, ...)` which produces the legacy fallback — matching what `_retag_appgarden_compose` writes into the catalog. Test: `TestComputeCanonicalRefs::test_empty_appgarden_entry_falls_through_to_legacy`.

**M2 — dry-run + `--digest` + appgarden-divergence now mirrors live derivation** (commit e905241)

Extracted `compute_canonical_refs` to `compose_transformer.py`. Both the live path and the dry-run path call it, derived once at the top of `run_publish` (before the dry-run early-return). The dry-run `dry_digest_map` now keys against `canonical_refs[buildable_services[0]]` so the preview matches what the live path would produce. Test: `TestPublishDigest::test_dry_run_digest_uses_appgarden_namespace`.

**M3 — `canonical_refs` filters profile-gated services** (commit e905241)

The shared `compute_canonical_refs` helper filters `not svc.get("profiles")`, mirroring `buildable_services`. Under `--no-build`, `image_refs = list(canonical_refs.values())` is the push list — without the filter, profile-gated dev helpers leaked to the registry. Test: `TestPublishWithBuild::test_no_build_push_excludes_profile_gated_services`.

**Validation:**
- 124 tests pass across `test_publish_cmd.py` + `test_compose_transformer.py` + `test_dev_canonical_refs.py`
- 1151 pass in full extension suite (1 deselected; unrelated brittle `test_available_port` that fails on hosts where port 59123 is taken)
- Ruff clean on all changed files
- Direct Python invocation against omniparse's real compose files confirms the predicate honors `ghcr.io/...` and the helper produces the same canonical ref for publish (with appgarden) and dev (without)

PR description updated with new commit list and test coverage.

jxstanford (changes requested):
# RE-REVIEW Report: fix(extensions): kz-ext publish honors compose image namespace + fails loudly on digest miss (ENG-4909)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/101
**Branch**: `feature/eng-4909-kz-ext-honor-compose-image` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | Cyclomatic Complexity (pre-existing Grade F on `run_publish`/`run_dev_remote`; unrelated to this PR's logic) |

## Critical Issues (Must Fix)
| 1 | `kamiwaza_extensions/commands/dev.py:516` (interacts with `dev.py:67-106` `_build_patch_service_specs`) | `run_dev_remote` now builds and pushes at the canonical (potentially non-legacy) repository, but the PATCH path still only sends `ImagePatch(tag=tag)` with no `registry`/`repository`. For an existing deployment whose CR was created against the legacy `{registry}/{ext}-{svc}` repository — common during the SDK-upgrade transition or whenever an author changes a declared image namespace — the next `kz-ext dev remote` builds+pushes to the new canonical path, then PATCHes only the tag. The operator keeps the old repository and pulls `old-repo:<new-tag>`, which doesn't exist → ImagePullBackOff. This regression hits the exact divergent-namespace scenario the PR exists to fix. Fix: populate `registry`/`repository` on `ImagePatch` from the rewritten ref (`ImagePatch` already supports both at `kamiwaza_sdk/schemas/extensions.py:139-146`), or force delete+recreate when the canonical repository differs from the CR's current image. | Codex | Correctness / Regression |

## High Priority Issues
_(none)_

| 1 | `kamiwaza_extensions/image_builder.py:74-77` | Fallback branch (`image_refs` provided but missing service entry) has only end-to-end coverage. A 3-line unit test against `ImageBuilder.build` with a partial map would lock the contract. | Claude Code | Test gap |
| 2 | `kamiwaza_extensions/commands/publish.py:244` / `kamiwaza_extensions/commands/dev.py:293` | `run_publish` cyclomatic complexity rose +4 (Grade F → F); pre-existing CI gate. Extracting build/push/digest-resolve stages into helpers would clear the long-standing CC failure. | Claude Code | Maintainability |
| 3 | `kamiwaza_extensions/compose_transformer.py:40-44` | `_looks_registry_qualified` doesn't handle IPv6 host forms (`[::1]:5000/foo`). Not a real shape in any current deployment; a comment noting the limitation would help future readers. | Claude Code | Docs |

## Test Coverage Assessment
- Strong unit coverage of `_looks_registry_qualified`, `_canonical_build_ref`, `_replace_image_tag`, and `compute_canonical_refs` (incl. empty-appgarden presence-based fallback, profile filtering, and missing-service cases).
- End-to-end coverage of the omniparse-style divergent-namespace flow through publish (build → push → resolve → catalog) and `run_dev_remote` (build kwargs).
- **Coverage gap matching Critical #1**: no test exercises the PATCH path with a deployment whose existing CR image points at the legacy repo while the new canonical ref points at a different repo. The dev test only verifies the build-time `image_refs=` kwarg, not the subsequent PATCH payload's repository field.
- Soft-fall behavior on resolve_digest under `--no-build --no-push` is now correctly inverted to abort (test renamed `_falls_back_to_tag_only` → `_aborts_by_default`).

## Positive Observations
- Single-source-of-truth refactor via `compute_canonical_refs` collapses five divergent ref-derivation sites (publish live, publish dry-run, dev, build, push, catalog) into one, eliminating drift between what gets built, pushed, and pinned.
- Presence-based appgarden lookup (`appgarden[name] if name in appgarden else svc`) matches `_retag_appgarden_compose` exactly, including the edge case where `services.foo: {}` falls through to legacy on both sides.
- `_looks_registry_qualified` correctly mirrors docker's namespace rule, blocking the Docker Hub silent-routing regression that a naïve "honor declared verbatim" policy would introduce.
- Profile-gated services are now filtered in canonical refs, matching `buildable_services`; closes the no-build push leakage path.
- Loud-fail on digest-resolve failure with markup-escaped, actionable error message addresses ENG-4909's root cause (silent tag-only fallback hid unpullable refs).
- Test fixtures use realistic divergent-namespace shapes that actually exercise the bug pattern.

## Overall Verdict
**Risk Level**: High
**Verdict**: REQUEST_CHANGES
**Engine Agreement**: 2/3 engines contributed; Claude Code returned APPROVE, Codex flagged the PATCH-only-tag regression. The Codex finding is a real correctness gap that bites the exact divergent-namespace scenario this PR targets: the build/push half is fixed for dev, but PATCH still ships tag-only, so existing CRs on the legacy repo will pull `old-repo:new-tag` and BackOff. Address the PATCH path (include `registry`/`repository` on `ImagePatch` when the canonical ref's repository differs from what the CR holds, or force delete+recreate on repository change) before merge. Once resolved, this is a clean APPROVE — the refactor is well-tested and the publish-side fixes are correct.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |
| Gemini | ⬜ Not available |

**Engines contributing**: 2/3
</details>

JohnStrick:
@jxstanford addressed C1 + M1 + M3 in 997dc3c. Deferring M2.

**C1 — PATCH carries registry+repository+tag** (fixed)

You're right that the PATCH was leaving the CR's repository field unchanged when the canonical build ref's repository differs from what the CR holds. SDK-upgrade from pre-fix kz-ext was the worst case — pre-fix kz-ext always synthesized `{registry}/{ext}-{svc}`, this PR builds at the declared namespace, and a tag-only PATCH would leave the operator pulling the legacy repo with a tag that was never pushed there.

Added `_split_image_ref` to `compose_transformer.py` (mirrors `_replace_image_tag`'s tag detection and `_looks_registry_qualified`'s host rule; strips digest defensively). `_build_patch_service_specs` now populates all three ImagePatch fields from the canonical ref. Tests cover the regression scenario directly: `TestBuildPatchServiceSpecs::test_repo_change_between_deploys_flows_through_patch`, plus 9 unit tests for the splitter (registry+port, localhost, digest stripping, unqualified short form, multi-segment repository).

**M1 — ImageBuilder fallback test** (fixed)

Added `test_image_refs_partial_map_falls_back_to_legacy` in `TestBuild`. Compose with two buildable services, `image_refs` map containing only one — the omitted service falls back to `{registry}/{ext}-{svc}:{tag}`. Locks the contract.

**M2 — `run_publish` cyclomatic complexity Grade F** (deferred)

The F-grade is pre-existing on `run_publish` (and `run_dev_remote`); this PR added +4 to it. The CC gate has been failing on these functions across multiple PRs per the project's known pre-existing CC gate convention. Extracting the build/push/digest-resolve stages into helpers is the right move but it's a scope-shifting refactor that deserves its own ticket and review focus. Happy to file an ENG ticket for it.

**M3 — IPv6 doc note** (fixed)

Added a sentence to `_looks_registry_qualified`'s docstring noting that IPv6 host literals (`[::1]`, `[::1]:5000`) are matched via the `:` rule. My predicate already returns True for these because `[::1]` contains `:`.

**Validation:**
- 1165 tests pass in full extension suite (1 deselected — brittle host-port test unrelated to PR)
- 150 pass across `test_compose_transformer.py` + `test_dev_patch.py` + `test_image_builder.py` + `test_dev_canonical_refs.py` + `test_publish_cmd.py`
- Ruff clean on changed files (also fixed three pre-existing F401s in `test_dev_patch.py` + `test_image_builder.py` that surfaced when my edits triggered re-lint)
- Sibling sweep: `ImagePatch(...)` is constructed in exactly one production site (`dev.py:91`), no other PATCH builders to keep in lockstep

jxstanford (approved):
# RE-REVIEW Report: fix(extensions): kz-ext publish honors compose image namespace + fails loudly on digest miss (ENG-4909)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/101
**Branch**: `feature/eng-4909-kz-ext-honor-compose-image` → `develop`
**Engines**: Claude Code, Codex, Kamiwaza (3/3 available)

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None |

## Critical Issues (Must Fix)
| — | — | None | — | — |

## High Priority Issues
| — | — | None | — | — |

| 1 | `kamiwaza_extensions/image_builder.py:68-77` | Live-build path still leaks profile-gated services into the push list. `ImageBuilder.build` iterates every `build:` service in `compose_data` with no `profiles:` filter; when a profile-gated service is absent from `image_refs`, it falls back to the legacy `{registry}/{ext}-{svc}:{tag}` form and gets appended to the returned list, which `run_publish`/`run_dev_remote` then push. `--no-build` is correctly fixed via `list(canonical_refs.values())`, but the live-build path retains the pre-existing leak. Either short-circuit when `image_refs is not None and svc_name not in image_refs`, or add the same `not svc.get("profiles")` filter alongside the existing `"build" not in svc` skip. | Claude | Architecture / Consistency |
| 2 | `kamiwaza_extensions/compose_transformer.py:154-155` | Design tension flag: `_canonical_build_ref` preserves the compose-declared registry host when the ref is registry-qualified, even if the active publish profile or `KAMIWAZA_REGISTRY` differs. Codex flagged this as a regression vs. "publish always pushes to the configured registry"; in practice this is the intentional ENG-4909 contract (the appgarden compose is authoritative — see `test_appgarden_image_namespace_drives_build_and_push` and `test_transform_preserves_divergent_namespace`). Worth a brief note in the PR description / docs so consumers understand that cross-profile re-targeting now requires authoring the appgarden compose to match. | Codex | Behavior / Documentation |
| 3 | `kamiwaza_extensions/compose_transformer.py:107-108` | `compute_canonical_refs` is missing the `isinstance(svc, dict)` guard that `_retag_appgarden_compose` (publish.py:107) uses. A malformed compose with `services.foo: null` would raise `TypeError` here while the appgarden path silently skips. `ComposeValidator` should catch this upstream, but adding `if not isinstance(svc, dict): continue` keeps the two source-of-truth derivations in lockstep on edge inputs. | Claude | Robustness |
| 4 | `kamiwaza_extensions/commands/publish.py:613-614` | Tight coupling: `published_refs = [canonical_refs[name] for name in buildable_services]` assumes `_collect_buildable_image_names`/`buildable_services` and `compute_canonical_refs` apply identical profile filters. They currently do (both check `"build" in svc and not svc.get("profiles")`), but the invariant is implicit. Consider deriving `published_refs` from `canonical_refs` directly, or asserting key alignment, so a future drift in either filter surfaces immediately rather than as a `KeyError` mid-publish. | Kamiwaza | Maintainability |
| 5 | `kamiwaza_extensions/compose_transformer.py:36-42` | The `_looks_registry_qualified` docstring claims IPv6 host literals (`[::1]/foo`, `[::1]:5000/foo`) are matched via the `:` rule, and they are, but there's no test under `TestLooksRegistryQualified` pinning that behavior. One assertion would prevent a future regression that's only visible at runtime. | Claude | Test Coverage |

## Test Coverage Assessment
**Verdict: Strong.** Coverage tracks the four-site canonical-refs invariant the PR sets out to enforce:
- `compose_transformer.py`: Focused unit tests for `_replace_image_tag`, `_looks_registry_qualified`, `_split_image_ref`, `_canonical_build_ref`, and `compute_canonical_refs` covering registry-qualified vs Docker Hub shortcut, ports, localhost, empty/whitespace, digest stripping, profile-gating, appgarden presence-vs-truthiness, and multi-segment repos.
- `commands/publish.py`: Appgarden-namespace-drives-build+push+digest, profile-gated exclusion under `--no-build`, dry-run digest pinning against appgarden ref, hard-fail on digest resolution, non-conventional namespace propagation.
- `commands/dev.py`: `_build_patch_service_specs` populates registry/repository/tag (incl. localhost-with-port, unqualified short-form); repo-change-between-deploys regression; `run_dev_remote` plumbs canonical_refs into `ImageBuilder.build`.
- `image_builder.py`: Partial `image_refs` map falls back to legacy form.

**Gaps**: (a) no test for the live-build profile-gating leak (Medium #1) — adding one would force the design decision; (b) IPv6 host-literal handling per Medium #5.

## Positive Observations
- The `compute_canonical_refs` + `_canonical_build_ref` refactor genuinely eliminates the four-site drift the PR targets; every read-back through publish (live, dry-run, `--no-build`), dev, builder, and PATCH now agrees on the same derivation.
- Presence-based appgarden lookup (`name in appgarden` vs truthiness) is documented at both the call site and the helper, with a regression test (`test_empty_appgarden_entry_falls_through_to_legacy`) pinning the behavior that would otherwise re-emerge.
- New error message in `_auto_resolve_digests` is excellent: actionable causes, named ref, escape-injection-proof via `rich.markup.escape`, points to the `--digest` escape hatch. The change from silent soft-fall to hard-fail directly addresses ENG-4909's "fails loudly" mandate.
- PATCH now carries the `(registry, repository, tag)` triple; `test_repo_change_between_deploys_flows_through_patch` makes the `ImagePullBackOff` failure mode explicit and verifiable.
- `_split_image_ref` / `_replace_image_tag` correctly disambiguate registry-port colons from tag colons via the last-slash rule.
- Comments throughout explain *why* (the failure mode being prevented), not *what* — good signal/noise.

## Overall Verdict
**Risk Level**: Low
**Engine Agreement**: 3/3 engines contributed. Claude and Kamiwaza approve; Codex's P2 is a design-intent flag (the behavior is intentional per the PR title) rather than a defect. The medium-priority items — pre-existing profile-gating leak in live-build, missing `isinstance` guard, implicit coupling between `buildable_services` and `canonical_refs`, missing IPv6 test — are hardening/documentation notes that don't block this fix and can be addressed in follow-up.

---

---
<details>
<summary>🔧 Engine Health (3/3 contributing — High (3 engines agree))</summary>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |
| Kamiwaza | ✅ OK |
| Gemini | ⬜ Not available |

**Engines contributing**: 3/3
**Confidence**: High (3 engines agree)
</details>

PR #108, opened 2026-05-13
fix(extensions): resolve extra_docker_images tag + digest; stop merging into docker_images (ENG-5181)

JohnStrick:
## Summary

`extra_docker_images` now goes through the same substitution and digest-resolution path as compose buildable services. The two catalog fields stay disjoint — `docker_images` is compose-derived, `extra_docker_images` is author-declared — matching the legacy `make publish-registry` contract that downstream consumers (kind-load, marketplace tarball, export-images) already assume.

- `RegistryBuilder.build_entry()` resolves `{version}` and applies the stage-derived tag suffix to substituted refs under the configured registry. Literal tags (e.g. `helper:0.5.0`) and external refs pass through verbatim. Tag parsing splits after the last `/` so registry-with-port refs like `localhost:5000/org/agent` aren't misread.
- `kz-ext publish` feeds resolved extras through `_auto_resolve_digests`, skipping refs already pinned in `digest_map` so an explicit `--digest` user pin survives.
- Extras are no longer merged into `docker_images`.

Kaizen's per-conversation sandbox agent — a docker-bake target absent from compose — was the case where `agent:{version}` leaked unresolved into both lists. Live `apps.json` will need a republish after this lands to clear the existing `agent:{version}` literal.

Linear: [ENG-5181](https://linear.app/kamiwaza/issue/ENG-5181)

## Verification

- `uv run pytest tests/unit/extensions/` — 1188 passed, 1 skipped (pre-existing BSD `sed` skip), 0 regressions.
- New tests cover: placeholder substitution, stage suffix per stage, external-ref passthrough, already-pinned-ref passthrough, fixed-tag passthrough, registry-with-port parsing, digest pinning into `extra_docker_images`, list-disjointness, absent-field omission, `--digest` user pin survives when redundant in extras.
- Side-by-side against Kaizen's actual `kamiwaza.json` + compose: new entry has `docker_images` = 4 compose-derived refs (all digest-pinned) and `extra_docker_images` = `agent:1.8.13-dev@sha256:<digest>`, with agent removed from `docker_images`.
- Codex CLI review: 3 P2 findings, all addressed in this PR.

## Test plan

- [ ] Reviewer confirms the contract change (extras separate from `docker_images`) matches the downstream-consumer expectation in `kind-load-images.sh` and `export-images.py`.
- [ ] After merge: republish Kaizen's catalog so the live `apps.json` picks up the resolved + pinned agent ref.

## Follow-ups (separate tickets)

- Teach `kamiwaza-marketplace/.../downloads.py` + `tar_parser.py` to package `extra_docker_images` alongside `docker_images` in the marketplace tarball download path. Current state pre-dates this PR (the field has always existed); no regression here, just gap visibility.

### DIFF
--- a/kamiwaza_extensions/commands/publish.py
+++ b/kamiwaza_extensions/commands/publish.py
@@ -270,7 +270,10 @@ def run_publish(
         validate_digest,
     )
     from kamiwaza_extensions.profile_manager import ProfileManager
-    from kamiwaza_extensions.registry_builder import RegistryBuilder
+    from kamiwaza_extensions.registry_builder import (
+        RegistryBuilder,
+        resolve_extra_image,
+    )
     from kamiwaza_extensions.validators.compose import ComposeValidator
     from kamiwaza_extensions.validators.metadata import MetadataValidator
 
@@ -623,6 +626,25 @@ def run_publish(
                 _verify_supplied_digest(ref, digest)
         else:
             digest_map = _auto_resolve_digests(published_refs)
+
+    # Extras under our registry get the same tag-and-digest pinning as
+    # compose buildable services. External refs and author-pinned
+    # `@sha256:...` refs pass through untouched. Refs already in
+    # digest_map (e.g. the buildable service redundantly listed in extras
+    # when the user supplied `--digest`) are left alone — re-resolving
+    # would hit the registry and could overwrite an explicit user pin.
+    resolved_extras = [
+        resolve_extra_image(img, registry, version, stage)
+        for img in (info.metadata.get("extra_docker_images") or [])
+    ]
+    extras_to_resolve = [
+        r for r in resolved_extras
+        if r.startswith(f"{registry}/")
+        and "@" not in r
+        and r not in digest_map
+    ]
+    if extras_to_resolve:
+        digest_map.update(_auto_resolve_digests(extras_to_resolve))
 
     # 7. Build catalog entry
     reg_builder = RegistryBuilder()
--- a/kamiwaza_extensions/registry_builder.py
+++ b/kamiwaza_extensions/registry_builder.py
@@ -70,12 +70,14 @@ class RegistryBuilder:
                 are tagged with publish's revision; services without one
                 keep their declared image ref verbatim).
             registry: Docker registry prefix (e.g. ``"kamiwazaai"``).
-            version: Semver version string for this release.
+            version: Semver version string for this release. Substituted
+                into ``{version}`` placeholders in ``extra_docker_images``
+                entries.
             stage: One of ``"prod"``, ``"stage"``, ``"dev"``, or any custom name.
-                Currently unused in the body — ``ComposeTransformer`` already
-                applied the stage-derived tag to buildable services. Kept on
-                the signature for call-site compatibility with prior callers
-                of ``RegistryBuilder``.
+                Applied as a tag suffix to ``extra_docker_images`` entries
+                under *registry* (``<registry>/agent:{version}`` →
+                ``<registry>/agent:<version>-<stage>``), mirroring the
+                rewrite ``ComposeTransformer`` applies to compose services.
             revision: Optional revision identifier. When provided, included
                 as a top-level ``revision`` field on the entry; consumed by
                 ``CatalogDedupGuard`` to make CI re-publishes idempotent.
@@ -103,26 +105,20 @@ class RegistryBuilder:
         )
         docker_images = self.extract_docker_images(transformed_compose)
 
-        extra_images = metadata.get("extra_docker_images") or []
-        if extra_images:
-            # Apply the same digest-pinning rule to extras so a service
-            # ref that's redundantly listed in `extra_docker_images`
-            # collapses against its already-pinned compose copy during
-            # dedup, instead of leaking an unpinned duplicate.
-            #
-            # Match is exact-string against the post-stage-suffix ref
-            # (e.g. `<reg>/<ext>-<svc>:<version>-dev`); a pre-suffix
-            # entry like `<reg>/<ext>-<svc>:<version>` won't collapse.
-            # Author the entry to match what compose carries after
-            # transform.
-            if digest_map:
-                extra_images = [
-                    f"{img}@{digest_map[img]}"
-                    if img in digest_map and "@" not in img
-                    else img
-                    for img in extra_images
-                ]
-            docker_images = list(dict.fromkeys(docker_images + extra_images))
+        # `docker_images` and `extra_docker_images` are disjoint catalog
+        # fields: compose-derived vs. author-declared. Downstream consumers
+        # iterate each list separately; do not merge.
+        extra_images = [
+            resolve_extra_image(img, registry, version, stage)
+            for img in (metadata.get("extra_docker_images") or [])
+        ]
+        if extra_images and digest_map:
+            extra_images = [
+                f"{img}@{digest_map[img]}"
+                if img in digest_map and "@" not in img
+                else img
+                for img in extra_images
+            ]
 
         # Source kamiwaza.json is the catalog contract: every top-level
         # field the developer authored reaches the catalog entry. The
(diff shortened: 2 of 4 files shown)


jxstanford (commented):
Now I have a complete picture. Let me write the review report.

# Review Report: fix(extensions): resolve extra_docker_images tag + digest; stop merging into docker_images (ENG-5181)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/108
**Branch**: `feature/eng-5181-extra-docker-images` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Failing checks | None (per PR description: 1188 passed, 1 pre-existing skip) |

## Critical Issues (Must Fix)
| # | File:Line | Issue | Category | Confidence |
|---|-----------|-------|----------|------------|
| — | — | None | — | — |

## High Priority Issues
| # | File:Line | Issue | Category | Confidence |
|---|-----------|-------|----------|------------|
| — | — | None | — | — |

| # | File:Line | Issue | Category | Confidence |
|---|-----------|-------|----------|------------|
| 1 | `registry_builder.py:469` | `re.sub(r"-(dev\|stage)$", "", tag)` only strips `dev`/`stage` suffixes. If the stage is a custom name (e.g. `preview`) and the source tag carries the same custom suffix (`agent:{version}-preview`), the resulting catalog tag becomes `agent:1.8.13-preview-preview`. The dev/stage-only matrix matches the documented `_STAGE_SUFFIXES`, so this is an edge case rather than a defect, but worth either widening the regex or asserting against custom stages. | Logic correctness | Medium |
| 2 | `commands/publish.py:577,646` | `ImagePusher.check_buildx_available()` preflight only fires when `not no_push and buildable_services`. The new extras path on line 646 also calls `_auto_resolve_digests` (which needs `docker buildx imagetools`), but for an external-only extension with extras under the registry, no preflight runs — the publish would fail later inside the auto-resolve loop with a less-targeted error. UX nit, not a correctness defect. | Error-handling UX | Medium |
| 3 | `commands/publish.py:636` & `registry_builder.py:111` | `resolve_extra_image()` is invoked twice for the same metadata: once in `run_publish` to compute the auto-resolve filter and again in `RegistryBuilder.build_entry`. Idempotent, but a single resolved-list passed through to `build_entry` (e.g. via a new `extras_resolved` kwarg, or by passing `resolved_extras` directly) would prevent the two call sites from drifting if `resolve_extra_image` ever gains stage-dependent state. | Code duplication | Medium |
| 4 | `tests/unit/extensions/test_publish_cmd.py` | Missing `publish.py` test for "literal-tag extra under our registry" (e.g. `myreg/shared-helper:0.5.0`). The existing `test_extra_docker_images_fixed_tag_under_registry_passthrough` in `test_registry_builder.py` only proves `resolve_extra_image` returns it unchanged; it does NOT cover that `run_publish` then submits it to `_auto_resolve_digests` (the filter at L640 admits it because it starts with `registry/`, has no `@`, and isn't in `digest_map`). This is current behaviour but is asserted nowhere. | Test coverage | High |

## Test Coverage Assessment
- **Files with adequate coverage**: `kamiwaza_extensions/registry_builder.py` — placeholder substitution, stage suffix per stage, external passthrough, fixed-tag passthrough, registry-with-port, digest pinning, list disjointness, absent-field omission all covered.
- **Files needing tests**: `kamiwaza_extensions/commands/publish.py` — extras digest resolution path is covered for `{version}` extras, `--digest` user-pin survival, and external/already-pinned extras. Gaps: (a) literal-tag extras under registry passing through `_auto_resolve_digests` (Issue #4), (b) extras-only extension with no buildable services — confirms behaviour and exposes the missing buildx preflight from Issue #2.

## Positive Observations
- Diff is tight and squarely scoped to ENG-5181: one helper added, one merge dropped, one digest-resolution loop added — no incidental refactors.
- `resolve_extra_image` correctly handles the registry-with-port case by splitting on the last `/` first, then the first `:` in the trailing segment — addresses a real OCI-grammar pitfall and is explicitly tested.
- The `--digest` user-pin survival path is defended both by code (`r not in digest_map`) and by an `AssertionError` side_effect that proves `resolve_digest` is never called when the user pin is in play. Strong test posture.
- Comments give the *why* (legacy `make publish-registry` contract, immutable-identity, profiled-helper redirect risk) without re-explaining the *what*. Matches the codebase's existing commenting voice.
- `entry.pop("extra_docker_images", None)` correctly handles the deepcopy carryover when extras are absent — avoids leaking unresolved `{version}` placeholders into the catalog.
- No injection surface: `resolve_digest` invokes subprocess with a list (no `shell=True`), and the new `resolve_extra_image` does pure string manipulation against author-controlled metadata that already passes `MetadataValidator`.

## Overall Verdict
**Risk Level**: Low
**Rationale**: Correctly implements the disjoint `docker_images` / `extra_docker_images` contract, restores `{version}` and stage-suffix resolution for extras, and preserves the explicit `--digest` user pin. Suggestions are coverage and UX nits — none block merge.

**Engines**: Claude Code (1/3 available)

---
🤖 Generated with [Claude Code](https://claude.ai/code) PR review

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ❌ Failed (exit 124) |
| Gemini | ⬜ Not available |

**Engines contributing**: 1/3
</details>


kbot-ships (approved):
_Reviewing head `986780f` against `develop`._

## Bottom line

This PR correctly separates `extra_docker_images` from `docker_images` by removing the merge that previously conflated compose-derived refs with author-declared extras. The new `resolve_extra_image` function handles `{version}` substitution and stage-suffix application with proper edge-case handling for registry ports, external refs, digest-pinned refs, and literal tags. The architecture is sound, the double-resolution between `publish.py` and `build_entry` is deterministic and idempotent, and the test coverage is thorough. Direction is right; approve with minor notes.

## Findings

### 1. Pre-existing collision test comment and name are now misleading — `test_registry_builder.py:1279` **(nit)**

`test_extra_docker_images_collision_pinned_and_deduped` still uses wording about "dedup collapses them" and only asserts on `docker_images`. Under the new model there's no cross-field dedup — `extra_docker_images` will contain a separate pinned copy (I confirmed: both fields get `kamiwazaai/my-app-backend:1.0.0@sha256:aaa…`). The test passes by accident of the compose-derived path still working, but the comment/name describe behavior that no longer exists. Consider renaming and adding an `extra_docker_images` assertion in a follow-up.

### 2. Redundant resolution across two call sites — `publish.py:636` + `registry_builder.py:111` **(nit)**

`resolve_extra_image` is called on the raw `extra_docker_images` entries in both `publish.py` (to build digest_map keys) and `build_entry` (to build the catalog field). This is safe because the function is deterministic and idempotent (second call finds no `{version}` placeholder and short-circuits). However, it means the substitution + suffix logic runs twice per extra image. A minor efficiency comment — not worth changing now since the list is typically 1–3 items, but if this ever gets refactored, passing pre-resolved extras to `build_entry` would avoid the redundancy and the implicit coupling that both sites must receive identical `(registry, version, stage)` arguments.

## Test gaps

Test coverage is strong. The new tests cover the major axes: placeholder substitution, stage suffixing (prod/dev/stage), external-ref passthrough, fixed-tag passthrough, digest-pinned passthrough, registry-with-port parsing, digest pinning via digest_map, list disjointness, absent-field omission, and `--digest` user pin survival. The integration-level `test_publish_cmd.py` tests verify the end-to-end flow through `run_publish`.

One minor gap: no unit test for `resolve_extra_image` directly (all tests go through `build_entry` or `run_publish`). A focused parametrized test on the function itself would make it easier to reason about edge cases in isolation, but the current integration-level coverage is adequate.

## What's good

- The `resolve_extra_image` function is cleanly factored with clear short-circuit conditions. The `rfind("/")` approach for tag parsing correctly handles registry-with-port refs — a subtle OCI detail that's easy to get wrong.
- The stage-suffix stripping via `re.sub(r"-(dev|stage)$", "", tag)` prevents double-suffixing when an author writes `agent:{version}-dev` and publishes against a different stage.
- Keeping the two catalog fields disjoint matches the downstream consumer contract and eliminates the confusing dedup semantics of the old merge path.
- The test suite is extensive and well-documented with scenario-specific comments explaining the "why" behind each assertion.

## Verdict

**Verdict**: APPROVE

Clean, well-tested fix that correctly separates catalog concerns. The two nits are minor and neither affects correctness.
---

PR #129, opened 2026-05-27
feat(tests+notebooks): receive extension_contract suite, gate-packages lifecycle test, jupyter quickstart from kamiwaza

jxstanford:
## Summary

Relocates from `kamiwaza-internal/kamiwaza` (paired with the deletion PR there) so the SDK contract tests live alongside the SDK they exercise, not in the platform repo that creates a circular dep.

**End state of the paired kamiwaza PR**: zero `kamiwaza-sdk` dependency on the kamiwaza platform repo. The SDK contract tests live in the SDK repo where they belong; their CI signal flows on per-PR PRs to the SDK (the right side to catch SDK regressions); and ENG-5295 (PyPI publish) becomes optional rather than blocking for federation.

## What this PR adds

| Path | What | Source |
|---|---|---|
| `tests/e2e/extension_contract/` (38 files, ~3500 lines) | Full extension contract suite — `LiveExtensionHarness` machinery, support modules, echo-check FastAPI test extension + docker-compose stack, harness unit tests, smoke + auth contract tests | Relocated verbatim from `kamiwaza/tests/extension_contract/`. Internal Python imports rewritten from `tests.extension_contract.*` → `tests.e2e.extension_contract.*`. |
| `tests/integration/gate_packages/test_lifecycle.py` | Gate-package install/replace/uninstall lifecycle test against a live cluster (TS-M5-25, AC2/6/7/8/10/11) | Relocated from `kamiwaza/tests/integration/services/authz/gate_packages/`. Drops the try/except SDK import gate (SDK is now host). Adds `KAMIWAZA_VERIFY_SSL` env-var opt-out for dev certs — defaults to TLS verification on, warns if disabled (never silently insecure). |
| `scripts/start-jupyter-lab.sh` | Local JupyterLab launcher. Points at `examples/` (the SDK notebooks canonical home); reads `KAMIWAZA_BASE_URL`; kills prior `jupyter-lab` processes on rerun to avoid port collisions | Adapted from `kamiwaza/restart-or-start-lab.sh`. |
| `examples/.jupyter/jupyter_lab_config.py` | Minimal jupyter config (base_url=/lab, no auth, port 8890). **Local dev only** — do not expose to network without auth + TLS | Moved verbatim from `kamiwaza/notebooks/.jupyter/`. |
| `[project.optional-dependencies].notebook` in `pyproject.toml` | jupyterlab + ipykernel + ipywidgets + nbclient + nbformat + openai (15 deps) | Curated from `kamiwaza/notebooks/extra-requirements.txt` (60 lines → 15 direct deps; pip resolves the rest). |
| `.github/workflows/extension-contract-live.yml` | Per-PR + manual-dispatch CI workflow for the relocated suites. Probes live cluster availability before running; skips cleanly if `KAMIWAZA_BASE_URL` / `KAMIWAZA_API_KEY` secrets aren't configured | New. Nightly cron is a TODO, enabled once per-PR signal pattern stabilizes. |
| `docs/jupyter-quickstart.md` | Local-dev quickstart: `pip install kamiwaza-sdk[notebook]` → `./scripts/start-jupyter-lab.sh` → notebooks against k0s-lima dev cluster. Includes PAT-minting walkthrough, base-url override, TLS-verification guidance | New. |
| `README.md` | Examples section gains a link to `docs/jupyter-quickstart.md` | Minimal edit. |

Total: **48 files** changed.

## Audit findings that drove this design

1. **Zero `kamiwaza.*` platform imports** in the relocated tests — they only call the platform via HTTP. Public-repo move is clean.
2. **Both suites are already live-env-gated**: extension_contract via `RUN_LIVE_EXTENSION_TESTS=1`, gate_packages via multi-env-var gating (`KAMIWAZA_BASE_URL`, `KAMIWAZA_ADMIN_TOKEN`, `M5_TEST_WHEEL_DIR`, `M5_TEST_INDEX_URL`). They don't run in kamiwaza's regular CI today — relocating them loses no CI signal on the kamiwaza side.
3. **SDK repo is well-equipped**: existing `tests/{unit,integration,e2e}/`, `examples/`, `docs/extensions/`, plus precedent CI workflows like `extension-starter-smoke.yml` and `extension-go-reference-e2e.yml`.

## CI behavior

The new `extension-contract-live.yml` workflow:
- Triggers on PRs touching the relocated tests + on `workflow_dispatch`
- Probes `${KAMIWAZA_BASE_URL}/health` (5s timeout) before running tests; emits a CI warning and skips cleanly if unreachable
- Doesn't fail PRs when the cluster is down — degrades to a warning so non-cluster-related changes can still merge
- Will be promoted to a scheduled nightly job on `develop` once the per-PR signal pattern stabilizes (per jx)

## Test plan

- [ ] CI green on this branch (unit tests, lint, mypy, build)
- [ ] Workflow `extension-contract-live.yml` runs on this PR — if cluster secrets configured, full suite passes; if not, emits the expected skip warning
- [ ] Manual: `pip install -e .[notebook]` then `./scripts/start-jupyter-lab.sh` launches JupyterLab against the local k0s-lima cluster

## Paired PR

The kamiwaza-side deletion PR (TBD link) removes:
- `tests/extension_contract/` (relocated here)
- `tests/integration/services/authz/gate_packages/test_lifecycle.py` (relocated here)
- `notebooks/` (entire — sdk-examples duplicate `examples/` here; .jupyter/ config moved; extra-requirements.txt becomes the `notebook` extra here)
- `restart-or-start-lab.sh` (adapted to `scripts/start-jupyter-lab.sh` here)
- The dead-code SDK seam in `model_resolution.py` + 22 redundant unit tests
- `kamiwaza-sdk` from all kamiwaza pyproject dep groups + `[tool.uv.sources]` entry + deptry config

End result: kamiwaza no longer depends on its own SDK in any form.

## Related

- Supersedes `kamiwaza-internal/kamiwaza#1814` (closed) and `kamiwaza-internal/kamiwaza#1757` (closed)
- ENG-4994 — Phase 3 parent ticket
- ENG-5295 — F1: Publish kamiwaza-sdk 1.0.0 to PyPI (becomes optional after this work)


jxstanford:
# Review Report: feat(tests+notebooks): receive extension_contract suite, gate-packages lifecycle test, jupyter quickstart from kamiwaza

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/129
**Branch**: `chore/eng-relocate-extension-contract-tests-and-jupyter` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 8/10 | Clean relocation, import-path rewrite, drop of circular kamiwaza dep. Lowered from 9 because Codex flagged that the new CI workflow assumes a bootstrap artifact that won't exist on fresh GitHub checkouts (workflow design mismatch). |
| Security | 7/10 | Strong workroom-trust + log-sanitization + path-traversal defenses. Loses points for inconsistent `KAMIWAZA_VERIFY_SSL` default across the two relocated suites (extension contract = off, gate-packages = on) and permissive CORS in echo-check (acceptable for a test fixture but undocumented). |
| Test Quality | 6/10 | Harness machinery has thorough unit coverage; relocated suite preserves rich validation tests. Lowered from 8 because Codex confirmed (a) unit tests leak env vars when the live workflow sets `KAMIWAZA_BASE_URL`/`KAMIWAZA_API_KEY`, (b) `echo-check/conftest.py` ignores the `backend/` directory so the new FastAPI tests never run from the root, and (c) `test_atomic_replace_to_v2` / `test_uninstall` run even when `test_install_v1` skipped. |
| Error Handling | 7/10 | `_best_effort_close` and `wait_for_*` paths handle retries and cleanup cleanly. Loses points for retry-on-401/403 in `_wait_for_response` and bare `except Exception: pass` in `cleanup_acme`. |
| Code Quality | 8/10 | Cleanly modularized `support/` package, frozen dataclasses, `MappingProxyType` for immutable state. Small smells: unused `target` arg, dense single-line method-per-line block in `LiveExtensionHarness`. |
| Refactoring | 7/10 | Sensible boundaries; `auth_ops` / `build_ops` / `runtime_ops` separation reads well. Stale "see test_bind_then_replace" reference and permanently-skipped `TestNetworkPolicyProbes` should be cleaned up. |

## Critical Issues (Must Fix)
| 1 | `.github/workflows/extension-contract-live.yml:85-92` | The workflow sets `RUN_LIVE_EXTENSION_TESTS=1` and runs the extension contract suite with only `KAMIWAZA_BASE_URL` / `KAMIWAZA_API_KEY` secrets. The session fixture calls `load_live_routed_integration_state()`, which on a fresh GitHub checkout will not find `../deploy/.artifacts/live-routed-integration/bootstrap-state.json`, then `pytest.fail`s with "no bootstrap state was found" before the API key is ever used. Either materialize `LIVE_ROUTED_INTEGRATION_STATE` in the workflow or make bootstrap state optional for API-key-only smoke coverage. | Codex | CI correctness |
| 2 | `tests/e2e/extension_contract/test_harness_runtime_ops.py:76-100` (and lines 119-123) | Unit tests that exercise `LiveExtensionSettings.from_env()` only `delenv` `KAMIWAZA_API_URL` / `KAMIWAZA_APP_ORIGIN` / `KAMIWAZA_VERIFY_SSL` / `KAMIWAZA_USERNAME` / `KAMIWAZA_PASSWORD`, leaving `KAMIWAZA_BASE_URL` and `KAMIWAZA_API_KEY` in the environment. The new CI workflow sets both for the whole `pytest tests/e2e/extension_contract/` run, so `from_env()` will prefer the workflow secrets over the fixture bootstrap state and `assert settings.base_url == "https://kamiwaza.test/api"` will fail. Clear all `from_env()` env inputs (`KAMIWAZA_BASE_URL`, `KAMIWAZA_API_KEY`) in these tests. | Codex | Test isolation |

## High Priority Issues
| 1 | `tests/e2e/extension_contract/support/settings.py:78` vs `tests/integration/gate_packages/test_lifecycle.py:56` | `KAMIWAZA_VERIFY_SSL` defaults to **False** in the extension contract suite (silently disables warnings) and to **True** in the new gate-packages test (warns when disabled). Two live-test suites in the same repo now disagree on the secure default. The PR description asserts a "never silently insecure" contract; unify both to verify-on default with an explicit opt-out warning. | Claude | Security misconfiguration |
| 2 | `.github/workflows/extension-contract-live.yml` (job-level) | No `timeout-minutes` on the `contract` job. Default deploy polling is 900s and probes default to 180s; multiple hung probes can run for hours before GitHub's 6-hour cutoff. Add `timeout-minutes: 30` (or similar). | Claude | CI / Reliability |
| 3 | `tests/integration/gate_packages/test_lifecycle.py:128-178` | `test_list_and_get` / `test_atomic_replace_to_v2` / `test_uninstall` still execute when `test_install_v1` skipped via its fixtures (missing `M5_TEST_WHEEL_DIR` / `M5_TEST_INDEX_URL`). The workflow comment claims the suite self-skips, but in practice these dependent tests run and assert against state that was never installed — on a live cluster the autouse cleanup removes any prior package and `assert "acme-gates" in names` fails. Gate the whole class/module on the wheel/index fixtures (e.g., a class-scoped fixture that calls `pytest.skip` when secrets are missing). | Codex | Test correctness |
| 4 | `tests/e2e/extension_contract/echo-check/backend/app/workroom_trust.py:91` + `Dockerfile` CMD | `trusted_routed_workroom_context()` requires `request.scope['root_path']` to equal `KAMIWAZA_APP_PATH`. The container starts `uvicorn app.main:app` without `--root-path`, so unless the platform proxy injects an ASGI root path, real routed requests will not populate `scope['root_path']` from the env var or `X-Forwarded-Prefix`. `/api/whoami` and workroom-check responses will drop identity fields and return 403 in the live deployment. Either pass `--root-path "$KAMIWAZA_APP_PATH"` to uvicorn at startup or validate the routed path/prefix differently. | Codex | Security / runtime correctness |
| 5 | `tests/e2e/extension_contract/echo-check/conftest.py:1` | `collect_ignore = ["backend"]` prevents pytest from descending into `backend/tests` during the root and workflow test runs, so the newly added FastAPI auth/runtime/security tests are never exercised — despite the PR hoisting `PyJWT` into root dev deps specifically for them. Either remove the ignore (and gate via markers) or run the nested suite explicitly in the workflow. | Codex | Test coverage |
| 6 | `tests/e2e/extension_contract/support/runtime_ops.py:149-163` | `_wait_for_response` retries on **any** non-200 (including 401/403). An auth misconfig blocks for the full `probe_timeout_seconds` (default 180s) before failing. Consider fail-fast on 4xx auth statuses. | Claude | Error handling / DX |
| 7 | `tests/integration/gate_packages/test_lifecycle.py:138-162` | `test_atomic_replace_to_v2` docstring claims "Bind acme-gates as an AttributeGate on a test dataset first to exercise the binding-aware classpath-superset check" and references a `test_bind_then_replace` that does not exist. Either implement the bind step or strip the misleading comment. | Claude | Test correctness |

| 1 | `tests/e2e/extension_contract/echo-check/docker-compose.yml:6` + `docker-compose.appgarden.yml:3` | Image namespace still references the source repo: `ghcr.io/kamiwaza-internal/kamiwaza/images/echo-check-app:*`. After relocation into the SDK repo the namespace is misleading; local-build retagging via `_align_built_images_to_template` masks this for the smoke path, but the published template tag still points elsewhere. | Claude | Code quality / convention |
| 2 | `tests/integration/gate_packages/test_lifecycle.py:181-208` | `TestNetworkPolicyProbes` — all three methods unconditionally `pytest.skip()` with hard-coded reasons. They're permanently manual stubs masquerading as automated tests. Move to docs or delete. | Claude | Code quality |
| 3 | `tests/e2e/extension_contract/support/settings.py:130-131` | `_origin_probe_succeeded(target, response)` declares `target` but never uses it. Minor code smell. | Claude | Code quality |
| 4 | `tests/integration/gate_packages/test_lifecycle.py:95-104` | Module-scope `cleanup_acme` uses bare `except Exception: pass` for both setup and teardown. Silently swallows transport failures that would leave the cluster in unknown state. Consider `except APIError as exc: logger.warning(...)`. | Claude | Error handling |
| 5 | `tests/e2e/extension_contract/test_harness_state_validation.py:126` | Test reaches into `deployed_app_contract.__wrapped__` to call the underlying generator. Fragile across pytest versions. Factor the deployed-app body into a plain function and a thin fixture wrapper, then test the plain function. | Claude | Test brittleness |
| 6 | `tests/e2e/extension_contract/support/auth_ops.py:68-76` | `auth_headers_for_role` falls back from `client.get_bearer_token()` to a raw PAT wrapped as `Authorization: Bearer <pat>`. If the server distinguishes bearer JWT vs PAT (e.g., requires `X-API-Key` for PATs), this fallback will 401. Worth confirming the auth middleware accepts PAT-as-Bearer. | Claude | Correctness |
| 7 | `tests/e2e/extension_contract/echo-check/backend/app/main.py:231-237` | CORS uses `allow_methods=["*"]` + `allow_headers=["*"]` + `allow_credentials=True`. Acceptable for a test fixture but worth a comment that it's intentionally permissive — readers may copy it into a non-test extension. | Claude | Security misconfiguration |
| 8 | `tests/e2e/extension_contract/echo-check/backend/tests/test_helpers.py:46-52` | Test fixture builds an unsigned JWT (`algorithm="none"`). Acceptable for tests, but several JWT libs reject `alg=none` on decode by default; annotate that the kamiwaza_extensions_lib `/api/session` endpoint accepts unsigned tokens only in test mode. | Claude | Test rigor |
| 9 | `tests/e2e/extension_contract/echo-check/README.md:32` | Local-build example uses `kamiwazaai/echo-check-app:0.2.0-dev` but the compose file uses `ghcr.io/kamiwaza-internal/...`. Disambiguate that the local build is then retagged to the appgarden image by `_align_built_images_to_template`. | Claude | Stale/misleading example |
| 10 | `tests/e2e/extension_contract/README.md:24` | Documents `make dev-full-local from ../deploy` and a sibling `deploy/` checkout layout — that cross-repo coupling now travels with the SDK repo and is no longer obvious. Document the expected layout. | Claude | Cross-repo coupling |

## Test Coverage Assessment
- **Adequately covered**: `support/harness.py`, `build_ops.py`, `state.py` / `state_loader.py`, `runtime_ops.py`, and the echo-check FastAPI app (auth endpoints, runtime endpoints, security helpers) — **provided the `collect_ignore` block in `echo-check/conftest.py` is removed or the backend suite is run explicitly in CI** (currently those tests never execute from the root).
- **Gaps**:
  - `tests/integration/gate_packages/test_lifecycle.py`: only the success path is covered; the 409 `uninstall_blocked` path is unexercised and the bind-aware classpath-superset replace path is documented but not implemented. Dependent tests also run when the install step skipped (see High #3).
  - `tests/e2e/extension_contract/support/auth_ops.py::_validate_client`: success and `APIError`/`RequestException` branches are not directly unit-tested (only exercised via the live-only conftest).
  - `LiveExtensionSettings.from_env()` unit tests do not isolate `KAMIWAZA_BASE_URL` / `KAMIWAZA_API_KEY` from the ambient environment (see Critical #2).

## Positive Observations
- Three fixup commits already addressed the local-review findings: deleted the insecure `jupyter lab` launcher (no more `token=""` / `password=""` / `--ip=0.0.0.0`), registered the `live_extension` marker, hoisted `PyJWT` into root dev deps, fixed `verify=settings.verify_ssl` propagation in `client_for_role`/`build_live_client`, and switched the cluster-availability probe to `/ping` (matching `tests/integration/test_ping_live.py`).
- Path anchoring fix (`EXTENSION_FIXTURES_ROOT = Path(__file__).resolve().parent.parent`) is robust to future relocations.
- Strong defensive validation for shell-execed args: `kubectl_secret_value` validates resource names against a k8s-safe regex; `safe_token_file_path` rejects traversal to roots outside bootstrap/REPO_ROOT/`~/.kamiwaza`; `resolve_deploy_login` enforces a trusted-roots check on the `kz-login` helper.
- `safe_log_field` (echo-check) strips CR/LF/`=`/control chars and truncates to 256 — clean log-injection defense.
- Workroom-trust logic in echo-check is well-designed in principle: forwarded identity headers are only trusted when the routed `root_path` matches the runtime prefix AND auth is enabled (defense against direct-container header spoofing). See High #4 for the deployment-config caveat.
- Test coverage of negative paths is solid where it runs: spoofed-header rejection, untrusted root-path rejection, invalid UUID rejection, traversal rejection, malformed bootstrap-state rejection.
- CI workflow correctly gates fork PRs (`pull_request.head.repo.full_name == github.repository`) and degrades gracefully when secrets are absent.

## Overall Verdict
**Risk Level**: Medium
**Verdict**: REQUEST_CHANGES

Rationale: Pure relocation PR with three good fixup commits, but Codex surfaced two Critical issues the initial review missed — the new CI workflow can't actually green (no bootstrap artifact on a fresh checkout) and the env-isolation bug in the runtime-ops unit tests will fail under that same workflow. A Linear ticket reference (`ENG-…` in the branch name) is present, so the policy gate is satisfied; the verdict is driven by the Critical findings, the dimension floor at Test Quality 6/10, and a security/runtime concern (echo-check `root_path` config) that would cause silent identity loss in live deployments.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

jxstanford:
# RE-REVIEW Report: feat(tests+notebooks): receive extension_contract suite, gate-packages lifecycle test, jupyter quickstart from kamiwaza

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/129
**Branch**: `chore/eng-relocate-extension-contract-tests-and-jupyter` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 8/10 | Implementation matches the stated relocation scope; jupyter launcher deletion is sensible. Lowered 1 point because the live workflow as-shipped (Codex + Claude) does not actually exercise the cluster — bootstrap state is never materialized, so `live_extension` tests silently skip even when secrets are configured. |
| Security | 9/10 | TLS verification default flipped to True consistently; auth header spoofing tests preserved; jupyter footgun removed entirely. Test JWT uses `algorithm="none"` (intentional, test-only). |
| Test Quality | 6/10 | 59 passing tests is strong coverage, but the fixup commit introduced two silently-skipping tests that no longer validate the new behavior (Highs 1+2 — verified reproducers), an autouse-scope bug that over-skips unrelated test classes (High 3), and a teardown order bug that mutates a live cluster even when the suite is configured to skip (High 4 — Codex). Multiple compounding test-quality regressions in the same patch. |
| Error Handling | 8/10 | Auth fail-fast (401/403) is well-placed and well-commented; bootstrap-state skip-vs-fail distinction is principled (though over-applied per High 6); `cleanup_acme` bare except is best-effort but doesn't log. |
| Code Quality | 8/10 | Modules well-separated (auth_ops, build_ops, runtime_ops, settings, state); comments explain the *why* on contentious changes (verify_ssl default, --root-path, state_loader skip-vs-fail). Single-line method delegation in `harness.py` is dense but consistent. |
| Refactoring | 8/10 | Path anchoring (`EXTENSION_FIXTURES_ROOT`) refactor was correct. Auto-fixture should be scoped to the class that needs it (High 3). |

## Critical Issues (Must Fix)

## High Priority Issues
| 1 | tests/e2e/extension_contract/test_harness_state_resolution.py:165–170 | `test_live_routed_integration_state_fails_when_no_bootstrap_candidate_exists` silently SKIPS instead of validating. The fixup commit changed `state_loader.py:48` from `pytest.fail` to `pytest.skip`, but the test still uses `pytest.raises(pytest.fail.Exception, match="no bootstrap state was found")`. The `Skipped` exception isn't caught by that `pytest.raises`, so the test is reported as SKIPPED (verified locally: `1 skipped`). Test name + assertion both reference "fails" behavior that no longer exists. Update to `pytest.raises(pytest.skip.Exception, ...)` and rename, or delete. | Claude | Test Quality |
| 2 | tests/e2e/extension_contract/test_harness_state_validation.py:17–26 | `test_bootstrap_state_returns_none_when_pytest_fail_is_mocked` silently SKIPS. The test mocks `pytest.fail` (line 21) to capture the failure message, but the fixup changed `state_loader.py:21` to call `pytest.skip(...)` for the `FileNotFoundError` branch. `pytest.skip` is not mocked, so it actually raises `Skipped` and the test skips with the exact source-line message. Neither `assert load_live_routed_integration_state() is None` nor `assert failures == ["missing bootstrap"]` actually runs. | Claude | Test Quality |
| 3 | tests/integration/gate_packages/test_lifecycle.py:107–122 | `_require_wheel_and_index` is a module-level autouse fixture with `scope="class"`, so it applies to **every** TestCase class in the module — not just `TestLifecycle`. As a result, `TestRegression::test_default_execution_gate_unchanged` (which only needs `kz` / KAMIWAZA_BASE_URL + KAMIWAZA_ADMIN_TOKEN) and `TestNetworkPolicyProbes` (which already self-skips) both incorrectly skip with a misleading "M5_TEST_WHEEL_DIR / M5_TEST_INDEX_URL" message whenever those env vars are absent. Move the fixture inside `class TestLifecycle:` (or attach via `@pytest.mark.usefixtures` only on TestLifecycle). | Claude | Test Quality / Logic |
| 4 | tests/integration/gate_packages/test_lifecycle.py:92–97 | `cleanup_acme` is a `scope="module"` autouse fixture that authenticates and runs `kz.gates.packages.uninstall("acme-gates")` during setup. Pytest sets up module-scoped fixtures BEFORE class-scoped fixtures, so when `_require_wheel_and_index` (class scope) would skip the lifecycle suite for missing `M5_TEST_WHEEL_DIR`/`M5_TEST_INDEX_URL`, `cleanup_acme` has already mutated the live cluster. The workflow comment claims this suite "self-skips" without side effects when those env vars aren't set — that's not true today. Either move `cleanup_acme` to class scope or gate it on the same env vars `_require_wheel_and_index` checks. | Codex | Test Quality |
| 5 | .github/workflows/extension-contract-live.yml:99–108 | The "Run echo-check backend unit tests" step runs `uv run pytest tests/` from `tests/e2e/extension_contract/echo-check/backend`. `backend/pyproject.toml` declares pytest only in `[project.optional-dependencies].test` — not in base deps or a dependency-group. `uv run` from that dir creates a fresh `backend/.venv` from base deps, which does not include pytest. The root `uv sync --group dev` does not populate this nested env. On Ubuntu CI runners (no system pytest), the step will fail with "Failed to spawn: pytest" the first time secrets are configured. Use `uv run --extra test pytest tests/` (or move pytest to a `[dependency-groups].dev` in backend/pyproject.toml). Verified reproducer locally. | Claude, Codex | CI / Workflow |
| 6 | .github/workflows/extension-contract-live.yml:90–97 + tests/e2e/extension_contract/support/state_loader.py:48–51 | Workflow sets `RUN_LIVE_EXTENSION_TESTS=1` and provides `KAMIWAZA_BASE_URL`/`KAMIWAZA_API_KEY`, but does NOT materialize the cross-repo bootstrap state file. `state_loader.py` then `pytest.skip()`s the entire session when no bootstrap candidate exists, so every `live_extension`-marked test silently skips and the job reports SUCCESS without exercising any actual contract behavior — making the workflow's signal value zero. Either materialize a minimal bootstrap-state.json in the workflow (or accept env-only configuration when bootstrap state is absent), or add an explicit guard step that fails when `live_extension` tests all skip while `RUN_LIVE_EXTENSION_TESTS=1`. | Claude, Codex | CI / Coverage |

| 7 | tests/e2e/extension_contract/echo-check/docker-compose.yml:6 & docker-compose.appgarden.yml:3 | Image namespace still references the source repo: `ghcr.io/kamiwaza-internal/kamiwaza/images/echo-check-app`. After relocation to kamiwaza-sdk, this should be `ghcr.io/kamiwaza-ai/kamiwaza-sdk/...` (or wherever the published image will live). Image-publish path will need to handle this. | Claude | Architecture |
| 8 | tests/e2e/extension_contract/echo-check/backend/Dockerfile:33 | `--root-path "${KAMIWAZA_APP_PATH:-}"` is set, while `app/main.py:209-219` also `include_router(..., prefix=runtime_prefix)`. Combination only works if the upstream proxy does NOT strip the prefix. If a future deployment configures the proxy to strip the prefix, all requests would 404. A one-line comment in `main.py` near `_register_routes()` documenting this assumption would prevent confusion. | Claude | Architecture / Docs |
| 9 | tests/integration/gate_packages/test_lifecycle.py:97–104 | `cleanup_acme` uses bare `except Exception: pass`. Acceptable for best-effort teardown, but at minimum log the exception so a real auth/connectivity failure isn't masked. | Claude | Code Quality |

## Test Coverage Assessment
- **Strong**: 11 harness modules with 59 passing unit tests; separate echo-check backend suite covering identity propagation, header spoofing rejection, and workroom boundary checks; install/list/replace/uninstall happy paths for gate-packages.
- **Regressed by the fixup commit**: Two state_loader skip paths now have tests that ran-but-don't-actually-validate (Highs 1, 2 — verified `1 skipped` reproducer). The autouse fixture scope error (High 3) silently drops `TestRegression` coverage on common CI configurations. The teardown-order bug (High 4) means the gate-packages suite can mutate the cluster while reporting "skipped." Live workflow (High 6) provides no contract-exercising coverage as configured.

## Positive Observations
- Self-review fixup correctly addresses the prior Critical workflow failure on fresh checkout and the env-leak in 4 settings tests (explicit `monkeypatch.delenv(...)`).
- `KAMIWAZA_VERIFY_SSL` default now consistent between extension_contract and gate_packages suites (both default True, both warn / require explicit opt-out). "Never silently insecure" contract honored.
- `_wait_for_response` 401/403 fail-fast (runtime_ops.py:159-168) is clean and well-commented; saves ~3 minutes per probe on auth misconfigs.
- Dockerfile `--root-path` propagation via `sh -c` wrapper is the right shape; default of `""` is safe.
- Local-review fixup chose deletion over hardening for jupyter — the right call. Eliminates an unauthenticated-on-all-interfaces footgun from the SDK surface.
- `live_extension` pytest marker properly registered in `pyproject.toml`, avoiding `--strict-markers` collection abort.
- `PyJWT>=2.8` correctly elevated to root dev group (was previously only in nested backend optional extra → ModuleNotFoundError on fresh checkouts).
- Path-anchored fixtures (`EXTENSION_FIXTURES_ROOT = Path(__file__).resolve().parent.parent`) survive future relocations — proactive fix.
- CI workflow has 30-min job timeout, fork-PR gate via `if:` condition on head.repo, correct probe target `/ping`.

## Overall Verdict
**Risk Level**: Medium
**Verdict**: REQUEST_CHANGES

Two engines independently confirm the live workflow as-shipped will either silently skip (no bootstrap state materialized → Codex + Claude High 6) or fail at the nested backend step (`uv run pytest` won't find pytest → Claude High 5, Codex High 5) the first time secrets are configured. Beyond CI surface, the fixup commit introduced four real test-quality regressions: two tests silently skip instead of validating the new behavior (Highs 1+2), an autouse fixture over-skips unrelated test classes (High 3), and the gate-packages cleanup fixture mutates the live cluster before the env-gate skip takes effect (High 4). Test Quality dimension at 6/10 forces REVISE under the gate rules. These are all small, mechanical fixes — tighten the test assertions, scope the fixtures correctly, install the right pytest extra, materialize or env-encode the bootstrap state — but they should land before merge because the test suite no longer guards the behavior the fixup commit message claims it does.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

jxstanford:
# RE-REVIEW Report: feat(tests+notebooks): receive extension_contract suite, gate-packages lifecycle test, jupyter quickstart from kamiwaza

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/129
**Branch**: `chore/eng-relocate-extension-contract-tests-and-jupyter` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 9/10 | Implementation matches relocation scope; image namespace deferral explicitly tracked. Codex agrees scope is clean. |
| Security | 9/10 | TLS-verify defaults consistent, workroom trust hardened (24 tests), kubectl/bootstrap arg validation, token file traversal rejection. Test-only JWT `algorithm="none"` is in-context. |
| Test Quality | 8/10 | 61+34 tests pass; silent-skip regressions fixed. Lowered 1 point: Codex caught a shared-token-store hazard in `auth_ops.py` (persona clients may load each other's cached tokens) and missing `live`/`e2e` markers that could leak past `make test`'s exclusion filter. |
| Error Handling | 9/10 | Best-effort cleanup logs warnings, fast-fail on 401/403, OSError on candidate probing surfaces as malformed-state, `_best_effort_close` doesn't mask teardown failures. |
| Code Quality | 9/10 | Frozen dataclasses, type hints, well-commented trade-offs (--root-path Dockerfile, collect_ignore rationale, _wheel_and_index_configured trap). |
| Refactoring | 8/10 | Minor duplication between `cleanup_acme` and `_require_wheel_and_index` env-var checks (extracted to helper). |

## Critical Issues (Must Fix)

## High Priority Issues
| 1 | `tests/e2e/extension_contract/support/auth_ops.py:47-55` | `UserPasswordAuthenticator` is constructed without a `token_store` override, so it falls back to the shared `~/.kamiwaza/token.json` (`FileTokenStore`). `_load_cached_token()` ignores the username, so the second persona client loaded in a test run can pick up the first persona's still-valid cached token. That would silently let `auth_headers_for_role("allowed_non_admin")` mint admin-tier headers, masking the very spoofing/workroom-boundary contracts these tests are meant to enforce. Pass per-persona `FileTokenStore(path=...)` instances or an in-memory store. | Codex (single-source) | Security/Test Correctness |

| 1 | `.github/workflows/extension-contract-live.yml:83` | Probe uses `curl --max-time 5 -sf "${KAMIWAZA_BASE_URL}/ping"`, which treats the auth-fronted `401 Not authenticated` response as failure and skips the live job. The harness's own `ping_response_ok` accepts that 401. On clusters that protect `/ping`, every downstream step silently skips even with secrets configured. Either send the bearer token or accept the documented 401 shape. | Codex (single-source) | CI/Reliability |
| 2 | `tests/e2e/extension_contract/test_echo_check_deploy_smoke.py:10`, `test_echo_check_auth_contracts.py:7` | Modules only carry `pytestmark = pytest.mark.live_extension`. The project's standard exclusion filter (`-m "not integration and not live and not e2e"`) won't match `live_extension`, so a developer running `make test` with `RUN_LIVE_EXTENSION_TESTS=1` in their env would collect cluster-mutating tests in the "unit" lane. The session fixture would still skip without bootstrap state, but it's brittle — add `pytest.mark.live` (or `e2e`) alongside `live_extension`. | Codex (single-source) | Test Hygiene |
| 3 | `tests/e2e/extension_contract/support/state_loader.py:23-29` | `candidate.exists()` raising `OSError` (e.g. `PermissionError`) is bucketed into `invalid_candidates` and may trigger `pytest.fail("No usable bootstrap state…")`. Failure message reads as "operator provided a malformed file" when the real cause is "couldn't determine if a file exists." Pre-existing from relocation; documented by a dedicated test. | Claude (single-source) | Logic/UX |
| 4 | `tests/e2e/extension_contract/echo-check/backend/tests/test_helpers.py:48-52` | Test-minted JWT uses `algorithm="none"` and empty key. Intentional (extensions-lib test-mode auth doesn't validate signatures), but an explicit "test-only fixture, never hits real verification" comment would future-proof against readers misreading it as a security smell. | Claude (single-source) | Code Quality |
| 5 | `tests/e2e/extension_contract/echo-check/docker-compose*.yml:3,6` | Image namespace still `ghcr.io/kamiwaza-internal/kamiwaza/images/echo-check-app`. Explicitly deferred (M7) per commit 107ed82 and consistent with rest of codebase; requires coordinated image-publish-pipeline change. | Claude (single-source) | Architecture |
| 6 | `pyproject.toml:123` | `PyJWT>=2.8` in root `[dependency-groups].dev` is defensive — backend tests are excluded from root collection via `collect_ignore = ["backend"]`. Harmless but not strictly necessary; CI step uses `--extra test` to install backend deps. | Claude (single-source) | Code Quality |

## Test Coverage Assessment
- **Adequate coverage**: 61 harness/state/build/runtime unit tests + 34 echo-check backend tests including 24 dedicated workroom-trust security tests covering identity-header spoofing, cross-deployment isolation, auth-disabled bypass, and untrusted forwarded headers.
- **Gaps**:
  - `tests/integration/gate_packages/test_lifecycle.py` itself has no unit coverage (acceptable — live-cluster test, structurally validated through code review).
  - The High finding above means the persona-spoofing live tests may not actually validate what they claim until the token-store sharing is fixed — coverage exists but its credibility is compromised.

## Positive Observations
- Re-review delta is clean: all 6 prior High findings + 1 Medium properly addressed in commit 107ed82; test count moved from 59 (with 2 silent skips) → 61 (real coverage).
- Excellent self-review discipline: author caught a `pytest.raises(pytest.fail.Exception)` silently passing when code switched to `pytest.skip`.
- Workflow scope correctly trimmed — removing the zero-signal live step rather than letting it produce SUCCESS is the right call; YAML comment explains the trap.
- Fixture scope correctness around module-vs-class ordering (`_require_wheel_and_index` + module-scope `cleanup_acme` gating on the same env vars).
- TLS-verify defaults made consistent (`KAMIWAZA_VERIFY_SSL=True`) with explicit opt-out + warning — matches PR description's "never silently insecure" claim.
- Auth fast-fail in `_wait_for_response` saves CI time on operator misconfigs (401/403 fails immediately instead of retrying for 180s).
- JWT/PyJWT dep gap correctly diagnosed and fixed at both backend-extra and root-dev declaration sites.

## Overall Verdict
**Risk Level**: Low

Per decision rules: no Critical issues, all dimensions ≥ 7, no quality-gate FAIL, Linear ticket references present in commit messages (ENG-4770, ENG-5038, ENG-5784, etc.). Codex's High finding on token-store sharing is a real test-correctness gap that should be addressed in a follow-up but is not a blocker for this relocation PR — the security contract being tested still holds in production code; only the test's ability to detect a regression is at risk. Recommend filing a small follow-up to inject `FileTokenStore` per persona (or use an in-memory store) before the next time the persona-spoofing tests are relied on for sign-off.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

jxstanford:
# RE-REVIEW Report: feat(tests+notebooks): receive extension_contract suite, gate-packages lifecycle test, jupyter quickstart from kamiwaza

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/129
**Branch**: `chore/eng-relocate-extension-contract-tests-and-jupyter` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 9/10 | Relocation matches stated scope; no scope creep. Jupyter launcher legitimately dropped after local-review feedback. |
| Security | 8/10 | TLS verify-on default, shell arg validation, layered trust model in workroom helpers. `ghcr.io/kamiwaza-internal` namespace and persona token_store sharing remain (latter raised independently by Codex). |
| Test Quality | 8/10 | 61 passing, validation paths well-parametrized. Lowered from 9 because Codex independently confirmed the persona token_store isolation gap (ENG-5955) — persona-spoofing tests cannot reliably detect privilege escalation until fixed. |
| Error Handling | 8/10 | `cleanup_acme` now logs, `state_loader` distinguishes missing vs malformed, validator clients distinguish `APIError`/`RequestException`/`OSError`. Minor edge: `PermissionError` on `exists()` treated as malformed. |
| Code Quality | 8/10 | Clean module decomposition (auth_ops/build_ops/runtime_ops/state/settings), comments explain non-obvious "why". One Codex-identified concern with `--root-path` + prefixed routing pattern. |
| Refactoring | 9/10 | Module split appropriate; harness delegation pattern is clear; nothing significant left on the table. |

## Critical Issues (Must Fix)
_(none)_

## High Priority Issues
| 1 | `tests/e2e/extension_contract/support/auth_ops.py:50-54,97-101` | `UserPasswordAuthenticator` constructed without `token_store` override. All persona clients (and the primary harness client) share `~/.kamiwaza/token.json`, so a second persona may load the admin's cached token and authenticate as the wrong identity. Persona-spoofing / workroom-boundary tests cannot reliably detect privilege-escalation regressions until each authenticator gets a per-client (or in-memory/disabled) `token_store`. Tracked as ENG-5955 but materially affects test fidelity. | Claude Code + Codex | Test fidelity / security |
| 2 | `tests/e2e/extension_contract/echo-check/backend/Dockerfile:33` + `app/main.py:208-219` | Combining `uvicorn --root-path "$KAMIWAZA_APP_PATH"` with routes registered under `prefix=runtime_prefix` is contradictory: Starlette's `get_route_path()` strips `root_path` from `scope["path"]` before route matching, so routed requests resolve to e.g. `/api/runtime` while routes are registered as `/runtime/apps/<id>/api/runtime` → 404 on production deploys. `/health` still works (unprefixed), masking the issue. The unit test `test_runtime_prefix_routes_work` passes because `TestClient` is constructed without `root_path`, so the bug isn't exercised. Either drop `--root-path` and rely on the application-level `prefix`, or drop the application-level prefix and rely on `root_path`. | Codex | Code Quality / Production bug |

| 1 | `tests/e2e/extension_contract/echo-check/docker-compose.{yml,appgarden.yml}:6,3` | `ghcr.io/kamiwaza-internal/...` namespace remains. Acknowledged/deferred (needs image-publish-pipeline coordination); recommend tracking-ticket reference inline. | Claude Code | Code Quality |
| 2 | `.github/workflows/extension-contract-live.yml:84` | Probe `curl --max-time 5` is tight for cluster cold-start. Consider 10s — current value will silently skip downstream steps on slow first response. | Claude Code | Robustness |
| 3 | `tests/e2e/extension_contract/support/build_ops.py:191-198` | `(context / dockerfile_rel).resolve()` accepts whatever the docker-compose YAML says without root-containment check. Internal test fixture so risk is low. | Claude Code | Security (defense-in-depth) |
| 4 | `tests/e2e/extension_contract/support/state_loader.py:38-47` | `OSError` from `candidate.exists()` aggregated into `invalid_candidates` and triggers `pytest.fail`; a transient `PermissionError` would now fail the suite where the previous design would skip. | Claude Code | Error handling |
| 5 | `tests/integration/gate_packages/test_lifecycle.py:62-70` | `warnings.warn(...)` for TLS opt-out is good, but a single `logger.warning(...)` would surface in pytest's captured output more reliably than `UserWarning` (often hidden unless `-W error`). | Claude Code | Code Quality |
| 6 | `README.md:379` | Says set `KAMIWAZA_BASE_URL` / `KAMIWAZA_API_KEY` to run notebooks, but existing notebooks (e.g. `examples/01_download_and_deploy.ipynb`) hard-code `"http://localhost:7777/api/"`. SDK does fall back to env when constructed without args, but users following the README will still need to edit notebooks. | Claude Code | Stale/inaccurate guidance |
| 7 | `tests/e2e/extension_contract/echo-check/backend/Dockerfile:1-8` | Comment correctly flags in-repo `kamiwaza_extensions_lib` vs PyPI mismatch — keep, but add a Linear ticket reference inline (currently just "see PR #129 description") so future maintainers find it. | Claude Code | Reference rot risk |

## Test Coverage Assessment
- **Files with adequate coverage**: `support/state.py` (validation cases parametrized); `support/state_loader.py` (3 paths: missing-state skip, malformed-state fail, valid candidate fallback); `support/settings.py` (origin-health branches + bootstrap-state defaults + control-plane-role-key gating); `support/build_ops.py` (build/push/timeout + image alignment); `echo-check/backend/app/main.py` and `workroom_trust.py` (security helpers thoroughly covered).
- **Files needing tests**:
  - `tests/integration/gate_packages/test_lifecycle.py` — module-level `_wheel_and_index_configured()` not directly unit-tested (only exercised indirectly via cleanup_acme).
  - `support/runtime_ops.py:cleanup_deployment` — best-effort error-path not unit-tested; OK for cleanup-only path.
  - `support/auth_ops.py:client_for_role` — token_store-sharing behavior (ENG-5955) has no test asserting persona isolation.
  - `echo-check/backend` — **no test exercises `KAMIWAZA_APP_PATH` env var + `TestClient(root_path=…)` together**, which is the actual production configuration per the Dockerfile (see High #2).

## Positive Observations
- **Strong response to prior review rounds.** Every actionable High and Medium finding from the previous two rounds was addressed inline (probe 401 acceptance, marker alignment, class-scope autouse, `uv run --extra test`, JWT comment, cleanup_acme env-gating and logging). Non-actionable items filed (ENG-5955) with clear rationale.
- **Test marker hygiene** — `pytestmark = [pytest.mark.live, pytest.mark.live_extension]` now correctly matches the standard `-m "not live"` exclusion.
- **Removed zero-signal CI step** — dropping the live-extension-contract step that was silently skipping was the right call.
- **TLS verify-off path never silently insecure** — explicit env opt-out + warning in both `test_lifecycle.py` and `settings.py`.
- **`workroom_trust.py`** identity/header trust model layered correctly: requires either trusted-routed context OR absence of forwarded identity headers. Security tests cover spoofing on direct access, cross-deployment root_path, and untrusted forwarded headers.
- **Safe shell argument handling** — `validated_bootstrap_arg`, `validated_username`, `validated_kubectl_resource_name`, `safe_token_file_path` all gate kubectl/kz-login invocations.
- **Module/class fixture-scope fix** in `test_lifecycle.py` — `cleanup_acme` no longer mutates the live cluster when `TestLifecycle` is going to skip anyway.
- **`EXTENSION_FIXTURES_ROOT`** anchored against `__file__` (not parent-counted from REPO_ROOT) — future relocations won't silently break path resolution.
- **Best-effort harness teardown** (`_best_effort_close`) properly logs cleanup failures instead of swallowing them.

## Overall Verdict
**Risk Level**: Medium

Rationale: No Critical issues; all dimension scores ≥7; Linear ticket references present (ENG-4770, ENG-5955, etc. in code and commit messages). The two High Priority issues — persona token-store sharing and the `--root-path`/prefix collision — are real and worth fixing, but are inherited from the source location rather than introduced by this relocation, so they don't block merge of the relocation itself. Recommend filing both as immediate follow-up tickets (ENG-5955 already exists for #1; #2 needs a new ticket) so the live extension-contract signal is trustworthy before any extension is shipped against it.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

jxstanford:
# RE-REVIEW Report: feat(tests+notebooks): receive extension_contract suite, gate-packages lifecycle test, jupyter quickstart from kamiwaza

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/129
**Branch**: `chore/eng-relocate-extension-contract-tests-and-jupyter` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | MERGEABLE |
| Merge state | CLEAN |
| Failing checks | None — all 9 checks SUCCESS (Claude Review, Build, Coverage, Ruff, Unit Tests, Mypy, Linear, Contract-live, Complexity) |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 9/10 | Matches stated scope (test-suite relocation). No production SDK code touched; jupyter launcher correctly deleted rather than hardened. |
| Security | 8/10 | TLS-verify defaults ON; kubectl/path-traversal/secret-name validation present; header-spoofing coverage comprehensive; `algorithm="none"` JWT helper correctly labeled test-only. ENG-5956 root-path collision is a latent fixture bug, not an SDK security regression. |
| Test Quality | 7/10 | 61 unit tests pass and marker registration verified. Lowered by 1 for two real fixture-isolation gaps Codex flagged: shared `FileTokenStore` between persona clients (can silently exercise the wrong user), and backend-unit step gated on live-cluster availability so it skips when secrets are absent. |
| Error Handling | 7/10 | try/except blocks documented with `noqa: BLE001`; cleanup paths log warnings rather than silently swallowing. Minor: `logger.warning` visibility in pytest without explicit `--log-cli-level`. |
| Code Quality | 8/10 | Clean module boundaries (`auth_ops`/`build_ops`/`runtime_ops`); rationale embedded where non-obvious; `EXTENSION_FIXTURES_ROOT` correctly anchored to its own location. |
| Refactoring | 7/10 | Appropriate restraint for a relocation PR; docker-build-context widening correctly deferred to ENG-5957. |

## Critical Issues (Must Fix)
| — | — | None | — | — |

## High Priority Issues
| — | — | None — previously-identified Highs filed as ENG-5955/5956/5957 follow-ups (consistent with prior reviewer's "don't block the relocation itself" call) | — | — |

| 1 | `.github/workflows/extension-contract-live.yml:86` | `status=$(curl ... -w "%{http_code}" ... \|\| echo "000")`: on a connection failure curl already writes `000` to stdout via `-w`, then exits non-zero and the `\|\| echo "000"` appends another `000`. Result is `000000`, which fails the `!= "000"` test and marks an unreachable cluster as "available". Downstream live steps then run and fail instead of skipping cleanly. | Codex | CI/Logic |
| 2 | `tests/e2e/extension_contract/support/auth_ops.py:47-54` (and `build_live_client` 94-101) | Persona clients constructed via `UserPasswordAuthenticator(...)` with no `token_store=` argument all share the default `FileTokenStore` at `~/.kamiwaza/token.json`. Once any persona (or the primary harness client) writes a token there, the next `_load_cached_token` call on a different persona reads it back and short-circuits the password grant — persona-scoped auth tests then exercise whatever user owns the cache, not `current_persona`. Pass an isolated `FileTokenStore(path=...)` per persona (or a null store) for harness clients. | Codex | Test Isolation |
| 3 | `.github/workflows/extension-contract-live.yml:107-108` | "Run echo-check backend unit tests" step is gated on `steps.probe.outputs.available == 'true'`. The root-level pytest already ignores `backend/` via `collect_ignore`, so this workflow step is the only place the backend suite runs; when `KAMIWAZA_BASE_URL` is missing or the cluster is unreachable, the backend unit tests (which don't need a live cluster) silently skip. Drop the conditional so the unit step always runs once `uv sync` succeeds. | Codex | CI Coverage |
| 4 | `echo-check/docker-compose.yml:6`, `docker-compose.appgarden.yml:3` | Image namespace `ghcr.io/kamiwaza-internal/kamiwaza/images/echo-check-app:0.2.0[-dev]` references a private internal registry from a public-facing SDK repo. External fork PRs will fail to pull. Acknowledged in commit body — deferred to coordinated image-publish-pipeline change. | Claude | Architecture |
| 5 | `tests/integration/gate_packages/test_lifecycle.py:64` | M5: switched `warnings.warn(...)` → `logger.warning(...)` for `KAMIWAZA_VERIFY_SSL=0`. Reasonable, but module-scope fixture `logger.warning` only surfaces in pytest output when the root logger admits WARNING (or `-o log_cli_level=WARNING` is set). Default pytest config may silently swallow it. | Claude | Code Quality |
| 6 | `tests/e2e/extension_contract/echo-check/backend/app/main.py:209-222` + `Dockerfile:37` | High-4 fix added `--root-path "${KAMIWAZA_APP_PATH:-}"` to uvicorn while `_register_routes` simultaneously calls `app.include_router(..., prefix=runtime_prefix)`. With both applied at runtime, Starlette strips `root_path` before matching, so prefixed routes 404 against routed requests. Acknowledged as ENG-5956; `test_runtime_prefix_routes_work` doesn't reproduce because `TestClient` is constructed without `root_path`. | Claude | Logic Correctness |
| 7 | `tests/e2e/extension_contract/support/state_loader.py:28-30` | `OSError` from `candidate.exists()` is aggregated into "malformed bucket" → `pytest.fail`. A `PermissionError` on a candidate path isn't a malformed-state condition; conflating could surface confusing diagnostics. | Claude | Error Handling |
| 8 | `tests/e2e/extension_contract/support/build_ops.py:196-197` | `(extension_root / context_rel).resolve()` and `(context / dockerfile_rel).resolve()` lack a root-containment check. Internal fixture only consumes the checked-in compose, low real-world risk, but a malicious compose could direct `docker build` outside the extension root. Defense-in-depth. | Claude | Security |
| 9 | `README.md:379` | Notebook quickstart now says install JupyterLab and `jupyter lab` from `examples/`. Several existing example notebooks hardcode `https://localhost/api`; README and notebook contents may diverge. Deferred to docs sweep. | Claude | Documentation |

## Test Coverage Assessment
- **Files with adequate coverage**: harness machinery (state, settings, build, runtime ops) — 61 passing tests; echo-check backend (auth, runtime, security helpers) — comprehensive negative-case coverage of header-spoofing and workroom-binding semantics.
- **Files needing tests**:
  - No test reproduces the ENG-5956 `--root-path` + app-prefix collision (would catch the production-routing 404). `test_runtime_prefix_routes_work` deliberately constructs `TestClient` without `root_path`, so it doesn't exercise the live-request shape.
  - `cleanup_acme` post-test path is best-effort by design; no test verifies the warning surfaces when uninstall fails.
  - Persona-client construction (Medium #2) lacks coverage that would have caught the shared `FileTokenStore` aliasing.

## Positive Observations
- Disciplined fixup history: each of 6 fixup commits carries a thorough rationale, Linear ticket trail, and verification (`61 passed`).
- TLS-verify default (`verify_ssl=True` opt-out) is now consistent between `test_lifecycle.py` and the extension_contract harness — the previous review's H1 inconsistency is fixed.
- Workflow `timeout-minutes: 30`, `--strict-markers` registration, `PyJWT>=2.8` moved to root dev group, `pytest.skip` instead of `pytest.fail` for missing bootstrap state — all real defects from previous review rounds were properly addressed.
- Auth-fronted `/ping` handling: probe accepts 401, harness accepts 401 with "Not authenticated" body — workflow probe and in-test ping-checker are aligned.
- Inline ENG-ticket references (ENG-5955/5956/5957) make the deferred-work trail self-documenting.
- `_wait_for_response` fail-fast on 401/403 avoids 180s timeouts on persistent auth misconfigurations.
- Class-scoped `_require_wheel_and_index` correctly gates the cluster-mutating `cleanup_acme` so it doesn't fire on half-skipped runs.

## Overall Verdict
**Risk Level**: Low
**Verdict**: APPROVE

Test-infrastructure relocation only — no production SDK code is touched. All 9 CI checks green, 61 harness tests passing, Linear ticket policy satisfied (ENG-5955/5956/5957 referenced in commit messages and inline). The author's fixup discipline closed every previous-review High in this PR or filed a tracked follow-up. Codex's three new findings (curl probe accumulation, persona token-cache aliasing, backend-unit step gated on live availability) are real but Medium-severity — fix on the next pass; none block the relocation.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

PR #130, opened 2026-06-01
ENG-5954: parse compose-spec long-form ports end-to-end

JohnStrick:
## What this ships

Teaches the kz-ext publish pipeline to handle compose-spec long-form port dicts and propagate `name` + `appProtocol` end-to-end to the K8s Service. Short-form behavior unchanged.

## Contract

Compose long-form is now valid in `docker-compose.yml`:

```yaml
ports:
  - target: 19530
    protocol: tcp
    name: grpc
    app_protocol: grpc
```

Translates to K8s Service:

```yaml
ports:
- name: grpc
  appProtocol: grpc
  port: 19530
  protocol: TCP
```

## Load-bearing decisions

- `serialize_by_alias=True` on `ExtensionPort` so default `model_dump()` emits camelCase. Without it, the K8s API server silently drops the field on `ExtensionService.create_extension` / `patch_extension`.
- `published`/`host_ip` are stripped in long-form by `_strip_host_ports` (closes a back-door host-port-binding vector that the validator alone couldn't catch).

## Verification

- 1405 extension tests pass (was 1402; +3 regression pins)
- Codex round 1 caught the `serialize_by_alias` gap; round 2 (after fixes) clean
- 3-agent local-review caught 3 more `str(port)`-on-dict sibling sites in `payload_builder.py` — all fixed in-PR
- End-to-end Python introspection through `ComposeTransformer → RegistryBuilder.build_entry → PayloadBuilder.build` confirmed the milvus example produces `name=grpc`/`appProtocol=grpc` on the wire

Part of ENG-5954 — companion PRs in `kamiwaza-internal/operators` and `kamiwaza-internal/kamiwaza-extensions-milvus`.

### DIFF
--- a/kamiwaza_extensions/compose_transformer.py
+++ b/kamiwaza_extensions/compose_transformer.py
@@ -576,10 +576,24 @@ def _rewrite_url_hosts(
     return new_value if new_value != value else None
 
 
-def _strip_host_ports(ports: List[Any]) -> List[str]:
-    """``"3000:3000"`` -> ``"3000"``; ``"8080:3000/tcp"`` -> ``"3000/tcp"``."""
-    result = []
+def _strip_host_ports(ports: List[Any]) -> List[Any]:
+    """``"3000:3000"`` -> ``"3000"``; ``"8080:3000/tcp"`` -> ``"3000/tcp"``.
+
+    Long-form (compose-spec dict) entries are preserved verbatim except
+    for ``published`` / ``host_ip``, which would otherwise re-introduce a
+    host port binding through the back door. ENG-5954: long-form is the
+    only way to carry L7 protocol hints (``name`` + ``app_protocol``)
+    through to the K8s Service.
+    """
+    result: List[Any] = []
     for port in ports:
+        if isinstance(port, dict):
+            cleaned = {
+                k: v for k, v in port.items() if k not in ("published", "host_ip")
+            }
+            if cleaned.get("target") is not None:
+                result.append(cleaned)
+            continue
         s = str(port)
         if ":" in s:
             # Take the part after the last colon (container port + optional protocol)
--- a/kamiwaza_extensions/payload_builder.py
+++ b/kamiwaza_extensions/payload_builder.py
@@ -409,8 +409,23 @@ class PayloadBuilder:
 
     @staticmethod
     def _parse_ports(ports: List[Any]) -> List[ExtensionPort]:
+        """Translate compose-spec ports into CR ExtensionPort entries.
+
+        Accepts both short-form (``"19530"``, ``"19530/udp"``) and long-form
+        compose-spec port entries (``{target, protocol, name, app_protocol}``).
+        For short-form the port name defaults to ``"http"`` — matches the
+        long-standing app/tool extension behavior. Long-form entries pass
+        ``name`` and ``app_protocol`` through to the K8s Service, which
+        istio reads for L7 protocol selection (ENG-5954).
+        """
         result = []
         for p in ports:
+            if isinstance(p, dict):
+                parsed = PayloadBuilder._parse_port_dict(p)
+                if parsed is not None:
+                    result.append(parsed)
+                continue
+
             s = str(p)
             # Strip protocol suffix if present
             proto = "TCP"
@@ -426,6 +441,48 @@ class PayloadBuilder:
             except (ValueError, TypeError):
                 continue
         return result
+
+    @staticmethod
+    def _parse_port_dict(port: Dict[str, Any]) -> Optional[ExtensionPort]:
+        """Translate one compose-spec long-form port entry."""
+        # Compose-spec long-form: ``target`` is the container port.
+        # ``published`` (host port) is stripped upstream by the transformer's
+        # ``_strip_host_ports`` and never reaches this function in the normal
+        # pipeline — we don't fall back to it because doing so would silently
+        # promote a host-port intent into a container port for any direct
+        # caller that bypasses the transformer.
+        target = port.get("target")
+        if target is None:
+            return None
+
+        proto_raw = str(port.get("protocol", "tcp")).upper()
+        proto = proto_raw if proto_raw in ("TCP", "UDP") else "TCP"
+
+        # Long-form ports without an explicit name still default to "http"
+        # so existing extensions that adopt long-form syntax for unrelated
+        # reasons (e.g. adding ``protocol: tcp``) don't silently lose the
+        # historical port name.
+        name = port.get("name") or "http"
+
+        # Prefer the compose-spec ``app_protocol`` key; fall back to the
+        # k8s-shaped ``appProtocol`` only when the spec key is absent.
+        # Explicit ``in`` rather than ``or`` so an explicitly-empty
+        # ``app_protocol: ""`` is treated as "set to empty" rather than
+        # silently falling through to ``appProtocol``.
+        if "app_protocol" in port:
+            app_protocol = port["app_protocol"]
+        else:
+            app_protocol = port.get("appProtocol")
+
+        try:
+            return ExtensionPort(
+                container_port=int(target),
+                protocol=proto,
+                name=name,
+                app_protocol=app_protocol,
+            )
+        except (ValueError, TypeError):
+            return None
 
     @staticmethod
     def _parse_env(env: Any) -> List[Dict[str, Any]]:
@@ -666,11 +723,20 @@ def _should_use_node_frontend_probe(svc_name: str, svc: Dict[str, Any]) -> bool:
 
     ports = svc.get("ports", [])
     for port in ports:
-        port_str = str(port).split("/", 1)[0]
-        try:
-            container_port = int(port_str.rsplit(":", 1)[-1])
-        except ValueError:
-            continue
+        if isinstance(port, dict):
+            target = port.get("target")
+            if target is None:
+                continue
+            try:
+                container_port = int(target)
+            except (ValueError, TypeError):
+                continue
+        else:
+            port_str = str(port).split("/", 1)[0]
+            try:
+                container_port = int(port_str.rsplit(":", 1)[-1])
+            except ValueError:
+                continue
         if container_port in {3000, 3001, 4173, 5173}:
             return True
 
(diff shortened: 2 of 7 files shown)


kbot-ships (approved):
_Reviewing head `14c6d41` against `develop`._

## Bottom line
This PR teaches the publish pipeline to accept compose-spec long-form port dicts and carry `name` + `app_protocol` through to the K8s `ExtensionPort`, while keeping short-form behavior intact. The core publish path — `_strip_host_ports` → `_parse_ports`/`_parse_port_dict` → `ExtensionPort` serialization → validator host-port detection → frontend node-probe heuristic — is implemented correctly and backed by genuinely thorough tests, including the subtle `serialize_by_alias` regression that would otherwise silently drop `appProtocol`. The direction is right and the implemented scope is solid. My only reservation is that the "end-to-end" claim overreaches slightly: sibling consumers of `svc["ports"]` outside the publish path still use the old `str(port)` colon heuristic and misbehave on long-form dicts. None are crashes or correctness bugs in the shipped CR path, so this is APPROVE with notes.

## Findings

### 1. `app_analyzer._check_deployment_compat` false-positives every long-form port as a host binding — **(worth fixing)**
`kamiwaza_extensions/app_analyzer.py:453-456` does `port_str = str(port_spec); if ":" in port_str`. For a long-form dict, `str({"target": 19530, "name": "grpc"})` is `"{'target': 19530, 'name': 'grpc'}"`, which always contains a colon, so *every* long-form port — including the standard target-only case with no `published` — gets reported in `has_host_ports`. This directly contradicts the validator fix in `validators/compose.py:104-112`, which you correctly updated to only warn when `published` is set. A user authoring the exact `milvus`-style port the PR is designed to support would get a spurious "host port" signal from the analyzer. Same colon-in-`str(dict)` bug class you fixed in three other places; mirror the dict-handling here.

### 2. `app_analyzer` port extraction silently drops long-form ports — **(nit)**
`kamiwaza_extensions/app_analyzer.py:381-388`: `str(port_spec).split(":")[-1]` → `int(...)` raises `ValueError` for a dict and is swallowed, so `info.ports` never records long-form targets. Degraded analysis, not a crash.

### 3. `sdk_override/classification.py` frontend detection blind to long-form — **(worth fixing)**
`detect_service_type` at `classification.py:44-47` checks `":3000" in str(port_spec)`. Python's dict repr renders `: 3000` (with a space), so a frontend service authored with long-form `{"target": 3000}` is classified `backend` and won't get the frontend SDK overlay. You fixed exactly this heuristic in `_should_use_node_frontend_probe` (`payload_builder.py:723-738`) but left this twin untouched — the two now disagree on the same input.

### 4. `dev_local.py` URL printing degrades on long-form — **(nit)**
`dev_local.py:482,519` pass `str(port_spec)` to `parse_port_mapping`. Long-form dicts won't parse, so the local-dev URL hints are lost (docker compose itself still runs fine). Outside the publish path; low priority.

### 5. `pydantic` lower bound not pinned for `serialize_by_alias` — **(worth fixing)**
`serialize_by_alias` in `ConfigDict` (`kamiwaza_sdk/schemas/extensions.py:18-20`) requires pydantic ≥ 2.11; older versions silently ignore the unknown config key and the nested `model_dump()` would emit snake-case `app_protocol`, reintroducing the exact field-drop this PR guards against. `uv.lock` pins 2.12.5 so the locked env is safe, but `pyproject.toml:25` only declares `pydantic[email]` with no floor. Add `pydantic>=2.11` so a non-locked install can't regress below the feature's requirement.

## Test gaps
Coverage for the implemented publish path is strong and reverting any piece of the fix would fail a test (`TestParsePorts`, `test_long_form_port_preserved_with_l7_hints`, `test_long_form_port_strips_published_and_host_ip`, the validator warn/no-warn pair, `test_frontend_long_form_port_3000_uses_node_probe`, and the two `serialize_by_alias` regression tests). The gap is the untested siblings above: no test asserts `app_analyzer`/`classification` behavior on long-form dicts, which is why findings 1–3 slipped through. If you fix any of them, add a case pinning long-form behavior so they don't drift back.

## What's good
- The `serialize_by_alias` + `populate_by_name` combination is the correct fix, and `test_default_dump_emits_appprotocol_for_nested_payload` pins the real `create_extension`/`patch_extension` dump path rather than the easy `by_alias=True` case — that's the failure mode that would actually bite in production.
- Refusing to fall back from `target` to `published` in `_parse_port_dict`, and the explicit `"app_protocol" in port` check (rather than `or`) so an empty-string value is honored, are both well-reasoned and documented.
- Comments explain *why*, not just *what*, throughout.

## Verdict
**Verdict**: APPROVE

The shipped publish→CR→Service path is correct and well-tested; the remaining items are out-of-scope sibling heuristics and a dependency-floor hardening, none of which are correctness bugs in the path this PR delivers.
---

jxstanford (approved):
# Review Report: ENG-5954: parse compose-spec long-form ports end-to-end

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/130
**Branch**: `feature/ENG-5954` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 8/10 | Publish pipeline implemented end-to-end and consistently. Minor gap between the broad "long-form is now valid" contract claim and un-updated analysis/dev_local consumers, which still `str(port_dict)`. |
| Security | 9/10 | No injection/secrets/auth surface; host-port intent (`published`/`host_ip`) stripped on both short and long forms; the privileged-port validator (`_parse_service_ports`) already handles dicts. Only gap is unvalidated K8s port `name` (fails safe downstream). |
| Test Quality | 9/10 | Thorough: short/long form, name fallback, UDP, camel/snake alias precedence, empty-string app_protocol, mixed forms, missing/invalid target, default-dump regression, node-probe heuristic. Missing only the explicit `app_protocol: null` edge. |
| Error Handling | 9/10 | `int()` conversions guarded, missing/invalid targets skipped cleanly, no silent crashes on malformed dicts. |
| Code Quality | 9/10 | Clear naming, well-factored `_parse_port_dict`, conventions matched; intent-rich comments are exemplary. |
| Refactoring | 7/10 | Dict-vs-string port parsing is now duplicated across `_strip_host_ports`, `_parse_ports`, `_should_use_node_frontend_probe`, `compose.py`, and `platform_runtime.py`; a shared `extract_container_port(port)` helper would DRY this and prevent the noted consumer drift. |

## Critical Issues (Must Fix)
| — | — | None | — | — |

## High Priority Issues
| — | — | None | — | — |

| 1 | `kamiwaza_sdk/schemas/extensions.py:19` / `pyproject.toml:25` | Feature depends on the `serialize_by_alias` ConfigDict option (Pydantic ≥ 2.11), but the dependency is declared as `pydantic[email]` with no lower bound. On an env constrained to pydantic 2.0–2.10 the option is silently ignored, `CreateExtension.model_dump()` emits `app_protocol`, and the K8s API server drops the field — the exact failure the code comment warns about. Pin `pydantic>=2.11`, or serialize extension payloads with `by_alias=True` explicitly. (Note: the in-repo `uv.lock` resolves 2.12.5, so local/CI is unaffected; risk is to external consumers.) | Claude Code, Codex | Components w/ known constraints |
| 2 | `payload_builder.py:465` / `validators/compose.py:104` | Long-form `name` flows through unvalidated. K8s Service port names must be DNS-1123 labels (≤15 chars). Short-form always used the valid `"http"`; long-form now lets arbitrary names through, so an invalid name fails at K8s apply time with a confusing downstream error rather than at publish/validate. Consider validating port `name` in the compose validator. | Claude Code (single-source) | Data validation |
| 3 | `app_analyzer.py:381,453`, `sdk_override/classification.py:44` | PR contract states "long-form is now valid in docker-compose.yml," but only the publish pipeline was updated. Other compose consumers still `str(port_dict)` and either skip the port (analysis under-counts) or false-match (`app_analyzer` reports a dict repr as a host port because it contains `": "`). No crash, but inconsistent with the stated contract. The security-relevant `platform_runtime._parse_service_ports` already handles dicts, so the security path is fine. | Claude Code (single-source) | Pattern consistency / scope |

## Test Coverage Assessment
- **Verdict: Strong.** In-scope pipeline is well-covered. `TestParsePorts` exercises short/long form, name fallback, UDP, camel/snake alias precedence, empty-string `app_protocol`, mixed forms, and missing/invalid target. `test_default_dump_emits_appprotocol_for_nested_payload` is exactly right — it pins the parent-level `model_dump()` shape the real `create_extension` produces, guarding against silent field-drop. Transformer tests confirm long-form L7 hints survive and `published`/`host_ip` are stripped. Validator tests cover target-only (no warn) and `published` (warns).
- **Gaps**: Long-form `app_protocol:` explicitly null (key present, value `None`) on the create path (`exclude_none` not applied) — low impact. The un-updated consumers in Medium #3 lack tests for dict-form ports.

## Positive Observations
- Excellent, intent-rich comments explaining *why* (the `in`-vs-`or` app_protocol precedence, the no-fallback-to-`published` security rationale, the `serialize_by_alias` necessity).
- `_strip_host_ports` correctly strips `published`/`host_ip` from long-form dicts, closing the host-port back door consistently with short-form.
- Short-form behavior is genuinely unchanged; long-form is purely additive.
- The serialization regression test pins the real create-path dump shape rather than the easier `by_alias=True` path, which is the failure mode that actually matters.

## Overall Verdict
**Risk Level**: Low
**Rationale**: A well-scoped, well-tested additive change that threads `name` + `appProtocol` through the publish pipeline end-to-end with strong regression coverage and no behavior change for short-form. No Critical issues, all dimensions ≥ 7, Linear ticket (ENG-5954) present. Both engines independently converged on the same single substantive follow-up — a defensive pydantic version pin — which does not block merge given the lockfile already resolves 2.12.5. The port-name validation and consumer-consistency items are reasonable follow-ups.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

PR #150, opened 2026-06-09
ENG-6538 Align Context no-workroom live SDK tests

kbot-ships:
## Summary
- align live Context SDK expectations with the no-workroom/global contract from kamiwaza PR #1893
- replace the stale Global Workroom read-only VectorDB assertion with a no-workroom lifecycle assertion using the shared fixture
- keep global export/ingestion checks focused on raw endpoint behavior so live smoke validates the server contract without extra SDK model coupling

## Verification
- python3 -m py_compile tests/integration/test_context_live.py tests/integration/test_workroom_isolation_live.py
- uv run pytest tests/integration/test_context_live.py tests/integration/test_workroom_isolation_live.py --collect-only -q

## Notes
- Live validation is expected via /kajiya smoke on kamiwaza PR #1893 with this SDK branch override.

PR #156, opened 2026-06-11
ENG-6538 Align Context no-workroom live SDK tests

kbot-ships:
## Summary

Align live SDK Context tests with the ENG-6538 no-workroom/workroom-scope contract exercised by kamiwaza PR #1893.

- Treat VectorDB create without an explicit non-Global workroom as `workroom_scope_required` instead of the stale Global-read-only contract.
- Bind the ephemeral workroom on the same SDK session used for workroom-scoped Context live tests.
- Run vector insert/query coverage against a workroom-scoped VectorDB while preserving both endpoint shapes.

## Related

- Linear: ENG-6538
- Product PR: https://github.com/kamiwaza-internal/kamiwaza/pull/1893
- Supersedes stale SDK branch/PR #150; this branch is rebuilt from current `origin/develop`.

## Verification

- `python -m py_compile tests/integration/test_context_live.py`
- `uv run ruff check tests/integration/test_context_live.py`
- `uv run pytest tests/integration/test_context_live.py tests/integration/test_workroom_isolation_live.py --collect-only -q`
- `uv run pytest tests/unit/test_context_service.py tests/unit/test_workroom_service.py --no-cov -q`

Live verification is being run from product PR #1893 via `/kajiya smoke` with `sdk: fix/eng-6538-context-no-workroom-sdk-tests-develop`.

### DIFF
--- a/kamiwaza_sdk/schemas/workrooms.py
+++ b/kamiwaza_sdk/schemas/workrooms.py
@@ -86,6 +86,24 @@ class DeleteWorkroomResponse(WorkroomResponseModel):
     message: str
 
 
+class EnterWorkroomResponse(WorkroomResponseModel):
+    """Response from entering a workroom session."""
+
+    workroom_id: UUID
+    access_token: Optional[str] = None
+    expires_in: Optional[int] = None
+    message: str = "Workroom session bound"
+
+
+class LeaveWorkroomResponse(WorkroomResponseModel):
+    """Response from leaving a workroom session."""
+
+    workroom_id: UUID
+    access_token: Optional[str] = None
+    expires_in: Optional[int] = None
+    message: str = "Returned to Global Workroom"
+
+
 class ExportManifestItem(WorkroomResponseModel):
     """A single item in the export manifest."""
 
--- a/kamiwaza_sdk/schemas/workrooms.py
+++ b/kamiwaza_sdk/schemas/workrooms.py
@@ -87,7 +87,7 @@ class DeleteWorkroomResponse(WorkroomResponseModel):
 
 
 class EnterWorkroomResponse(WorkroomResponseModel):
-    """Response from entering a workroom session."""
+    """Response from the backend workroom-enter endpoint."""
 
     workroom_id: UUID
     access_token: Optional[str] = None
@@ -96,7 +96,7 @@ class EnterWorkroomResponse(WorkroomResponseModel):
 
 
 class LeaveWorkroomResponse(WorkroomResponseModel):
-    """Response from leaving a workroom session."""
+    """Response from the backend workroom-leave endpoint."""
 
     workroom_id: UUID
     access_token: Optional[str] = None
--- a/kamiwaza_sdk/services/workrooms.py
+++ b/kamiwaza_sdk/services/workrooms.py
@@ -7,8 +7,10 @@ from ..exceptions import APIError, NotFoundError
 from ..schemas.workrooms import (
     CreateWorkroom,
     DeleteWorkroomResponse,
+    EnterWorkroomResponse,
     ExportManifest,
     IngestionSummary,
+    LeaveWorkroomResponse,
     UpdateWorkroom,
     Workroom,
 )
@@ -218,6 +220,26 @@ class WorkroomService(BaseService):
             if e.status_code == 404:
                 raise NotFoundError(f"Workroom {wid} not found")
             raise
+
+    # -------------------------------------------------------------------------
+    # Session binding
+    # -------------------------------------------------------------------------
+
+    def enter(self, workroom_id: Union[str, UUID]) -> EnterWorkroomResponse:
+        """Bind the current SDK session to a workroom."""
+        wid = self._ensure_uuid(workroom_id)
+        try:
+            response = self.client.post(f"/workrooms/{wid}/enter")
+            return EnterWorkroomResponse.model_validate(response)
+        except APIError as e:
+            if e.status_code == 404:
+                raise NotFoundError(f"Workroom {wid} not found")
+            raise
+
+    def leave(self) -> LeaveWorkroomResponse:
+        """Return the current SDK session to the Global Workroom."""
+        response = self.client.post("/workrooms/leave")
+        return LeaveWorkroomResponse.model_validate(response)
 
     # -------------------------------------------------------------------------
     # Export & ingestion
--- a/kamiwaza_sdk/services/workrooms.py
+++ b/kamiwaza_sdk/services/workrooms.py
@@ -64,14 +64,16 @@ class WorkroomService(BaseService):
         Raises:
             APIError: If creation fails (e.g. limit exceeded -> 409).
         """
-        payload = CreateWorkroom(
-            name=name,
-            type=workroom_type,
-            description=description,
-            labels=labels,
-            classification=classification,
-            attributes=attributes,
-            scg_references=scg_references,
+        payload = CreateWorkroom.model_validate(
+            {
+                "name": name,
+                "type": workroom_type,
+                "description": description,
+                "labels": labels,
+                "classification": classification,
+                "attributes": attributes,
+                "scg_references": scg_references,
+            }
         )
         response = self.client.post(
             "/workrooms/",
@@ -145,8 +147,8 @@ class WorkroomService(BaseService):
             APIError: If validation fails (400) or workroom is read-only (409).
         """
         wid = self._ensure_uuid(workroom_id)
-        payload = UpdateWorkroom(
-            **self._provided_fields(
+        payload = UpdateWorkroom.model_validate(
+            self._provided_fields(
                 name=name,
                 description=description,
                 labels=labels,
--- a/kamiwaza_sdk/services/workrooms.py
+++ b/kamiwaza_sdk/services/workrooms.py
@@ -228,7 +228,12 @@ class WorkroomService(BaseService):
     # -------------------------------------------------------------------------
 
     def enter(self, workroom_id: Union[str, UUID]) -> EnterWorkroomResponse:
-        """Bind the current SDK session to a workroom."""
+        """Call the backend workroom-enter endpoint and return its response.
+
+        The SDK does not install returned access tokens into the configured
+        authenticator. Callers that need scoped Context access should continue
+        to pass explicit ``workroom_id`` values where the SDK exposes them.
+        """
         wid = self._ensure_uuid(workroom_id)
         try:
             response = self.client.post(f"/workrooms/{wid}/enter")
@@ -239,7 +244,11 @@ class WorkroomService(BaseService):
             raise
 
     def leave(self) -> LeaveWorkroomResponse:
-        """Return the current SDK session to the Global Workroom."""
+        """Call the backend workroom-leave endpoint and return its response.
+
+        Returned token fields are exposed for callers that manage tokens
+        themselves; this method does not mutate the SDK authenticator.
+        """
         response = self.client.post("/workrooms/leave")
         return LeaveWorkroomResponse.model_validate(response)
 
--- a/tests/integration/test_context_live.py
+++ b/tests/integration/test_context_live.py
@@ -302,16 +302,9 @@ def shared_vectordb(shared_context_service: ContextService) -> str:
 def session_workroom(shared_context_service: ContextService) -> str:
     """Per-session writable workroom for Context Service write-path tests.
 
-    The Global Workroom (``DEFAULT_WORKROOM_ID``) is read-only for context
-    writes: ontology / pipeline / collection / object-storage creation against
-    it return ``403 "Global Workroom is read-only for ..."``. Write-path tests
-    therefore need a real, writable workroom.
-
-    The id is supplied to the SDK via ``workroom_id=`` (the ``X-Workroom-ID``
-    header). The Context Service authorizes that header against the caller's
-    verified workroom membership, and the istio ingress preserves it
-    end-to-end, so writes land in -- and stay isolated to -- this workroom.
-    Read-only assertions against Global continue to use ``DEFAULT_WORKROOM_ID``.
+    Room-scoped Context routes require an explicit non-Global workroom scope.
+    Bind that workroom on the same SDK session that provisions shared test
+    resources so follow-up calls carry verified workroom authority.
     """
     workrooms = shared_context_service.client.workrooms
     workroom = workrooms.create(
@@ -320,6 +313,7 @@ def session_workroom(shared_context_service: ContextService) -> str:
         description="Ephemeral workroom for SDK context live tests",
     )
     workroom_id = str(workroom.id)
+    shared_context_service.client.post(f"/workrooms/{workroom_id}/enter")
     try:
         yield workroom_id
     finally:
@@ -382,19 +376,10 @@ def test_context_required_llm_available(context_required_llm: str) -> None:
     assert context_required_llm
 
 
-def test_context_vectordb_create_in_global_workroom_is_read_only(
+def test_context_vectordb_create_without_workroom_requires_scope(
     live_kamiwaza_client,
 ) -> None:
-    """Global Workroom is read-only for VectorDB creation (ENG-4352, PR #1635).
-
-    Verifies the server-side gate at
-    ``kamiwaza/services/context/lifecycle.py:_raise_if_global_workroom_write``
-    rejects ``create_vectordb`` without an explicit workroom_id (which
-    defaults to the Global Workroom UUID). Returns 403 regardless of
-    caller role, matching commit 73071d5d1's stated intent:
-    "Keeps Global Workroom read paths available to members and blocks
-    Global write paths in both ReBAC-on and RBAC-off modes."
-    """
+    """VectorDB creation is room-scoped and requires a non-Global workroom."""
     service = _context_service(live_kamiwaza_client)
 
     with pytest.raises(APIError) as exc_info:
@@ -403,8 +388,8 @@ def test_context_vectordb_create_in_global_workroom_is_read_only(
             engine="milvus",
         )
 
-    assert exc_info.value.status_code == 403
-    assert "Global Workroom is read-only" in str(exc_info.value)
+    assert exc_info.value.status_code == 400
+    assert "workroom_scope_required" in str(exc_info.value)
 
 
 # A dedicated non-Global VectorDB *instance* lifecycle test (create/scale/
@@ -420,11 +405,12 @@ def test_context_vectordb_create_in_global_workroom_is_read_only(
 
 
 def test_context_vectordb_insert_vectors_instance(
-    live_kamiwaza_client,
-    shared_vectordb: str,
+    shared_context_service: ContextService,
+    session_workroom: str,
+    shared_workroom_vectordb: str,
 ) -> None:
-    service = _context_service(live_kamiwaza_client)
-    vectordb_id = shared_vectordb
+    service = shared_context_service
+    vectordb_id = shared_workroom_vectordb
     collection_name = _sdk_collection_name()
 
     inserted = service.insert_vectors(
@@ -432,16 +418,18 @@ def test_context_vectordb_insert_vectors_instance(
         collection_name=collection_name,
         vectors=[_sample_vector()],
         metadata=[{"source": "sdk-context-live"}],
+        workroom_id=session_workroom,
     )
     assert inserted["inserted_count"] == 1
 
 
 def test_context_vectordb_insert_vectors_global(
-    live_kamiwaza_client,
-    shared_vectordb: str,
+    shared_context_service: ContextService,
+    session_workroom: str,
+    shared_workroom_vectordb: str,
 ) -> None:
-    service = _context_service(live_kamiwaza_client)
-    vectordb_id = shared_vectordb
+    service = shared_context_service
+    vectordb_id = shared_workroom_vectordb
     collection_name = _sdk_collection_name()
 
     inserted = service.insert_vectors_global(
@@ -449,16 +437,18 @@ def test_context_vectordb_insert_vectors_global(
         collection_name=collection_name,
         vectors=[_sample_vector()],
         metadata=[{"source": "sdk-context-live"}],
+        workroom_id=session_workroom,
     )
     assert inserted["inserted_count"] == 1
 
 
 def test_context_vectordb_query_vectors_instance(
-    live_kamiwaza_client,
-    shared_vectordb: str,
+    shared_context_service: ContextService,
+    session_workroom: str,
+    shared_workroom_vectordb: str,
 ) -> None:
-    service = _context_service(live_kamiwaza_client)
-    vectordb_id = shared_vectordb
+    service = shared_context_service
+    vectordb_id = shared_workroom_vectordb
     collection_name = _sdk_collection_name()
 
     service.insert_vectors(
@@ -466,22 +456,25 @@ def test_context_vectordb_query_vectors_instance(
         collection_name=collection_name,
         vectors=[_sample_vector()],
         metadata=[{"source": "sdk-context-live"}],
+        workroom_id=session_workroom,
     )
     queried = service.query_vectors(
         vectordb_id,
         collection_name=collection_name,
         vectors=[_sample_vector()],
         limit=1,
+        workroom_id=session_workroom,
     )
     assert isinstance(queried["results"], list)
 
 
 def test_context_vectordb_query_vectors_global(
-    live_kamiwaza_client,
-    shared_vectordb: str,
+    shared_context_service: ContextService,
+    session_workroom: str,
+    shared_workroom_vectordb: str,
 ) -> None:
-    service = _context_service(live_kamiwaza_client)
-    vectordb_id = shared_vectordb
+    service = shared_context_service
+    vectordb_id = shared_workroom_vectordb
     collection_name = _sdk_collection_name()
 
     service.insert_vectors_global(
@@ -489,12 +482,14 @@ def test_context_vectordb_query_vectors_global(
         collection_name=collection_name,
         vectors=[_sample_vector()],
         metadata=[{"source": "sdk-context-live"}],
+        workroom_id=session_workroom,
     )
     queried = service.query_vectors_global(
         vectordb_id=vectordb_id,
         collection_name=collection_name,
         vectors=[_sample_vector()],
         limit=1,
+        workroom_id=session_workroom,
     )
     assert isinstance(queried["results"], list)
 
@@ -643,10 +638,10 @@ def test_context_ontology_delete_group(
 
 @pytest.mark.requires_embedding_model
 def test_context_workroom_lists_and_job_creation(
-    live_kamiwaza_client,
+    shared_context_service: ContextService,
     session_workroom: str,
 ) -> None:
-    service = _context_service(live_kamiwaza_client)
+    service = shared_context_service
     workroom_id = session_workroom
     created_job_ids: list[str] = []
 
@@ -701,10 +696,10 @@ def test_context_workroom_lists_and_job_creation(
 
 @pytest.mark.requires_embedding_model
 def test_context_workroom_pipeline_followup_access(
-    live_kamiwaza_client,
+    shared_context_service: ContextService,
     session_workroom: str,
 ) -> None:
-    service = _context_service(live_kamiwaza_client)
+    service = shared_context_service
     workroom_id = session_workroom
 
     payload = base64.b64encode(b"hello context service").decode("utf-8")
@@ -733,11 +728,11 @@ def test_context_workroom_pipeline_followup_access(
 
 
 def test_context_workroom_collection_lifecycle(
-    live_kamiwaza_client,
+    shared_context_service: ContextService,
     session_workroom: str,
     shared_workroom_vectordb: str,
 ) -> None:
-    service = _context_service(live_kamiwaza_client)
+    service = shared_context_service
     workroom_id = session_workroom
     assert shared_workroom_vectordb
     collection_name = _sdk_collection_name()
@@ -783,11 +778,11 @@ def test_context_workroom_collection_lifecycle(
 
 @pytest.mark.requires_embedding_model
 def test_context_search_contract(
-    live_kamiwaza_client,
+    shared_context_service: ContextService,
     session_workroom: str,
     shared_workroom_vectordb: str,
 ) -> None:
-    service = _context_service(live_kamiwaza_client)
+    service = shared_context_service
     workroom_id = session_workroom
     assert shared_workroom_vectordb
 
@@ -801,11 +796,11 @@ def test_context_search_contract(
 
 @pytest.mark.requires_embedding_model
 def test_context_retrieve_contract(
-    live_kamiwaza_client,
+    shared_context_service: ContextService,
     session_workroom: str,
     shared_workroom_vectordb: str,
 ) -> None:
-    service = _context_service(live_kamiwaza_client)
+    service = shared_context_service
     workroom_id = session_workroom
     assert shared_workroom_vectordb
 
--- a/tests/integration/test_context_live.py
+++ b/tests/integration/test_context_live.py
@@ -260,6 +260,26 @@ def _is_stale_sdk_resource(resource: dict, max_age: timedelta) -> bool:
 _STALE_THRESHOLD = timedelta(minutes=15)
 
 
+def _api_error_code(error: APIError) -> str | None:
+    """Extract the stable server error code from an APIError payload."""
+    payload = error.response_data
+    if not isinstance(payload, dict):
+        return None
+
+    detail = payload.get("detail")
+    if isinstance(detail, dict):
+        for key in ("error", "code", "reason"):
+            value = detail.get(key)
+            if isinstance(value, str):
+                return value
+
+    for key in ("error", "code", "reason"):
+        value = payload.get(key)
+        if isinstance(value, str):
+            return value
+    return None
+
+
 @pytest.fixture(scope="session", autouse=True)
 def _cleanup_stale_sdk_vdbs(shared_context_service: ContextService) -> None:
     """Delete leftover sdk-* VDB/ontology instances from prior crashed runs.
@@ -288,17 +308,6 @@ def _cleanup_stale_sdk_vdbs(shared_context_service: ContextService) -> None:
 
 
 @pytest.fixture(scope="session")
-def shared_vectordb(shared_context_service: ContextService) -> str:
-    """Shared global VectorDB instance for non-destructive vector tests."""
-    service = shared_context_service
-    vectordb_id = _create_temp_vectordb(service, prefix="sdk-shared-vdb")
-    try:
-        yield vectordb_id
-    finally:
-        _safe_delete_vectordb(service, vectordb_id)
-
-
-@pytest.fixture(scope="session")
 def session_workroom(shared_context_service: ContextService) -> str:
     """Per-session writable workroom for Context Service write-path tests.
 
@@ -313,10 +322,15 @@ def session_workroom(shared_context_service: ContextService) -> str:
         description="Ephemeral workroom for SDK context live tests",
     )
     workroom_id = str(workroom.id)
-    shared_context_service.client.post(f"/workrooms/{workroom_id}/enter")
     try:
+        entered = workrooms.enter(workroom_id)
+        assert str(entered.workroom_id) == workroom_id
         yield workroom_id
     finally:
+        try:
+            workrooms.leave()
+        except APIError:
+            pass
         # delete() raises NotFoundError (a sibling of APIError, not a subclass)
         # when the workroom is already gone, so catch both to keep teardown
         # best-effort -- matching the sibling test_workroom_isolation_live.py.
@@ -389,7 +403,7 @@ def test_context_vectordb_create_without_workroom_requires_scope(
         )
 
     assert exc_info.value.status_code == 400
-    assert "workroom_scope_required" in str(exc_info.value)
+    assert _api_error_code(exc_info.value) == "workroom_scope_required"
 
 
 # A dedicated non-Global VectorDB *instance* lifecycle test (create/scale/
--- a/tests/integration/test_context_live.py
+++ b/tests/integration/test_context_live.py
@@ -5,10 +5,12 @@ from __future__ import annotations
 import base64
 import os
 import time
+from collections.abc import Generator
 from datetime import datetime, timedelta, timezone
 from uuid import uuid4
 
 import pytest
+from pydantic import ValidationError
 
 from kamiwaza_sdk import KamiwazaClient
 from kamiwaza_sdk.authentication import UserPasswordAuthenticator
@@ -308,12 +310,15 @@ def _cleanup_stale_sdk_vdbs(shared_context_service: ContextService) -> None:
 
 
 @pytest.fixture(scope="session")
-def session_workroom(shared_context_service: ContextService) -> str:
+def session_workroom(
+    shared_context_service: ContextService,
+) -> Generator[str, None, None]:
     """Per-session writable workroom for Context Service write-path tests.
 
     Room-scoped Context routes require an explicit non-Global workroom scope.
-    Bind that workroom on the same SDK session that provisions shared test
-    resources so follow-up calls carry verified workroom authority.
+    Exercise the backend enter endpoint as a checked lifecycle seam, while
+    Context calls below pass explicit workroom_id so authority is not inferred
+    from SDK-local session state.
     """
     workrooms = shared_context_service.client.workrooms
     workroom = workrooms.create(
@@ -329,7 +334,7 @@ def session_workroom(shared_context_service: ContextService) -> str:
     finally:
         try:
             workrooms.leave()
-        except APIError:
+        except (APIError, ValidationError):
             pass
         # delete() raises NotFoundError (a sibling of APIError, not a subclass)
         # when the workroom is already gone, so catch both to keep teardown
@@ -344,7 +349,7 @@ def session_workroom(shared_context_service: ContextService) -> str:
 def shared_workroom_vectordb(
     shared_context_service: ContextService,
     session_workroom: str,
-) -> str:
+) -> Generator[str, None, None]:
     """Shared workroom-scoped VectorDB for collection/search/retrieve tests."""
     service = shared_context_service
     vectordb_id = _create_temp_vectordb(
@@ -366,7 +371,7 @@ def shared_workroom_vectordb(
 def shared_ontology(
     shared_context_service: ContextService,
     context_required_llm: str,
-) -> str:
+) -> Generator[str, None, None]:
     """Shared ontology instance for non-destructive ontology tests."""
     assert context_required_llm
     service = shared_context_service
--- a/tests/unit/test_workroom_service.py
+++ b/tests/unit/test_workroom_service.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import io
 import uuid
+from types import SimpleNamespace
 
 import pytest
 from pydantic import ValidationError
@@ -506,6 +507,34 @@ def test_leave_posts_to_leave_endpoint(dummy_client):
     assert str(result.workroom_id) == "ffffffff-ffff-ffff-ffff-ffffffffffff"
     assert result.access_token == "global-token"
     assert result.expires_in == 3600
+
+
+def test_enter_leave_expose_tokens_without_mutating_client_auth(dummy_client):
+    responses = {
+        ("post", f"/workrooms/{WORKROOM_UUID}/enter"): {
+            "workroom_id": WORKROOM_ID,
+            "access_token": "scoped-token",
+            "expires_in": 3600,
+            "message": "Workroom session bound",
+        },
+        ("post", "/workrooms/leave"): {
+            "workroom_id": "ffffffff-ffff-ffff-ffff-ffffffffffff",
+            "access_token": "global-token",
+            "expires_in": 3600,
+            "message": "Returned to Global Workroom",
+        },
+    }
+    client = dummy_client(responses)
+    session = SimpleNamespace(headers={"Authorization": "Bearer original"})
+    setattr(client, "session", session)
+    service = WorkroomService(client)
+
+    entered = service.enter(WORKROOM_ID)
+    left = service.leave()
+
+    assert entered.access_token == "scoped-token"
+    assert left.access_token == "global-token"
+    assert session.headers["Authorization"] == "Bearer original"
 
 
 # =============================================================================
--- a/tests/unit/test_workroom_service.py
+++ b/tests/unit/test_workroom_service.py
@@ -464,6 +464,51 @@ def test_archive_not_found_raises(dummy_client):
 
 
 # =============================================================================
+# Session Binding
+# =============================================================================
+
+
+def test_enter_posts_to_enter_endpoint(dummy_client):
+    responses = {
+        ("post", f"/workrooms/{WORKROOM_UUID}/enter"): {
+            "workroom_id": WORKROOM_ID,
+            "access_token": "scoped-token",
+            "expires_in": 3600,
+            "message": "Workroom session bound",
+        }
+    }
+    client = dummy_client(responses)
+    service = WorkroomService(client)
+
+    result = service.enter(WORKROOM_ID)
+
+    assert client.calls[0][1] == f"/workrooms/{WORKROOM_UUID}/enter"
+    assert result.workroom_id == WORKROOM_UUID
+    assert result.access_token == "scoped-token"
+    assert result.expires_in == 3600
+
+
+def test_leave_posts_to_leave_endpoint(dummy_client):
+    responses = {
+        ("post", "/workrooms/leave"): {
+            "workroom_id": "ffffffff-ffff-ffff-ffff-ffffffffffff",
+            "access_token": "global-token",
+            "expires_in": 3600,
+            "message": "Returned to Global Workroom",
+        }
+    }
+    client = dummy_client(responses)
+    service = WorkroomService(client)
+
+    result = service.leave()
+
+    assert client.calls[0][1] == "/workrooms/leave"
+    assert str(result.workroom_id) == "ffffffff-ffff-ffff-ffff-ffffffffffff"
+    assert result.access_token == "global-token"
+    assert result.expires_in == 3600
+
+
+# =============================================================================
 # Export Manifest
 # =============================================================================
 


jxstanford (dismissed):
# Review Report: ENG-6538 Align Context no-workroom live SDK tests

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/156
**Branch**: `fix/eng-6538-context-no-workroom-sdk-tests-develop` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 9/10 | Precisely encodes the new `workroom_scope_required` contract (Global read-only → 400 + `workroom_scope_required`); no scope creep. Only ding: a now-orphaned `shared_vectordb` fixture left behind. |
| Security | 9/10 | Test-only change; no secrets, injection, or auth surface introduced. Workroom scoping is a server-authorized request, correctly documented as non-spoofable. |
| Test Quality | 8/10 | Strong symmetric contract coverage on both endpoint shapes (instance + `_global`) and the negative path (status code *and* message asserted); live-gated, so collection-verified only — appropriate. |
| Error Handling | 7/10 | Teardown is best-effort and correct *except* the `enter` call placed outside `try/finally` (Medium #1) — a real cleanup gap both engines flagged. |
| Code Quality | 8/10 | Consistent with surrounding fixtures and the sibling `test_workroom_isolation_live.py`; the dead `shared_vectordb` fixture is the main blemish. |
| Refactoring | 7/10 | Two missed cleanups: removing the dead fixture and tightening the `enter` placement. |

## Critical Issues (Must Fix)
| — | — | None | — | — |

## High Priority Issues
| — | — | None | — | — |

| 1 | test_context_live.py:316 | The `client.post(f"/workrooms/{workroom_id}/enter")` call sits *outside* the `try/finally`. If `enter` raises after `workrooms.create()` succeeds, pytest never reaches the fixture's `finally`, so the newly created `sdk-ctx-session-*` workroom is leaked permanently — the autouse stale cleanup (`_cleanup_stale_sdk_vdbs`) only reaps VDBs/ontologies, not workrooms. Move the `enter` call inside the `try`, or delete the workroom on setup failure. | Claude Code, Codex | Error Handling / Resource Cleanup |
| 2 | test_context_live.py:290-298 | `shared_vectordb` fixture is now dead code — no test references it after the insert/query tests migrated to `shared_workroom_vectordb` (grep confirms only the definition remains). It still provisions a Global VectorDB nobody consumes. Remove it, or note why it's retained. | Claude Code | Code Quality / Dead Code |
| 3 | test_context_live.py:316 | Raw `client.post(...)` for "enter" bypasses `WorkroomService`, which has no `enter()` method. Acceptable since none exists, but consider adding one to keep the SDK surface consistent (other workroom ops go through the service). | Claude Code | Codebase Consistency |

## Test Coverage Assessment
- **Adequate.** `tests/integration/test_context_live.py` *is* the test surface. The contract change (Global read-only → `workroom_scope_required`) is covered by a renamed negative test (asserts `400` + `"workroom_scope_required"`) and positive workroom-scoped insert/query/search/retrieve coverage. Both endpoint shapes (instance + `_global`) are preserved.
- **No new gaps.** These are live integration tests gated behind `@pytest.mark.live`/`requires_embedding_model`, validated by `--collect-only` (a full run requires a live Kamiwaza server CI can't provide). All passed `workroom_id` kwargs match existing SDK signatures (`insert_vectors`, `query_vectors_global`, etc.) — verified.

## Positive Observations
- Contract migration is clean and symmetric: status-code *and* message asserted on the negative path; positive paths exercise a real per-session workroom.
- `session_workroom` teardown correctly catches both `APIError` and the sibling `NotFoundError` (a non-subclass sibling), matching the established pattern in `test_workroom_isolation_live.py`.
- Stale docstrings describing the obsolete Global-read-only contract were removed rather than left to rot — good comment hygiene.
- The `X-Workroom-ID` header path is real and documented inline with the istio strip-identity-headers exemption rationale.
- Linear ticket ENG-6538 present in the branch name — Kamiwaza repo policy gate satisfied.

## Overall Verdict
**Risk Level**: Low
**Verdict**: APPROVE

**Rationale**: A focused, correct test-alignment PR that accurately encodes the new `workroom_scope_required` contract with symmetric coverage. No Critical issues, all dimension scores ≥ 7, no quality-gate failures, and the required Linear ticket (ENG-6538) is present. The two substantive findings — a workroom-leak risk from the `enter` call outside `try/finally` (both engines) and a dead `shared_vectordb` fixture — are Medium cleanups that do not block approval, but fixing #1 before merge is recommended since it can orphan live resources on failed runs.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

**Reviewed-SHA**: db4be3fc620b26f2091a3fe41b762b65e9c402be

m9e (dismissed):
# PR Review Report: ENG-6538 Align Context no-workroom live SDK tests

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/156
**Branch**: `fix/eng-6538-context-no-workroom-sdk-tests-develop` → `develop`
**Scope**: 1 file, +44/-49 (`tests/integration/test_context_live.py`)

---

| 1 | `tests/integration/test_context_live.py:316` | `session_workroom` calls `shared_context_service.client.post(f"/workrooms/{workroom_id}/enter")` on the **session-scoped** shared client and never reverts the bound workroom in teardown. This mutates server-side session state for every later test using that client. Codex identifies a concrete functional regression: the `_global` tests and ontology tests that omit `workroom_id` now silently exercise the per-session workroom instead of Global/default scope, making the suite **order-dependent**. After workroom delete in the `finally`, the shared client is left pinned to a non-existent workroom for any subsequent test. Plus: the POST return value is discarded, so a 4xx/5xx from `/enter` (or future contract change requiring a body) silently fails and cascades into every workroom-scoped test as a confusing downstream failure. | **SP, CL, CX, KZ, AG** (5/5) | Bug / Test Pollution / Silent Failure |

| 2 | `tests/integration/test_context_live.py:290-298` | `shared_vectordb` fixture is now **dead code** — no test in the file references it after the renames to `shared_workroom_vectordb`. Worse, it calls `_create_temp_vectordb(service, prefix="sdk-shared-vdb")` with `workroom_id=None`, which under the very contract this PR codifies (`workroom_scope_required` 400 on un-scoped create) would now fail if it were ever invoked. Delete it or alias it to `shared_workroom_vectordb`. | SP, CL | Dead Code / Contract Drift |
| 3 | `tests/integration/test_context_live.py:426-442` | `test_context_vectordb_insert_vectors_global` and `test_context_vectordb_query_vectors_global` now pass `workroom_id=session_workroom` — but the `_global` suffix is misleading since these no longer exercise Global-workroom semantics. They exercise the body-`vectordb_id` endpoint shape (`/context/vectordbs/insert\|query`) against a workroom-scoped VDB. Rename to `_collection_level` / `_body_payload` / similar, or deprecate the SDK `_global` methods if they no longer have a Global meaning. | CL, AG | Naming / API Contract |
| 4 | `tests/integration/test_context_live.py:271-287` | `_cleanup_stale_sdk_vdbs` still iterates `(None, DEFAULT_WORKROOM_ID)` for `list_vectordbs`/`list_ontologies`. Under the new contract, no SDK code path will ever write into the `None` scope, so this iteration is wasted round-trips that misleadingly suggest no-workroom cleanup is needed. Narrow to scopes the suite actually writes to (or, to keep coverage, iterate live workrooms). | SP, CL | Code Quality / Performance |

| 5 | `tests/integration/test_context_live.py:391-392` | `assert "workroom_scope_required" in str(exc_info.value)` is a brittle substring match against the stringified `APIError`. Brittle to error-payload serialization changes. Prefer a structured assertion: `exc_info.value.code == "workroom_scope_required"` (or whatever attribute the SDK surfaces on `APIError`). | SP, CL, AG | Test Robustness |
| 6 | `tests/integration/test_context_live.py:316` | **Layering smell**: reaching past the SDK service facade into raw `client.post(...)` for what is now a first-class workroom operation under ENG-6538. Add a `WorkroomService.enter()` (and probably a paired `leave()`) helper mirroring `create`/`delete`/`archive`. This fixes the unchecked-return issue, replaces the stringly-typed path, and gives the SDK a stable surface for the new contract. | SP, CL | Refactoring / Architecture |
| 7 | `tests/integration/test_context_live.py:445-468` | Insert/query tests share one workroom VDB but each creates a fresh collection (`_sdk_collection_name()` with random suffix). Collections from `insert_vectors` tests are never explicitly deleted; only the parent VDB is GC'd on session end. Add `_safe_delete_collection` in a `finally` to keep the workroom tidy across in-session reruns. | SP | Test Hygiene |
| 8 | `tests/integration/test_context_live.py` | Four `insert/query (instance\|global)` tests are near-duplicates. Strong candidate for `pytest.mark.parametrize` over `(insert_fn_name, query_fn_name)` — cuts ~70 lines and makes future contract shifts a one-line edit. | SP | Refactoring |
| 9 | `tests/integration/test_context_live.py:379` | The renamed `test_context_vectordb_create_without_workroom_requires_scope` test could be migrated from per-test `_context_service(live_kamiwaza_client)` to `shared_context_service` for consistency. **Counter-consideration**: keeping a fresh, unbound client is intentional here — the whole point is asserting the un-entered behavior, and using the bound shared client could mask the regression. Use a fresh, *un-bound* fixture instead. | AG | Code Quality |
| 10 | `tests/integration/test_context_live.py:386-392` | The new test exercises only the **implicit** `workroom_id=None` case. Add a paired negative for explicit `workroom_id=DEFAULT_WORKROOM_ID` (Global) to confirm the contract is "non-Global required" rather than just "non-empty". | CL | Test Coverage |
| 11 | `tests/integration/test_context_live.py` | No cross-workroom negative test. Create a second workroom, attempt insert into `shared_workroom_vectordb` with the second workroom's id in the header; assert rejection. This would actually validate end-to-end the "verified workroom authority" the new docstring promises, vs. just routing happy-path. | CL | Test Coverage |
| 12 | `tests/integration/test_context_live.py:379-392` | Dropped the cross-reference to where the contract is enforced server-side. A one-line pointer (server file path or ENG ticket) keeps grep-driven archaeology cheap. | CL | Documentation |

## Test Coverage Assessment

- **Coverage verdict**: Adequate for the happy paths and the implicit-no-workroom negative; **gaps** on explicit-Global-rejected and cross-workroom-authority cases.
- **Missing test cases** (consensus from CL + KZ):
  - `create_vectordb(workroom_id=DEFAULT_WORKROOM_ID)` — confirm the contract is *non-Global required*, not just *non-empty*.
  - Cross-workroom write attempt — confirm `X-Workroom-ID` mismatch is rejected (validates the "verified workroom authority" wording).
  - Implicit-scope coverage: if the new `/enter` binding is supposed to let later calls omit `workroom_id`, no test exercises that path — every workroom-scoped test still passes `workroom_id=` explicitly.
  - Status-code assertion on `/workrooms/{id}/enter` so a server regression on that endpoint fails loud, not silent.

## Reviewer Agreement Matrix

| Category | Superpowers | Claude | Codex | Kamiwaza | Antigravity | Consensus |
|----------|-------------|--------|-------|----------|-------------|-----------|
| Session state mutation in `/enter` | High | Important (80) | **P1 only finding** | High | **Critical** | **5/5** — primary blocker |
| Dead `shared_vectordb` fixture | High | Important (95) | — | — | — | 2/5 — strong |
| Brittle string-match on error | Low | Suggestion (70) | — | — | Low | 3/5 |
| Misleading `_global` naming | Low | Suggestion (60) | — | — | High | 2/5 |
| `_cleanup_stale_sdk_vdbs` stale None iter | Medium-High | Suggestion (50) | — | — | — | 2/5 |
| Layering / `WorkroomService.enter()` | Medium (refactor) | Important | — | — | — | 2/5 |
| Test coverage gaps | — | High | — | Medium | — | 2/5 |

## Reviewer Verdicts

| Superpowers | Comment |
| Claude | Comment |
| Codex | Request Changes (P1: order-dependent functional regression in the suite) |
| Kamiwaza | Comment (Risk: Medium — flags the session-state issue as High but stops short of Request Changes) |
| Antigravity | Request Changes (Critical: test pollution from un-reverted shared session state) |

**Vote Tally**: Approve: 0 | Comment: 3 | Request Changes: 2 (out of 5 successful reviewers)

- Tight, focused diff: 1 file, +44/-49, single commit, single purpose — easy to review and revert.
- Docstring rewrite on `session_workroom` correctly captures the new contract — terse, accurate, no stale Global-read-only references.
- Renamed test (`..._create_in_global_workroom_is_read_only` → `..._create_without_workroom_requires_scope`) is descriptive and aligned with the actual assertion (`403` → `400` + `workroom_scope_required`).
- Reusing `shared_context_service` for workroom-bound tests is structurally the right pattern — workroom-enter is per-session state, so amortizing across the suite makes sense (the only issue is the missing teardown).
- The fresh `_context_service(live_kamiwaza_client)` in the negative test correctly isolates the scope-required failure from the pre-bound shared client.
- Teardown preserves the `(APIError, NotFoundError)` catch with the explanatory comment about the sibling-not-subclass relationship.
- The old Global-read-only rationale is preserved as a block comment after the renamed test, so the institutional knowledge isn't lost.
- Session-scoped fixtures cut wall-clock cost of the live suite materially vs. per-test workroom/VDB creation.
- Diff stat reduction (49 deletions vs. 44 additions) shows real simplification.

## Overall Verdict

**Risk Level**: Medium (low blast radius — integration tests only — but the session-state mutation undermines the suite's signal on every test that follows)
**Reviewer Consensus**: 5/5 reviewers flagged the `/workrooms/{id}/enter` session state mutation as the highest-priority issue; 2/5 escalated to Request Changes outright, and 3/5 raised it as a High/Important finding. Per the synthesis rules, a remaining Critical/High finding mandates Request Changes regardless of vote tally.

### What "Request Changes" Means Here

Land after addressing — at minimum — these three items:

1. **Fix or document the `/workrooms/{id}/enter` semantics.** Either:
   - Add a `WorkroomService.enter()` / `leave()` pair to the SDK and call `leave()` in teardown (preferred), OR
   - Add an explicit `assert response.ok` on the raw POST AND a teardown step that resets/leaves the workroom before deletion, AND inline-comment why `/enter` is required in addition to per-call `X-Workroom-ID` headers (since the SDK already injects that header — Claude's review couldn't determine what `/enter` mutates beyond that).
2. **Delete the orphaned `shared_vectordb` fixture** (or alias it to `shared_workroom_vectordb`) — it can't be invoked without violating the very contract this PR codifies.
3. **Make the `workroom_scope_required` assertion structured** rather than substring-matching the stringified `APIError`.

Items 4–12 are non-blocking follow-ups that can land in a sibling PR.

---

*This review was generated by 5 independent reviewers running in parallel via the PRReview skill, then synthesized. Each reviewer read the diff and surrounding context independently; agreement noted in the Flagged By column.*

kbot-ships:
_Reviewing head `db4be3f` against `develop`._

## Bottom line
This is a test-only PR that realigns the live Context SDK tests with the ENG-6538 no-workroom/workroom-scope contract from kamiwaza PR #1893. It (a) flips the global VectorDB-create expectation from the stale `403 "Global Workroom is read-only"` to `400 workroom_scope_required`, (b) adds an explicit `/workrooms/{id}/enter` bind on the shared session, and (c) repoints the vector insert/query coverage at a workroom-scoped VectorDB while keeping both instance and global endpoint shapes exercised. I traced the SDK plumbing end-to-end and the changes are internally consistent and correct. Direction is right; verdict is **APPROVE** with one cleanup note.

## Findings

### 1. `shared_vectordb` fixture is now dead code — **(nit)**
`tests/integration/test_context_live.py:290-298`. After repointing the four vector tests at `shared_workroom_vectordb`, the session-scoped `shared_vectordb` fixture (global, non-workroom VDB) has no remaining consumers — grep confirms line 291 is the only reference. Since pytest only instantiates requested fixtures it's harmless at runtime, but it now provisions/teardowns nothing and reads as live coverage that doesn't exist. Either delete it or wire it into a test that still asserts the global-create path. Ruff won't flag this (fixture defs aren't "unused imports"), so it'll silently rot.

### 2. New-contract assertion is sound (verified, not a defect)
`test_context_vectordb_create_without_workroom_requires_scope:379-392` asserts `pytest.raises(APIError)` + `status_code == 400` + `"workroom_scope_required" in str(...)`. I confirmed this holds against the client: `(400, "workroom_scope_required")` is **not** in `_REASON_TO_EXCEPTION` (`kamiwaza_sdk/exceptions.py:286`), so `error_for_response` returns base `KamiwazaError`, which makes `_raise_for_error_response` (`client.py:583+`) fall through to `raise APIError(...)`. The APIError message is `f"API request failed with status {status}: {response_text}"`, so the reason string surfaces in `str(exc)`. Good — no fragile coupling to a typed subclass. Noting it because the assertion silently depends on `workroom_scope_required` never being promoted into the typed-reason table later; if it ever is, this test breaks since the typed class isn't an `APIError`.

## Test gaps
- These are `live`/`integration`/`withoutresponses` tests that require a real cluster and are not exercised by the listed verification (`py_compile`, `ruff`, `--collect-only`, and *unrelated* unit suites). So a revert of the server-side contract would not be caught by anything that runs in CI here — that's inherent to live SDK tests, not a regression this PR introduces, but worth stating plainly: the new `400 workroom_scope_required` expectation is unverified against a running server in this PR.
- `session_workroom:316` issues `client.post("/workrooms/{id}/enter")` with the default `expect_json=True` and discards the result. If that endpoint ever returns 200 with an empty/non-JSON body, `_parse_response` raises `APIError("Failed to parse JSON response")` and would fail the bind rather than the assertion under test. Low risk (enter endpoints typically return JSON, and 204 is handled), but it's an untested assumption in a fixture that gates most of this file.

## What's good
The docstrings and the `:404` comment block explaining the `X-Workroom-ID` istio-exemption and why the dedicated VDB-instance lifecycle test stays in the kamiwaza repo are genuinely useful context for future readers. Teardown correctly catches both `APIError` and the sibling `NotFoundError`. All four vector methods confirmed to accept `workroom_id`, so threading `session_workroom` through both instance and global shapes is type-consistent.

## Verdict
**Verdict**: APPROVE
Correct, well-documented test alignment with the new workroom-scope contract; only a dead `shared_vectordb` fixture to clean up and two non-blocking coverage caveats.
---

kbot-ships:
Fresh-review note for current head `56f813f`:

- Addressed the `m9e` blocker by adding typed `WorkroomService.enter()` / `leave()` SDK methods and using them from `session_workroom`.
- `session_workroom` now enters inside the `try` and always attempts `leave()` before deleting the ephemeral workroom, so the shared live client is not left pinned to a deleted workroom.
- Removed the dead `shared_vectordb` fixture that could still provision an unscoped VDB under the new no-workroom contract.
- Replaced the brittle stringified `APIError` assertion with structured `response_data.detail.error` extraction for `workroom_scope_required`.

Local verification:
- `uv run pytest tests/unit/test_workroom_service.py --no-cov -q` -> 55 passed
- `python -m py_compile tests/integration/test_context_live.py kamiwaza_sdk/services/workrooms.py kamiwaza_sdk/schemas/workrooms.py` -> clean
- `uv run ruff check tests/integration/test_context_live.py kamiwaza_sdk/services/workrooms.py kamiwaza_sdk/schemas/workrooms.py tests/unit/test_workroom_service.py` -> clean
- `uv run pytest tests/integration/test_context_live.py tests/integration/test_workroom_isolation_live.py --collect-only -q` -> 42 collected
- `uv run pytest tests/unit/test_context_service.py tests/unit/test_workroom_service.py --no-cov -q` -> 90 passed
- `git diff --check` -> clean

kbot-ships:
Follow-up on current head `5e32e8c`: fixed the CI mypy failure exposed by touching `kamiwaza_sdk/services/workrooms.py` by preserving Pydantic runtime coercion through `CreateWorkroom.model_validate(...)` / `UpdateWorkroom.model_validate(...)`. Local `uv run mypy --follow-imports=silent kamiwaza_sdk/schemas/workrooms.py kamiwaza_sdk/services/workrooms.py` now reports no issues; prior review-remediation tests remain green.

jxstanford (dismissed):
# RE-REVIEW Report: ENG-6538 Align Context no-workroom live SDK tests

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/156
**Branch**: `fix/eng-6538-context-no-workroom-sdk-tests-develop` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

CI Status: All checks passed / no conflicts. Locally verified: `py_compile` OK, `ruff` clean, unit suite (incl. 2 new enter/leave tests) green, integration collection wires all tests without fixture errors. The Critical finding below is a code-quality/correctness concern, not a CI failure.

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 7/10 | Test-alignment scope (no-workroom→`workroom_scope_required`, session-bound workroom, workroom-scoped VDB coverage) is met cleanly. Lowered from Claude's 9 because the supporting `enter()`/`leave()` API does not deliver the binding its docstring advertises (Codex Critical #1). |
| Security | 9/10 | Test alignment + thin SDK wrappers; no injection/secret/authz surface introduced. Only test UUIDs, no hardcoded credentials. |
| Test Quality | 7/10 | Solid unit coverage for the new SDK methods and specific live contract assertions. Lowered from Claude's 8: both engines confirm the bound-session path is never actually exercised — every data-plane test threads an explicit `workroom_id`, so `enter()`/`leave()` is belt-and-suspenders and the unit tests only assert response parsing, not real binding. |
| Error Handling | 7/10 | Teardown `leave()` can leak the ephemeral workroom on a `pydantic.ValidationError` (sibling of `APIError`, not caught). Otherwise best-effort cleanup matches sibling tests. |
| Code Quality | 8/10 | Idiomatic, lint-clean, good docstrings/section comments. Lowered from Claude's 9 because `enter()`/`leave()` docstrings ("Bind the current SDK session") overclaim relative to actual behavior. |
| Refactoring | 9/10 | Dead `shared_vectordb` fixture and obsolete Global-read-only test removed; duplication avoided via shared fixtures. Minor missed opportunity to unify teardown exception handling. |

## Critical Issues (Must Fix)
| 1 | kamiwaza_sdk/services/workrooms.py:230-243 | `enter()`/`leave()` validate and return the response (incl. `access_token`/`expires_in`) but never install the scoped token into the client session or authenticator. `_prepare_request_kwargs` calls `authenticator.authenticate()` before every request, re-stamping the original bearer/PAT — so subsequent calls do NOT carry the workroom-scoped token, and the live `session_workroom` fixture's claim of "verified workroom authority" via `enter()` is not actually realized (the tests only pass because they thread explicit `workroom_id`). The presence of `access_token`/`expires_in` in the response models strongly implies the client is meant to apply them. Resolve by applying the returned token in `enter()` (and restoring the global token in `leave()`), OR document that binding is purely server-side and justify returning unused token fields. | Codex (P1); Claude noted same observation at Medium/Low confidence | Architecture / Correctness |

## High Priority Issues
| — | — | None beyond the Critical item above | — | — |

| 1 | tests/integration/test_context_live.py:330-333 | Teardown wraps `leave()` in `except APIError`, but `leave()` → `LeaveWorkroomResponse.model_validate(response)` raises `pydantic.ValidationError` (a sibling of `APIError`, not a subclass) on an unexpected payload shape. That would propagate out of `finally` and skip `workrooms.delete(...)`, leaking the ephemeral workroom. Mirror the delete block and also catch `ValidationError` (or broaden). | Claude (single-source) | Error Handling |
| 2 | tests/integration/test_context_live.py:326 | `enter()` runs on the session-scoped `shared_context_service` client and is not reset until teardown `leave()`. Safe today, but introduces ordering coupling for any future `shared_context_service` test expecting unbound behavior. | Claude (single-source) | Test Design |

## Test Coverage Assessment
- **Adequate**: `enter`/`leave` have unit tests asserting endpoint path + parsed fields; response models exercise `extra="allow"` forward-compat. Integration realignment threads `session_workroom` + `shared_workroom_vectordb` through insert/query/collection/search/retrieve.
- **Gaps**: The actual bound-session behavior is not tested — the unit tests confirm only that the response parses, not that any token is applied or that requests carry workroom scope by binding (Critical #1). `enter()`'s 404→`NotFoundError` branch has no unit test (minor). Live-only contract assertions (`400 workroom_scope_required`, `_api_error_code` against a real payload) are not exercised by CI's `--collect-only` and depend on a live cluster.

## Positive Observations
- `_api_error_code` is a well-designed defensive helper handling nested `detail` and top-level `error`/`code`/`reason` shapes, returning `None` safely on non-dict payloads.
- `workroom_id` is threaded consistently through every data-plane call, keeping tests self-contained regardless of session binding.
- New response models inherit `extra="allow"` for forward compatibility with additional server fields.
- The `CreateWorkroom.model_validate({...})` / `UpdateWorkroom.model_validate(...)` refactor is functionally equivalent and satisfies the type checker — clean, low-risk.
- Stale `shared_vectordb` fixture and the obsolete Global-read-only test were removed rather than left dangling; docstrings updated to the new contract.

## Overall Verdict
**Risk Level**: Medium
**Verdict**: REQUEST_CHANGES

The test-alignment goal (the core of ENG-6538) is delivered correctly, lint/unit/collection checks pass, and no security defects were introduced. However, Codex raised a concrete, mechanically-substantiated P1 on code newly added in this PR: the `enter()`/`leave()` public methods — and the live fixture that depends on them for "verified workroom authority" — do not actually bind the session, since the returned scoped token is never installed and is overwritten by `authenticator.authenticate()` on every request. Claude independently observed the same gap at lower confidence. Because this is a public API whose behavior contradicts its docstring (and a live fixture relies on it), it should be resolved before merge — either wire the returned token into the session/authenticator, or document server-side binding semantics and remove/justify the unused token fields. Also address the teardown `ValidationError` leak (Medium #1).

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

**Reviewed-SHA**: 5e32e8cf76077b042cc7aee42aed9e16125ee1e0

kbot-ships:
_Reviewing head `5e32e8c` against `develop`._

## Bottom line
This PR realigns the live Context SDK tests to the ENG-6538 no-workroom contract from kamiwaza PR #1893: VectorDB create without an explicit non-Global workroom is now expected to fail `400 workroom_scope_required` (replacing the stale `403 Global Workroom is read-only`), the per-session workroom is bound via two new additive SDK methods (`enter`/`leave`), and vector insert/query coverage moved onto a workroom-scoped VectorDB while keeping both endpoint shapes. The direction is correct and the change is well-scoped: the removed `shared_vectordb` global fixture has no dangling references, every `_global`/instance method already accepts the `workroom_id` kwarg the tests now pass, `model_validate` swaps are behavior-preserving (neither `CreateWorkroom` nor `UpdateWorkroom` declares field aliases), and the new service methods carry unit coverage. Previous-review items appear addressed (session reset via `leave()` in teardown — commit 56f813f; type-check cleanup — 5e32e8c). No blocking issues. **APPROVE** with two advisory notes.

## Findings

### 1. `enter()`/`leave()` discard the returned `access_token`; scope is actually carried by the header — `kamiwaza_sdk/services/workrooms.py:230-244`
`enter()` parses `EnterWorkroomResponse` (which includes `access_token`/`expires_in`) but never applies that token back to the client, and `leave()` likewise. Meanwhile the actual workroom scoping in every test is carried by the explicit `workroom_id=` argument, which `context.py` turns into the `X-Workroom-ID` header / query param on each call (confirmed at `context.py:19-31`). So the `enter()` call in the `session_workroom` fixture (`test_context_live.py:326`) does not change SDK request behavior — the tests would scope identically without it. That's fine for matching the server-side session contract, but a consumer who calls `enter()` expecting subsequent SDK calls to be workroom-scoped *without* passing `workroom_id` would silently operate against Global. Worth either wiring the returned token into the client session or documenting in the docstring that scope still requires the per-call `workroom_id`/header. **(worth fixing)**

### 2. `_api_error_code` parsing has no offline coverage — `test_context_live.py:263-280`
The new helper that drives the rewritten contract assertion (`assert _api_error_code(...) == "workroom_scope_required"`) is only exercised inside the live integration test, which is gated behind `live`/`integration` markers and skipped in normal CI. Its branch logic (detail-dict vs top-level, `error`/`code`/`reason` precedence) is non-trivial and untested offline. A small unit test feeding a synthetic `APIError(response_data=...)` would lock the parsing contract independent of a live server. **(nit)**

## Test gaps
- The new `enter`/`leave` methods have unit tests asserting endpoint path and response parsing (`test_workroom_service.py:466-509`) — adequate for the happy path. There is no unit test for the `enter()` 404→`NotFoundError` mapping (`workrooms.py:237-239`), which is the one piece of branching logic in the new methods; if reverted to a bare raise, nothing would catch it. Consider adding a `_raise_404` case mirroring the existing `get`/`archive` 404 tests.
- The contract-flip assertion (`400`/`workroom_scope_required`) lives only in a live test, so a server-side regression to the old `403` behavior would not be caught in CI — acceptable given this is intentionally a live-suite alignment, but noting the coverage boundary.

## What's good
- Clean removal of the stale global fixture with no orphaned references, and the docstring rewrite on `session_workroom` accurately states the new contract.
- New SDK methods are additive, follow the existing `_ensure_uuid` + 404→`NotFoundError` pattern, and ship with unit tests.
- Teardown is correctly best-effort: `leave()` wrapped in `try/except APIError` and `delete()` catching both `APIError` and the sibling `NotFoundError`.

## Verdict
**Verdict**: APPROVE

Test-alignment plus additive, unit-tested SDK methods that correctly track the ENG-6538 contract; the two notes are advisory and none rise to a correctness or security blocker.
---

kbot-ships:
Current-head review update (`fe7d477`):

- Clarified `WorkroomService.enter()` / `leave()` as checked backend endpoint wrappers; they parse and expose returned token fields but do not mutate the SDK authenticator or imply client-side token rebinding.
- Updated the live Context fixture so it no longer claims SDK-session-derived authority; Context calls continue to pass explicit `workroom_id`.
- Hardened teardown so a `pydantic.ValidationError` from `leave()` cannot skip ephemeral workroom deletion.
- Added a unit contract check that enter/leave expose returned tokens without mutating the client auth header.
- Tightened generator fixture annotations in the touched live-test file so targeted mypy verification is clean.

Local verification:
- `uv run pytest tests/unit/test_workroom_service.py --no-cov -q` -> 56 passed
- `uv run pytest tests/unit/test_context_service.py tests/unit/test_workroom_service.py --no-cov -q` -> 91 passed
- `uv run pytest tests/integration/test_context_live.py tests/integration/test_workroom_isolation_live.py --collect-only -q` -> 42 collected
- `uv run ruff check kamiwaza_sdk/services/workrooms.py kamiwaza_sdk/schemas/workrooms.py tests/integration/test_context_live.py tests/unit/test_workroom_service.py` -> pass
- `uv run mypy --follow-imports=silent kamiwaza_sdk/schemas/workrooms.py kamiwaza_sdk/services/workrooms.py tests/integration/test_context_live.py tests/unit/test_workroom_service.py` -> pass
- `uv run python -m py_compile ...` -> pass
- `git diff --check` -> pass

Re-requested review from `jxstanford` and `m9e` on the current head.

jxstanford (approved):
# RE-REVIEW Report: ENG-6538 Align Context no-workroom live SDK tests

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/156
**Branch**: `fix/eng-6538-context-no-workroom-sdk-tests-develop` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

CI Status: All checks passed / no conflicts. (ruff clean, 91 unit tests pass, 19 integration tests collect.)

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 9/10 | Implementation matches stated scope exactly — workroom-scope contract, session binding via enter/leave seam, dual endpoint-shape coverage; no scope creep. Codex flagged no design deviations. |
| Security | 9/10 | No token leakage (`enter`/`leave` deliberately do not mutate the authenticator), no secrets logged, thin passthrough with no injection surface. Test asserts `Authorization` header is left intact. |
| Test Quality | 8/10 | Strong unit + live coverage of the new seam; one effectively tautological unit test and a missing `enter()` 404 negative-path unit case. |
| Error Handling | 9/10 | Consistent `NotFoundError` mapping on 404; best-effort teardown with precise sibling-vs-subclass exception handling; intentional fail-loud on the enter seam in the fixture. |
| Code Quality | 9/10 | Clean, idiomatic, well-documented; `model_validate` refactor in `create()`/`update()` is behavior-preserving and consistent across methods. |
| Refactoring | 9/10 | Dead `shared_vectordb` fixture fully removed (zero orphan refs), `Generator[...]` annotations tightened; minor remaining opportunity (collapse the tautological test). |

## Critical Issues (Must Fix)
| — | — | None | — | — |

## High Priority Issues
| — | — | None | — | — |

| 1 | tests/unit/test_workroom_service.py:512-542 | `test_enter_leave_expose_tokens_without_mutating_client_auth` is effectively tautological — `enter()`/`leave()` never touch `client.session`, so the `Authorization` assertion holds regardless. Useful as a regression guard documenting the no-token-install contract, but it does not exercise a real mutation path. | Claude Code | Test Quality |
| 2 | tests/integration/test_context_live.py:331-332 | `workrooms.enter(...)` in the `session_workroom` fixture is unguarded; if the backend enter-response shape diverges (e.g. omits `workroom_id`), a `ValidationError` breaks setup for every workroom-scoped test. Intended "checked lifecycle seam," but couples the fixture tightly to the exact response contract. | Claude Code | Test Design |
| 3 | kamiwaza_sdk/services/workrooms.py:230-244 | `enter()` is exercised but every Context call still passes explicit `workroom_id`, so the entered session state is never actually relied on. Tests validate the endpoint round-trip, not that session binding confers authority — consistent with documented design, noting the coverage boundary. | Claude Code | Test Coverage |

## Test Coverage Assessment
- **Adequate**: `kamiwaza_sdk/services/workrooms.py` (enter/leave covered by 3 new unit tests asserting endpoint paths, response field parsing, and no-auth-mutation) and `kamiwaza_sdk/schemas/workrooms.py` (new `EnterWorkroomResponse`/`LeaveWorkroomResponse` models validated via unit tests + live fixture). The `model_validate` refactor in `create()`/`update()` is behavior-preserving and stays covered by the passing 91-test suite.
- **Gap (non-blocking)**: A negative-path unit test for `enter()` raising `NotFoundError` on 404 — the only branch in the new code lacking direct unit coverage — would round out the seam. Codex independently identified no actionable correctness gaps.

## Positive Observations
- `enter()`/`leave()` explicitly do **not** install returned `access_token` into the authenticator, and the docstrings say so — avoids silently leaking scoped tokens into client auth state and keeps the security model explicit. Tokens are exposed in models but never logged.
- `WorkroomResponseModel(extra="allow")` makes the new response models forward-compatible with additive backend changes, so the live seam won't break on server additions.
- Stale `shared_vectordb` fixture fully removed with zero orphan references; `Generator[...]` return annotations correctly added to all generator fixtures.
- `_api_error_code` helper extracts the stable server error code from both `detail`-nested and top-level shapes, making the `workroom_scope_required` assertion robust to payload nesting.
- Teardown is best-effort with precise sibling-vs-subclass exception handling (`NotFoundError` is a sibling of `APIError`), consistent with the existing `test_workroom_isolation_live.py` pattern.
- Linear ticket ENG-6538 is present in the branch name (Kamiwaza policy gate satisfied).

## Overall Verdict
**Risk Level**: Low
**Verdict**: APPROVE

A well-scoped, behavior-preserving alignment of live SDK tests with the ENG-6538 no-workroom contract, plus a thin, security-conscious enter/leave passthrough. No Critical or High issues from either engine; all dimension scores ≥7; no quality-gate failures; Linear ticket present. The three Medium notes are minor test-design observations, none blocking. All unit tests and lint pass, integration tests collect.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

**Reviewed-SHA**: fe7d477c9e1efd9e9b27919a9d346bf18060dd10

m9e:
# PR Re-Review Report: ENG-6538 Align Context no-workroom live SDK tests

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/156
**Branch**: `fix/eng-6538-context-no-workroom-sdk-tests-develop` -> `develop`
**Scope**: 4 files, +166/-70
**Re-Review SHA**: `5e32e8c` (previous review at `db4be3f`)
**Delta commits**: `56f813f`, `5e32e8c` (2 commits since prior review)

---

## Delta Focus — Did the Previous Findings Get Fixed?

The prior synthesized review (DISMISSED) had a 5/5-reviewer Critical and several Highs. This re-review focuses on the delta. Reviewer agreement on resolution:

| Prior finding (severity) | Status | Reviewer agreement | Evidence |
|---|---|---|---|
| **Critical (5/5)** — `session_workroom` raw `client.post('/workrooms/{id}/enter')` mutated the shared session-scoped client and never reset; downstream tests pinned to the deleted workroom | **Fixed (with caveat — see Important #3)** | SP, CL, KZ agree; CX says fix is incomplete (token not installed) | `tests/integration/test_context_live.py:325-340` now wraps `workrooms.enter(workroom_id)` inside `try:` and the `finally` calls `workrooms.leave()` (swallows `APIError`) **before** `workrooms.delete()`. The new `enter`/`leave` go through typed `WorkroomService.enter()` / `WorkroomService.leave()` in `kamiwaza_sdk/services/workrooms.py:230-244`. `enter()`'s return value is now also asserted: `assert str(entered.workroom_id) == workroom_id`. The test-suite pollution mode is removed; Codex argues a related SDK-correctness concern persists (see Important #3) |
| **High** — Dead `shared_vectordb` fixture provisioning a non-workroom Global VDB that the new contract would reject | **Fixed** | SP, CL agree | Fixture is fully removed at the prior file location (no remaining references) |
| **High** — Brittle substring assertion `"workroom_scope_required" in str(exc_info.value)` | **Fixed** | SP, CL, KZ agree | New helper `_api_error_code()` at `tests/integration/test_context_live.py:263-280` extracts the code structurally from `error.response_data["detail"]["error"|"code"|"reason"]` with `isinstance(str)` guards, then falls back to top-level. Assertion at `:406` reads `_api_error_code(exc_info.value) == "workroom_scope_required"` |
| **High** — `_global` test naming misleading (now exercise workroom-scoped VDBs) | **Not Addressed** | CL: still misleading; SP: not a real issue since `*_global` is the SDK method name | Tests at `:440-456`, `:485-508` still pass `workroom_id=session_workroom`. They exercise the `insert_vectors_global` / `query_vectors_global` SDK *methods*, which still require workroom scope. SP argues the test name mirrors the SDK method; CL argues the suffix implies Global Workroom scope. Cosmetic/readability tradeoff. |
| **High** — `_cleanup_stale_sdk_vdbs` still iterates `(None, DEFAULT_WORKROOM_ID)` even though the SDK won't write to `None` scope | **Not Addressed** | SP, CL agree | Line `:291` unchanged. The `None` iter now silently 400s-and-swallows on the first call, costing one wasted round-trip per session. Best left as transitional legacy-residue cleanup. Minor. |
| **Medium** — Missing typed `WorkroomService.enter()` / `leave()` helpers (layering smell) | **Fixed** | SP, CL, KZ agree | Added at `kamiwaza_sdk/services/workrooms.py:230-244` with typed `EnterWorkroomResponse` / `LeaveWorkroomResponse` schemas (`kamiwaza_sdk/schemas/workrooms.py:89-104`). `enter()` maps 404 -> `NotFoundError` consistent with sibling methods. New unit tests at `tests/unit/test_workroom_service.py:471-508` cover the happy path. |

**Bottom line on the delta**: the test-suite pollution failure mode (the actual reason the prior review was DISMISSED) is genuinely removed; the two unaddressed prior items are cosmetic / minor. Codex surfaces two new substantive concerns about whether `enter()` is a *complete* SDK API and whether the session-scoped fixture leaks workroom binding into `shared_ontology` — see Important Issues below.

---

## Critical Issues (Must Fix)

|---|---|---|---|---|
| — | — | None — no reviewer flagged a Critical issue introduced by the delta | — | — |

## High Priority Issues

|---|---|---|---|---|
| — | — | None at consensus High severity. See Important Issues #3 (codex calls these P1) for the highest-severity new concerns. | — | — |

## Important Issues (New, Worth Following Up Before Merge)

|---|---|---|---|---|
| 1 | `kamiwaza_sdk/schemas/workrooms.py:98-104` | `LeaveWorkroomResponse.workroom_id: UUID` is **required** (no default). If the server's `/workrooms/leave` response ever omits `workroom_id` (plausible — "you're returning to Global, here's no destination workroom"), `LeaveWorkroomResponse.model_validate(response)` at `services/workrooms.py:244` raises `pydantic.ValidationError`. That's not an `APIError`, so the `session_workroom` fixture's `except APIError: pass` (`tests/integration/test_context_live.py:331-333`) does NOT catch it. The exception then escapes `finally`, skipping `workrooms.delete(workroom_id)` and leaking the ephemeral test workroom; the shared client is also left bound to a non-existent workroom. Recommendation: either make `workroom_id: Optional[UUID] = None`, or broaden the fixture's `except` to `(APIError, ValidationError, Exception)` with a log. | CL (conf 70), KZ | Schema correctness / teardown robustness |
| 2 | `tests/integration/test_context_live.py:330-333` | `session_workroom`'s `finally` only swallows `APIError` from `workrooms.leave()`. Any non-`APIError` raised by `leave()` (transport error, `ValidationError` per #1, etc.) skips the subsequent `workrooms.delete()` — partially regressing the same failure mode the prior review's Critical was meant to prevent. Recommendation: broaden to `except Exception` (with a log) for the leave step; matches the existing best-effort idiom on the delete step. | CL (conf 78), KZ | Teardown robustness |
| 3 | `kamiwaza_sdk/services/workrooms.py:230-235` (P1 from Codex) | `WorkroomService.enter()` parses `/workrooms/{id}/enter`'s response (typed as `EnterWorkroomResponse` with `access_token`, `expires_in`) and returns it — but never installs the returned `access_token` onto `self.client` / `self.client.session`. Because `KamiwazaClient._prepare_request_kwargs()` re-runs `self.authenticator.authenticate(self.session)` on **every** request, follow-up calls keep sending the **pre-enter** bearer token (and, for `UserPasswordAuthenticator`, keep overwriting `session.cookies["access_token"]` back to the pre-enter value at `authentication.py:99`). For API-key / password-authenticated clients, `client.workrooms.enter(...); client.context.do_thing()` will therefore not actually bind the SDK session in the way the typed return value suggests. The live tests in this PR work around this by passing `workroom_id=session_workroom` explicitly (X-Workroom-ID header), so the *tests* don't depend on the binding — but the SDK ships an API that won't behave as documented for consumers who do depend on it. Recommendation: either (a) install the scoped token onto the session/authenticator inside `enter()` and revert it inside `leave()`, or (b) tighten the docstring to explicitly say "session binding only takes effect on transport stacks that propagate `Set-Cookie`; programmatic callers should keep passing `workroom_id=` on every call." | **CX (P1)** | SDK API contract / correctness |
| 4 | `kamiwaza_sdk/services/workrooms.py:230-244` (P1 from Codex) | `enter()` is called from a **session-scoped** fixture (`session_workroom`) on the **session-scoped** `shared_context_service`'s client, and the binding persists until session teardown. Other session-scoped fixtures that share that same client — most notably `shared_ontology` at `tests/integration/test_context_live.py:365-377`, which calls `_create_temp_ontology(shared_context_service, prefix="sdk-shared-ont")` with no explicit `workroom_id` — can be created **either before or after** `session_workroom.enter()` depending on pytest's fixture-instantiation order driven by which test runs first. If the vector/context tests (which request `session_workroom`) collect before the ontology tests, `shared_ontology` is provisioned inside the entered workroom; once that ephemeral workroom is deleted at session end, the ontology delete in `shared_ontology`'s `finally` may target a no-longer-resolvable scope. Tests that consume `shared_ontology` go through `_context_service(live_kamiwaza_client)` (a **different** client and a **different** authenticated session — verified at `:206-227`), so the read path is not pinned to the entered workroom — meaning the cross-client mismatch can also surface as "ontology was created in workroom X but the test reads it under client Y's default scope". Recommendation: either make `shared_ontology` pass an explicit `workroom_id=` (e.g. `DEFAULT_WORKROOM_ID` or `session_workroom` if it should be workroom-scoped), or scope `session_workroom` more tightly (per-module or function scope) so the binding doesn't outlive the tests that need it. | **CX (P1)**, KZ | Cross-fixture test isolation |
| 5 | `tests/unit/test_workroom_service.py:471-508` | New unit tests cover only the **happy path** for `enter()` / `leave()`. Missing: `enter()` 404 → `NotFoundError` branch, `enter()` propagation of non-404 `APIError`, `enter()` rejecting an invalid UUID, `leave()` handling of a minimal/forward-compatible response shape, `leave()` handling of an `APIError`. Pattern is established by the sibling `test_archive_not_found_raises` test (`:454-463`). | CL, KZ | Test coverage |

## Medium / Low Suggestions

|---|---|---|---|
| 6 | `kamiwaza_sdk/services/workrooms.py:230-244` | Docstrings on `enter()` / `leave()` are one-liners; sibling methods (`create`/`get`/`update`/`delete`/`archive`) have full `Args/Returns/Raises` blocks. Add at least `Raises: NotFoundError (404), APIError (other)` for `enter` and `Returns:` / `Raises:` for `leave` for consistency. They also hide the side-effect contract — these mutate **server-side session state** via the underlying `requests.Session` cookie. Worth an explicit one-line note (especially per Important #3, document that this binding may not persist across SDK-managed re-auth). | SP, CL |
| 7 | `kamiwaza_sdk/services/workrooms.py:241-244` | `leave()` does not map 404 → `NotFoundError`. All siblings (`get`/`delete`/`archive`/`update`/`enter`) have that mapping. Probably intentional since `/workrooms/leave` has no path-bound workroom resource, but worth a one-line comment if intentional. | SP, CL |
| 8 | `kamiwaza_sdk/services/workrooms.py:236-239` | `raise NotFoundError(...)` in `enter()` is missing `from e`, breaking the exception cause chain. Matches the pre-existing convention at lines 117, 168, 196, 223, 270, etc. — pre-existing tech debt, worth a follow-up sweep. | SP |
| 9 | `kamiwaza_sdk/services/workrooms.py:67-77`, `:150-159` | The `CreateWorkroom.model_validate(...)` / `UpdateWorkroom.model_validate(...)` migration is fine and runtime-equivalent (verified — same coercion, same `Enum` handling, same dump output). The tradeoff: it loses mypy field-level checking on the dict literal vs. direct kwargs. Pre-existing alternatives (`TypedDict`, `cast`, `# type: ignore`) would preserve more static signal; current approach is acceptable. | SP, CL |
| 10 | `tests/integration/test_context_live.py:291` | `_cleanup_stale_sdk_vdbs` still iterates `(None, DEFAULT_WORKROOM_ID)`. Under the new `workroom_scope_required` contract, the `None` iter is dead work — `list_vectordbs(workroom_id=None)` 400s, the helper swallows it, and no resources are reaped. Either drop it (`for workroom_id in (DEFAULT_WORKROOM_ID,)`) or comment that it's transitional residue cleanup from older SDK versions. | SP, CL |
| 11 | `tests/integration/test_context_live.py:263-280` | `_api_error_code()` is generic enough to live next to `APIError` in `kamiwaza_sdk/exceptions.py` (or as a typed `APIError.code` attribute) so other integration suites can stop substring-matching error payloads. Refactor candidate. | SP |
| 12 | `kamiwaza_sdk/services/workrooms.py` | Consider a `WorkroomSession` context manager (`with workrooms.session(workroom_id): ...`) wrapping `enter()` + `leave()` so test fixtures and consumers cannot forget the teardown. The `session_workroom` fixture would collapse to ~6 lines and the leave-then-delete order would be enforced by the type. | SP |

---

## Test Coverage Assessment

- **Coverage verdict**: **Adequate** for the happy paths plus the implicit-no-workroom negative case; **gaps** on `enter()` / `leave()` failure modes (see Important #5).
- **Files with adequate coverage**: `kamiwaza_sdk/services/workrooms.py` (existing CRUD/archive/admin/export coverage plus 2 new `enter`/`leave` happy-path tests); `kamiwaza_sdk/schemas/workrooms.py` (exercised via service tests and the schema-validation block in unit tests).
- **Suggested test cases for `enter` / `leave`**:
  1. `test_enter_not_found_raises` — 404 → `NotFoundError` (mirrors `test_archive_not_found_raises`).
  2. `test_enter_other_error_propagates` — 403 → raw `APIError` re-raise.
  3. `test_enter_accepts_uuid_object` — pass `WORKROOM_UUID` directly (mirrors `test_get_accepts_uuid_object`).
  4. `test_enter_ensure_uuid_raises_for_invalid_string` — pass `"not-a-uuid"` → `ValueError`.
  5. `test_leave_handles_minimal_response` — server returns just `{"workroom_id": "..."}`; verify default `message`/`access_token`/`expires_in` fill in.
  6. `test_leave_handles_missing_workroom_id` — server returns `{"message": "left"}` (no `workroom_id`). Currently fails (Important #1). A test would force the fix.
  7. `test_leave_allows_extra_fields` — verify forward-compatible extra fields (mirrors `test_response_models_allow_extra_fields:759-780`).
  8. `test_enter_response_with_extra_fields` — same for `EnterWorkroomResponse`.

---

## Reviewer Agreement Matrix

| Category | Superpowers | Claude | Kamiwaza | Codex | Consensus |
|---|---|---|---|---|---|
| Prior Critical (session-state pollution) test-suite impact fixed | Fixed | Fixed | Fixed | Fixed (test pattern), but SDK API is incomplete (Important #3) | **4/4 on test-suite impact** |
| Prior High (dead `shared_vectordb`) fixed | Fixed | Fixed | Fixed | n/a | **3/3** |
| Prior High (substring assertion) fixed | Fixed | Fixed | Fixed | n/a | **3/3** |
| Prior High (`_global` naming) addressed | Not a real issue | Not addressed | n/a | n/a | Mixed (cosmetic) |
| Prior High (`_cleanup_stale_sdk_vdbs` `None` iter) addressed | Not addressed | Not addressed | n/a | n/a | Not addressed (minor) |
| Prior Medium (typed `enter`/`leave`) fixed | Fixed | Fixed | Fixed | Fixed (but argues API is incomplete) | **4/4 on shape**, 1/4 calls API incomplete |
| **NEW** — `LeaveWorkroomResponse.workroom_id` required | — | Important | mentioned | — | 2/4 (Important, low confidence) |
| **NEW** — narrow `except APIError` around `leave()` | — | Important | Important | — | 2/4 (Important) |
| **NEW** — `enter()` doesn't install `access_token` on client | — | — | — | **P1** | 1/4 (CX single-source, P1) |
| **NEW** — Session-scoped enter() leaks binding into other session fixtures (e.g. `shared_ontology`) | — | — | Important | **P1** | 2/4 (Important / P1) |
| Test coverage for `enter`/`leave` failure paths | — | Suggestion | Important | — | 2/4 (follow-up) |

## Reviewer Verdicts

| Reviewer | Individual Verdict | Status |
|---|---|---|
| Superpowers | **Approve** (Risk: Low) | OK |
| Claude | **Approve** (Confidence: High; with optional follow-ups) | OK |
| Kamiwaza | **Comment** (output truncated mid-emit; substantive findings reached during reasoning were all "Important" follow-ups, no Critical/High blockers in the body that landed) | Partial |
| Codex | **Request Changes** (2 P1 findings: SDK API incomplete + session-scope leak into other fixtures) | OK |

**Vote tally (successful reviewers, N=4)**: Approve: 2 | Comment: 1 | Request Changes: 1

---

## Positive Observations

- The load-bearing prior Critical — the actual test-suite-breaking session-state pollution where the shared client was pinned to a deleted workroom for every later test — is **genuinely fixed**. Verified by code inspection: `tests/integration/test_context_live.py:325-340` now (a) calls typed `workrooms.enter(workroom_id)` inside `try`, (b) `finally:` calls `workrooms.leave()` (best-effort) **before** `workrooms.delete()`, (c) asserts the entered workroom id matches the created one.
- New typed `WorkroomService.enter()` / `WorkroomService.leave()` methods replace the prior raw `client.post` layering violation. They use dedicated typed response schemas (`EnterWorkroomResponse` / `LeaveWorkroomResponse`), inherit `WorkroomResponseModel` with `extra="allow"` for forward-compat, and `enter()` correctly maps 404 → `NotFoundError` consistent with sibling methods.
- The `_api_error_code()` helper at `tests/integration/test_context_live.py:263-280` is well-designed: scans `detail.{error,code,reason}` then top-level keys with `isinstance(str)` guards. Future-proof against minor server payload shape drift while still asserting a stable error code rather than a free-text message substring.
- New unit tests at `tests/unit/test_workroom_service.py:471-508` assert both the request path AND the response decoding (workroom_id, access_token, expires_in, message), giving regression coverage on the schema shape.
- Dead `shared_vectordb` fixture cleanly deleted (not just commented out); no dangling consumers.
- The 5e32e8c mypy fix using `CreateWorkroom.model_validate({...})` / `UpdateWorkroom.model_validate(...)` preserves runtime behavior exactly — same coercion, same enum handling, same downstream `model_dump(exclude_unset=True, mode="json")` output. Verified equivalent.
- Defensive `try/except APIError: pass` around `leave()` correctly handles the case where the SDK session was never bound (e.g. `enter()` raised mid-fixture-setup) — though see Important #1, #2 for non-`APIError` edge cases.
- Diff stat (`+166/-70`) shows real simplification despite adding the new methods, schemas, and tests.
- The two follow-up `kbot-ships` comments in the PR thread show the author proactively re-verified the fix locally with the relevant test suites and `mypy` after each commit.

---

## Overall Verdict

**Risk Level**: **Low–Medium**
**Recommendation**: **Comment**
**Reviewer Consensus**: 4 reviewers contributed; 2 Approve (SP, CL), 1 Comment (KZ), 1 Request Changes (CX). No clear approve majority and no Critical/High findings remain — which lands the synthesis in **Comment** territory per the workflow recommendation rules.

### Why Comment over Approve or Request Changes

- The previously-blocking Critical (test-suite session-state pollution) is genuinely resolved. SP and CL both reach Approve, and KZ's substantive findings are all Important follow-ups, not blockers.
- Codex's two P1 findings are real and well-reasoned, but their concrete impact on this specific PR is bounded: the tests pass `workroom_id=` explicitly via X-Workroom-ID header on every call, so the test suite does not depend on the `/enter` binding actually sticking. The findings matter most for (a) the SDK's *future* consumers who would trust `enter()` to bind the session, and (b) the `shared_ontology` fixture which could see workroom-scope leak depending on pytest collection order.
- No reviewer flagged a Critical or unambiguous-High issue introduced by the delta, so Rule 1 (Request Changes) does not apply.
- `approveVotes` = 2/4, not a majority, so Rule 2 (Approve) does not apply either.
- Per Rule 3, the residual is **Comment**: surface the findings, defer the merge decision to the author / maintainer who can judge whether codex's two P1s warrant blocking.

### Suggested follow-ups (candidate for a sibling PR or before merge if maintainer wants to be cautious)

1. **(Codex P1 #1)** Either install the scoped `access_token` returned by `enter()` onto the client / authenticator (and revert on `leave()`), or tighten the docstring to clarify that the binding is server-side and may not persist for SDK-managed re-auth.
2. **(Codex P1 #2)** Pass an explicit `workroom_id=` to `shared_ontology._create_temp_ontology` (e.g. `DEFAULT_WORKROOM_ID` if it should be Global, or `session_workroom` if workroom-scoped) — OR scope `session_workroom`'s `enter()` more tightly so it does not outlive other session-scoped fixtures that share the same client.
3. Make `LeaveWorkroomResponse.workroom_id` `Optional[UUID] = None` **and/or** broaden `session_workroom`'s leave-step `except` to a wider class (recommended: do both, log on any leave-step exception).
4. Add the 8 enter/leave failure-path unit tests listed above.
5. Either drop the `None` iteration in `_cleanup_stale_sdk_vdbs` or annotate why it stays.
6. Decide on the `_global` test naming (rename to `_via_global_endpoint` or document why the `_global` suffix matches the SDK method name).
7. Optional refactor: introduce `WorkroomService.session()` context manager so consumers can't forget `leave()`; relocate `_api_error_code()` next to `APIError`.

---

Legend: SP=Superpowers, CL=Claude, KZ=Kamiwaza, CX=Codex

*This re-review was generated by 4 independent reviewers running in parallel via the PRReview skill. All 4 returned usable output; Kamiwaza's was partial. The synthesis focuses on the delta between SHA `db4be3f` (prior review) and `5e32e8c` (HEAD).*

PR #196, opened 2026-06-28
feat(connectors): generalize the declarative config-form contract (ENG-7810)

shkevin:
Generalize the connector config-form contract so a connector declares its whole admin form in its manifest and the platform renders it generically — no per-connector form in core.

- Rich `ConfigField`/`ConfigSchema`: types (incl. multi-value lists), labels/help, group+order, importance, regex validation, labelled options for select/multi-select, static or templated defaults, visibility, and an `out` that routes a value to the connector config or the OAuth scopes.
- `provider.to_manifest()` emits `config_fields` (rich form spec) alongside `config_schema` (validation JSON Schema).
- 23 unit tests; ruff/black/isort clean.

Consumed by the google/m365/linear/hubspot connector PRs and the kamiwaza core renderer PR. **Merge this first.**

### DIFF
--- a/kamiwaza_sdk/connectors/__init__.py
+++ b/kamiwaza_sdk/connectors/__init__.py
@@ -15,7 +15,15 @@ instead of vendoring the platform source.
 
 from __future__ import annotations
 
-from .config_schema import ConfigField, ConfigSchema, ConfigType
+from .config_schema import (
+    ConfigField,
+    ConfigOption,
+    ConfigOutput,
+    ConfigSchema,
+    ConfigType,
+    FieldWidth,
+    Importance,
+)
 from .connector_proxy import ConnectorProxyRequest, ConnectorProxyResponse
 from .connector_token import ConnectorMintRequest, ConnectorMintResponse
 from .exceptions import ConnectorException, InvalidConfigException
@@ -40,28 +48,32 @@ from .server_kit import (
 )
 
 __all__ = [
+    "AuthModel",
     "ConfigField",
+    "ConfigOption",
+    "ConfigOutput",
     "ConfigSchema",
     "ConfigType",
-    "AuthModel",
-    "PerUserOAuth",
-    "ServiceToken",
-    "auth_model_from_kind",
-    "OAuthDescriptor",
-    "DeploymentDescriptor",
-    "ConstraintDescriptor",
-    "SurfaceDescriptor",
-    "ConnectorSpec",
-    "ConnectorProvider",
-    "validate_icon",
-    "create_connector_app",
-    "ExecuteRequest",
-    "VerifyRequest",
-    "OpResult",
-    "ConnectorProxyRequest",
-    "ConnectorProxyResponse",
+    "ConnectorException",
     "ConnectorMintRequest",
     "ConnectorMintResponse",
-    "ConnectorException",
+    "ConnectorProvider",
+    "ConnectorProxyRequest",
+    "ConnectorProxyResponse",
+    "ConnectorSpec",
+    "ConstraintDescriptor",
+    "DeploymentDescriptor",
+    "ExecuteRequest",
+    "FieldWidth",
+    "Importance",
     "InvalidConfigException",
+    "OAuthDescriptor",
+    "OpResult",
+    "PerUserOAuth",
+    "ServiceToken",
+    "SurfaceDescriptor",
+    "VerifyRequest",
+    "auth_model_from_kind",
+    "create_connector_app",
+    "validate_icon",
 ]
--- a/kamiwaza_sdk/connectors/config_schema.py
+++ b/kamiwaza_sdk/connectors/config_schema.py
@@ -210,7 +210,14 @@ class ConfigField:
                 ),
                 default_template=data.get("default_template", ""),
                 depends_on=data.get("depends_on", ""),
-                visible_when=tuple(data.get("visible_when") or ()),
+                # A bare string is wrapped, not iterated -- a connector author writing
+                # ``visible_when="advanced"`` would otherwise char-splat into
+                # ``("a","d","v",...)``.
+                visible_when=(
+                    (data["visible_when"],)
+                    if isinstance(data.get("visible_when"), str)
+                    else tuple(data.get("visible_when") or ())
+                ),
                 out=_enum(ConfigOutput, data.get("out"), ConfigOutput.CONFIG),
             )
         except (KeyError, ValueError, TypeError) as exc:
--- a/kamiwaza_sdk/connectors/config_schema.py
+++ b/kamiwaza_sdk/connectors/config_schema.py
@@ -1,15 +1,33 @@
-"""Declarative connector config contract -- the framework's ConfigDef analog.
+"""Declarative connector config contract.
 
 A connector declares the admin configuration it needs as a :class:`ConfigSchema`
-(a set of :class:`ConfigField`). The framework derives BOTH validation and the
-admin-form JSON Schema from that one declaration, and marks secret fields so the
-UI and secret store can protect them. So a connector author writes the config once
-and gets validation + UI for free -- the Kafka Connect ``ConfigDef`` idea adapted
-to Kamiwaza connectors.
+(an ordered set of :class:`ConfigField`). The platform derives the whole admin form,
+its validation, and a JSON Schema from that one declaration -- so the config UI is
+written once, *in the connector*, and core renders it generically with no
+connector-specific UI code.
+
+Each :class:`ConfigField` carries enough metadata for a faithful form:
+
+- a value ``type`` (incl. multi-value ``LIST``) and a write-only ``secret`` flag;
+- a ``label`` + ``description`` (help text) and a ``placeholder``;
+- ``group`` + ``order`` for layout and ``importance`` for prominence;
+- optional validation: a regex ``pattern`` (+ ``pattern_error``) and ``required``;
+- recommended/allowed ``options`` -- each with its own label -- which render a
+  select (single value) or a multi-select (a ``LIST`` field);
+- a static ``default`` or a ``default_template`` resolved by the renderer
+  (e.g. ``"{origin}/api/connectors/public/{connector_type}/callback"``);
+- a simple visibility rule (``depends_on`` + ``visible_when``) so a field can show
+  only when another field has a given value.
+
+A field's value routes to the connector ``config`` by default, or to the OAuth
+``scopes`` set when ``out`` says so -- which is how a connector offers selectable
+permissions (the admin ticks capabilities; the ticked values become the requested
+scopes) without any of that living in core.
 """
 
 from __future__ import annotations
 
+import re
 from dataclasses import dataclass
 from enum import Enum
 from typing import Any
@@ -18,26 +36,86 @@ from .exceptions import InvalidConfigException
 
 
 class ConfigType(str, Enum):
-    """The value type of a config field (kept small and JSON-Schema-mappable)."""
+    """The value type of a config field."""
 
     STRING = "string"
     INTEGER = "integer"
     BOOLEAN = "boolean"
+    LIST = "list"  # multi-value; rendered as a multi-select when ``options`` are given
+
+
+class Importance(str, Enum):
+    """How prominently the admin form should surface a field."""
+
+    HIGH = "high"
+    MEDIUM = "medium"
+    LOW = "low"
+
+
+class FieldWidth(str, Enum):
+    """A hint for how wide the input should render."""
+
+    SHORT = "short"
+    MEDIUM = "medium"
+    LONG = "long"
+
+
+class ConfigOutput(str, Enum):
+    """Where a field's value goes when the admin saves the form."""
+
+    CONFIG = "config"  # -> the connector's stored config (the default)
+    SCOPES = "scopes"  # -> the OAuth scopes requested (selectable permissions)
 
 
 _PY_TYPES: dict[ConfigType, type] = {
     ConfigType.STRING: str,
     ConfigType.INTEGER: int,
     ConfigType.BOOLEAN: bool,
+    ConfigType.LIST: list,
 }
+
+
+def _humanize(name: str) -> str:
+    """A default display label from a field name (``client_id`` -> ``Client Id``)."""
+    return name.replace("_", " ").replace("-", " ").strip().title()
+
+
+@dataclass(frozen=True)
+class ConfigOption:
+    """One recommended/allowed value for a field (a select / multi-select choice)."""
+
+    value: str
+    label: str = ""
+    description: str = ""
+    default_selected: bool = False
+
+    def to_dict(self) -> dict[str, Any]:
+        return {
+            "value": self.value,
+            "label": self.label or self.value,
+            "description": self.description,
+            "default_selected": self.default_selected,
+        }
+
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> ConfigOption:
+        return cls(
+            value=str(data["value"]),
+            label=data.get("label", ""),
+            description=data.get("description", ""),
+            default_selected=bool(data.get("default_selected", False)),
+        )
 
 
 @dataclass(frozen=True)
 class ConfigField:
-    """One admin-config field a connector declares.
+    """One admin-config field a connector declares, with its full rendering metadata.
 
-    ``required`` with no ``default`` means the admin must supply it. ``secret``
-    marks credential material (rendered write-only, routed to the secret store).
+    ``required`` with no ``default`` / ``default_template`` means the admin must
+    supply it. ``secret`` marks credential material (rendered write-only, routed to
+    the secret store). ``options`` make the field a select (or a multi-select when
+    ``type`` is ``LIST``). ``out`` routes a ``LIST`` field's selections to the OAuth
+    scope set instead of the config blob.
     """
 
     name: str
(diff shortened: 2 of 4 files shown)


m9e (dismissed):
# PR Review Report: feat(connectors): generalize the declarative config-form contract (ENG-7810)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/196
**Branch**: feat/declarative-config-form → develop
**Scope**: 3 files, +557/-70 (reviewed at HEAD `64c6dac`, unchanged through review)

Multi-engine review (4 reviewers returned; Kamiwaza endpoint was down). Every Critical/High below was **independently re-verified against the code at HEAD** by the synthesizer — including refuting one over-rated Critical.

---

## Critical Issues

_None confirmed._ Antigravity rated the scope-field validation gap "Critical" (arbitrary OAuth scopes); this is **refuted as Critical** — see L1. No issue rises to Critical after verification.

## High Priority Issues (verified, should fix before downstream connectors build on this)

| H1 | config_schema.py:301, 334-341 | **JSON-Schema LIST round-trip is lossy.** `to_json_schema()` emits a `LIST` field as `{"type":"array"}`, but `from_json_schema()._resolve_type` only maps `string/integer/boolean/list` and has no `"array"` key, so it falls back to `STRING`. A `LIST` field rebuilt from `config_schema` alone then rejects valid configs (`{"regions":["us"]}` → "config regions must be string"). This PR newly publishes `config_schema` in `to_manifest()`, so any consumer validating from it breaks. **Codex reproduced this empirically; I re-confirmed by reading the code.** | CX, AG | Correctness |
| H2 | config_schema.py:239-243 (+ `must_supply` 147-149) | **`validate()` ignores field visibility.** `depends_on`/`visible_when` are net-new, but `validate()` treats every required field as unconditionally required. A `required=True` field hidden by `visible_when` can never be satisfied — saving the controlling field in its "hidden" state always raises `missing required config: <field>`, making conditional **required** fields unusable. | CX | Business logic |
| H3 | config_schema.py:266-274 | **Manifest-supplied regex run with no compile-guard or timeout.** `re.search(field.pattern, value)` (a) raises a raw `re.error`/`PatternError` — not `InvalidConfigException` — on a malformed manifest pattern (`"(unclosed"`), so callers catching the documented exception see a 500; and (b) is a **ReDoS** surface: catastrophic backtracking reproduced (SP: `(a+)+$` × 28 chars → 15.6s; CL: × 30 chars → 62.5s). `pattern` originates from the connector manifest, `value` from admin input. Compile at ingest (catch `re.error` → `InvalidConfigException`), cap input length, and/or run under a timeout. | SP, CL, AG | Security / Robustness |
| H4 | connectors/__init__.py:18, 42-67 | **Public API omits the new option/scope types.** `__all__`/imports still expose only `ConfigField, ConfigSchema, ConfigType`. Connector authors — the documented consumers — must write `options=(ConfigOption(...),)` and `out=ConfigOutput.SCOPES`, but `ConfigOption`, `ConfigOutput`, `Importance`, `FieldWidth` are reachable only via the private `…connectors.config_schema` submodule (exactly what the test file does). Export them so the contract this PR ships is usable from the package root. | SP | API contract |

## Medium Priority (suggestions & robustness)

| M1 | config_schema.py:247, 256-265 | **`LIST` element types unvalidated.** `validate()` checks only `isinstance(value, list)`; with no `options`, `{"tags":["a",1,{}]}` passes, contradicting the emitted `items:{type:"string"}`. Validate element types against `items`. | CL, AG | Validation |
| M2 | config_schema.py:178-203 | **`from_spec` leaks raw exceptions on malformed manifests** — `data["name"]`→KeyError, `int(data.get("order",0))`→ValueError (and `TypeError` on explicit `"order":null`), `ConfigOption.from_dict` `data["value"]`→KeyError. Inconsistent with the otherwise-forgiving `_enum`; fail with a typed error. | CL, AG | Robustness |
| M3 | config_schema.py:286-320 | **No invariant test** asserts `config_schema` (JSON Schema) and `config_fields` (form spec) — both published by `to_manifest()` and consumed by separate core paths — stay consistent. Add a cross-consistency test to prevent drift. | SP | Test / Contract |
| M4 | config_schema.py:147-149 vs 343-353 | **Asymmetric "required wins over default."** `from_json_schema` deliberately zeroes a required field's default (its 2nd-commit purpose), but a directly-declared `ConfigField(required=True, default="x")` has `must_supply=False` and is dropped from JSON-Schema `required`. Document or guard the asymmetry. | CL | Consistency / Docs |

## Low / Nice-to-have

| L1 | config_schema.py:220-226 | **Scope-field validation** — `validate()` intentionally excludes `out=SCOPES` fields (they aren't stored in config; they route to OAuth scopes). **This is correct design** — SP & CL confirm it, with dedicated tests. AG's "Critical: arbitrary unvalidated OAuth scopes" is **over-rated**: the admin is already trusted and the external OAuth provider enforces its own scope allowlist. An optional `validate_scopes()` helper enforcing option-membership would be a nice-to-have, not a blocker. | AG (Critical → downgraded); SP/CL (intentional) | Design |
| L2 | config_schema.py:269 | `re.search` is unanchored (`pattern="example\.com"` matches `evil.com?example.com…`). Matches JSON-Schema `pattern` semantics, but the docstring only end-anchors — note the substring behavior. | SP, CL | Docs |
| L3 | config_schema.py:262-265 | `_value_errors` echoes raw bad values into the error string; validate element type and bound the count echoed. | SP, CL | Code quality |
| L4 | config_schema.py:260-263 | Options membership uses implicit `str(v)` coercion (`1` matches option `"1"`). Harmless today; make the stringify explicit. | CL | Code quality |

## Test Coverage Assessment

- **Coverage verdict**: Good for what it tests — **26 tests pass** (CL ran `uv run --with pytest python -m pytest tests/unit/test_connectors_config_schema.py` → 26 passed in 0.27s; SP independently confirmed 26 green; PR description undercounts at 23). Strong on validate happy/error paths, bool-not-int, collect-all-errors, option membership, scope exclusion, form-spec round-trip, JSON-safety, unknown-enum fallback.
- **Missing test cases** (map to the findings above): malformed-regex `pattern` → `InvalidConfigException` (H3); `from_json_schema(to_json_schema(...))` LIST round-trip (H1); `validate()` with a hidden required `visible_when` field (H2); `LIST` element-type rejection (M1); `from_spec` with missing `name`/non-int `order`/missing option `value` (M2); `to_manifest()` asserting both `config_schema` **and** the new `config_fields` keys are present + JSON-safe.

## Reviewer Verdicts

| Superpowers | Request Changes |
| Claude | Comment (leaning Request Changes) |
| Codex | Request Changes (2 × P1) |
| Kamiwaza | N/A — endpoint down (HTML, not JSON) |
| Antigravity | Request Changes (High risk) |

**Vote Tally**: Approve: 0 | Comment: 1 | Request Changes: 3 (out of 4 successful reviewers)

## Reviewer Agreement Matrix

| Category | Superpowers | Claude | Codex | Antigravity | Consensus |
|----------|-------------|--------|-------|-------------|-----------|
| Regex (ReDoS / `re.error`) | Critical | Critical | — | Med/Low | **High** (3 engines, 2 reproduced) |
| LIST round-trip → STRING | — | — | P1 (repro) | High | **High** |
| Conditional-required (`visible_when`) | — | — | P1 | — | **High** (verified) |
| Missing public exports | High | — | — | — | **High** (verified) |
| LIST element-type validation | suggestion | High | — | High | **Medium** |
| Scope-field validation | intentional | intentional | — | Critical | **Low** (AG over-rated) |

## Positive Observations

- Clean, immutable design — `frozen=True` dataclasses with tuple fields give hashable, faithful round-trips via `to_spec`/`from_spec`.
- Scope-vs-config routing (`out`) is cleanly separated: scope fields excluded from `validate()`, `to_json_schema()`, and `_config_fields()`, each with dedicated tests.
- `_enum` fallback makes manifest deserialization forward-compatible against unknown enum members (tested) — good for connector/core version skew.
- `bool`-is-not-`int` guard preserved and tested; `provider.to_manifest()` change is additive (`config_fields` alongside `config_schema`) and backward-compatible.
- Unusually thorough, accurate docstrings; the 2nd commit's "required wins over default" hardening in `from_json_schema` is well-reasoned.

## Overall Verdict

**Risk Level**: Medium
**Reviewer Consensus**: 3/4 reviewers Request Changes; the 4th (Claude) leans the same way.

The design is sound, well-tested, and backward-compatible, and this is a good foundation. But four **verified** defects should be fixed before the downstream google/m365/linear/hubspot/renderer PRs build on this contract: the JSON-Schema LIST round-trip silently corrupts list fields (H1, reproduced), required conditional fields can never be saved (H2), manifest regex has an uncaught-exception + ReDoS surface (H3, reproduced), and the option/scope types the contract requires aren't exported from the package root (H4). None are large fixes. AG's scope-validation "Critical" is explicitly **not** a blocker.

---

Legend: SP=Superpowers, CL=Claude, CX=Codex, KZ=Kamiwaza (down), AG=Antigravity

shkevin:
Thanks for the thorough multi-engine pass — all High + Medium addressed in `00cffd8`. Module at 100% coverage, 73 connector tests green, ruff clean.

| # | Resolution |
|---|---|
| **H1** LIST round-trip lossy | `from_json_schema` now maps `"array"` → `LIST` and rebuilds `options` from `enum`. Added `test_list_field_round_trips_through_json_schema` (the `{"regions":["us"]}` repro now passes). |
| **H2** visibility ignored in validate | `validate()` skips fields gated off by `depends_on`/`visible_when` via `_is_active()`, mirroring the renderer — a hidden required field is no longer falsely required. |
| **H3** unguarded manifest regex / ReDoS | regex now runs in `_pattern_errors`: `re.error` → typed `InvalidConfigException` (no 500), and an input-length cap (`_MAX_PATTERN_INPUT=1024`) rejects oversize input *before* matching so the catastrophic-backtracking case never runs. (Noted: a length cap, not a true regex timeout — flagged in-code if untrusted manifests ever ship patterns.) |
| **H4** new types not exported | `ConfigOption`, `ConfigOutput`, `Importance`, `FieldWidth` now in `connectors/__init__.py` `__all__` + imports; added `test_public_api_exports_the_form_types`. |
| **M1** LIST elements unvalidated | `validate()` rejects non-string `LIST` elements (matches emitted `items:{type:string}`). |
| **M2** `from_spec` leaks raw exceptions | wrapped → `InvalidConfigException` (covers missing `name`, non-int `order`, missing option `value`). |
| **M3** no consistency invariant | added `test_json_schema_and_form_spec_cover_the_same_fields`. |
| **M4** required-vs-default asymmetry | documented on `must_supply`. |
| **L2/L3** unanchored regex / unbounded echo | docstring note on substring semantics; bad-value echo bounded to 5. |
| **L1** scope-field validation | left as-is — agreed itʼs intentional (trusted admin + provider-enforced scopes). |

Re-requesting review when you have a moment.

kbot-ships (dismissed):
_Reviewing head `00cffd8` against `develop`._

## Bottom line
This PR generalizes the connector config-form contract: `ConfigField`/`ConfigSchema` gain rich form metadata (labels, groups, options, importance, regex validation, visibility rules, scope routing) and `provider.to_manifest()` now emits both a validation-faithful `config_schema` (JSON Schema) and an ordered `config_fields` form spec. The direction is right — moving the form declaration into the connector manifest and rendering it generically in core is a clean contract, the dataclasses are frozen/hashable/JSON-stable, validation fails-closed and aggregates errors, and the serialization round-trips are tested. The code is careful and the test suite is genuinely good. I have no correctness or security defect that rises to blocking; the findings below are edge-case hardening and documentation. **APPROVE with notes.**

## Findings

### 1. The `_MAX_PATTERN_INPUT` cap does not actually bound ReDoS — only flags it (worth fixing)
`config_schema.py:228-254`. The 1024-char cap meaningfully limits *input length* but does not prevent catastrophic backtracking: a pathological pattern like `(a+)+$` against ~1024 `a`s followed by a non-matching char still backtracks super-linearly and can hang the request thread, with no `re` timeout. The inline `ponytail:` comment is honest that this is "a length cap, not a regex timeout," and the test (`test_overlong_value_rejected_before_running_regex`, line 325) only exercises the 5000-char length rejection — it never demonstrates that a 1024-char malicious input is safe, because it isn't. Under the stated threat model (patterns come from *trusted* connector manifests, values from admins) this is an accepted risk, not an exploit surface. But the comment understates it: a buggy-but-benign manifest pattern plus normal admin input is enough to cause an accidental DoS. Recommend either lowering the cap substantially for pattern-validated fields, or running the regex under a worker/timeout, before any path lets untrusted manifests ship patterns. At minimum, tighten the comment to say the cap *reduces blast radius but does not eliminate backtracking blowup at 1024 chars*.

### 2. Visibility is evaluated against raw config, not defaults-applied config (worth fixing)
`config_schema.py:304-314`. `_is_active` reads `config.get(field.depends_on)` directly. If the controlling field carries a `default`/`default_template` and the admin omits it, `validate()` sees `None`, treats the dependent field as hidden, and skips its required check — even though the rendered form (which materializes the controller's default) would show the dependent field and require it. So a conditionally-required field gated by a *defaulted* controller can slip through validation. The existing visibility tests (`TestVisibility`, lines 294-315) use a controller with no default, so they don't catch this. Whether it bites depends on whether core applies defaults before calling `validate()` — that lives in the consuming core renderer PR. Worth either resolving the controller's effective value (default-substituted) inside `_is_active`, or documenting explicitly that `validate()` assumes defaults are already merged into `config` by the caller.

### 3. Boolean/non-string controller values rely on stringified comparison (nit)
`config_schema.py:314`. `str(config.get(field.depends_on)) in field.visible_when` means a boolean controller stores `True`/`False` but `visible_when` must contain the literal strings `"True"`/`"False"`. That's a reasonable convention but undocumented and easy to get wrong for a connector author. A one-line note in the `ConfigField` docstring on `depends_on`/`visible_when` (controller value is compared as a string) would prevent silent always-hidden fields.

## Test gaps
- **Malformed option in `from_spec`**: `test_from_spec_rejects_malformed_spec` (line 339) covers missing `name` and non-int `order`, but not a malformed `options` entry (e.g. `{"options": [{}]}` — `ConfigOption.from_dict` does `data["value"]`). It's inside the try/except so it *is* converted to `InvalidConfigException`, but nothing pins that behavior; a refactor moving option parsing outside the guard would leak a raw `KeyError` uncaught. Cheap to add.
- **Scope-field option validation**: a `SCOPES`-routed field's selected values are never validated against its `options` here (by design — they're not in the config blob). That contract boundary is asserted indirectly, but there's no test documenting that selected scope values are validated *somewhere* downstream. Worth a comment or a cross-PR note so the core renderer PR owns that check.
- **Visibility-with-default**: see finding #2 — no test exercises a dependent field whose controller is satisfied only via a default.

## What's good
Thorough, hypothesis-labeled tests (H1–H4, M1–M3) that map directly to the hardening decisions and would catch regressions if the fixes were reverted — the LIST↔array round-trip, the bool-is-not-int check, the typed-error-not-`re.error` conversion, and the `required`-wins-over-default rule are all pinned. `bool` exclusion from integer fields, the unanchored-`re.search` documentation, error aggregation, and forward-compatible unknown-key/unknown-enum tolerance are all handled deliberately. `__all__` is correctly expanded with no previously-exported symbol dropped.

## Verdict
**Verdict**: APPROVE
The contract is sound, serialization round-trips and validation are well-tested, and the remaining issues are documented edge cases (ReDoS cap framing, visibility-vs-defaults) rather than correctness or security defects that block merge.
---

m9e (commented):
# PR Review Report (Round 2 — re-review): feat(connectors): generalize the declarative config-form contract (ENG-7810)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/196
**Branch**: feat/declarative-config-form → develop
**Scope**: 4 files, +770/−88 · **HEAD reviewed**: `00cffd88e9`
**Delta since Round 1** (`64c6dac207` → `00cffd88e9`): one commit, *"address config-form review"* (+252/−57 across `config_schema.py`, `__init__.py`, tests)

---

## Round-1 outcome → Round-2 status

Round 1 was **Request Changes** over 4 High + 4 Medium findings. I verified every one against HEAD (traced the code path + ran the suite). **All eight are resolved.**

| R1 Finding | Status at `00cffd88e9` | Evidence |
|---|---|---|
| **H1** LIST↔`array` round-trip lossy (`{"regions":["us"]}` rejected) | **Fixed** | `type_by_value["array"] = ConfigType.LIST` + new `_options()` rebuilds options from `enum`/`items.enum` (`config_schema.py:388,405-411,424`). New test `test_list_field_round_trips_through_json_schema`. |
| **H2** `validate()` ignored visibility → required hidden field always rejected | **Fixed (common case)** | `validate()` now skips `not _is_active(...)` fields; `_is_active` mirrors the renderer's `visible_when`/`depends_on` rule (`:284-287,304-314`). See M-new-1/2 for residual edges. |
| **H3** unguarded manifest regex (`re.error`→500) + ReDoS | **Fixed (guard) / Mitigated (ReDoS)** | `_pattern_errors` wraps `re.search` in `try/except re.error` → typed `InvalidConfigException`, no raw 500 (`:236-249`). ReDoS bounded by 1024-char input cap — a length cap, not a timeout (honest `ponytail:` in-code note). |
| **H4** `__all__` omitted `ConfigOption/ConfigOutput/Importance/FieldWidth` | **Fixed** | All four imported + added to `__all__`, which is now a strict superset (24→28, nothing dropped — AST-confirmed) (`__init__.py:18-26,48-78`). |
| **M1** LIST element-types unvalidated | **Fixed** | `_value_errors` rejects non-string LIST elements ("must be a list of strings") (`:316-322`). |
| **M2** `from_spec` leaked KeyError/ValueError | **Fixed** | Body wrapped → `InvalidConfigException`; `order` hardened `int(... or 0)` (`:204,216-217`). |
| **M3** no `config_schema`↔`config_fields` invariant test | **Fixed** | `test_json_schema_and_form_spec_cover_the_same_fields`. |
| **M4** required-vs-default asymmetry | **Documented** | `must_supply` docstring + `from_json_schema` drops default for required field. |

**Tests**: `35 passed` (I ran `uv run --with pytest python -m pytest tests/unit/test_connectors_config_schema.py` independently). New tests map 1:1 to H1–H4 / M1–M3.

The Round-1 blockers are gone. The items below are **non-blocking fast-follows**, several surfaced by the new visibility/options logic.

---

## Residual findings (non-blocking)

| # | File:Line | Issue | Flagged By | Sev |
|---|-----------|-------|-----------|-----|
| 1 | config_schema.py:304-314 | `_is_active` reads **raw** config, not defaults-applied. A conditionally-required field gated by a **defaulted** controller the admin omits → `validate()` sees `None`, treats it hidden, skips the required check, even though the rendered form (defaults materialized) shows it required. **Fail-open** under-validation, conditional on whether core merges defaults before `validate()` (the core renderer PR owns that). | CX(P2), AG, CL, SP, kbot | Medium |
| 2 | config_schema.py:314 | `str(config.get(depends_on)) in visible_when` mis-handles non-STRING controllers: a **bool** controller needs literal `"True"`/`"False"` in `visible_when`; a **LIST/multi-select** controller stringifies to `"['x']"` and never matches; a **SCOPES**-routed controller isn't in the config blob at all. All fail-open edges; common STRING-controller case is correct. | CL, AG, SP | Low-Med |
| 3 | config_schema.py:356-365 | `to_json_schema` emits `enum` as **strings** for an INTEGER/BOOLEAN field carrying `options` (`ConfigOption.value` is `str`-typed) → JSON Schema whose `enum` types don't match the property `type`. **Pre-existing** (delta did not touch `to_json_schema`); niche (typed selects). Constrain `options` to STRING/LIST, or cast enum to the field's primitive type. | CX(P2), AG | Medium |
| 4 | config_schema.py:225-249 | ReDoS: the `re.error`→typed-exception half is **fully fixed**. The 1024-char `_MAX_PATTERN_INPUT` cap **mitigates but does not eliminate** catastrophic backtracking — reviewers reproduced `(a+)+$` hanging at ~28–40 chars, well under the cap. **Accepted** under the current threat model (patterns from first-party/reviewed connector manifests, values from authenticated admins) and self-documented in-code. Replace with a real regex timeout before any path lets *untrusted* manifests ship patterns. | SP, AG, CL, CX, kbot | Low (accepted) |

**Recommended for this PR**: pick one of {resolve effective default inside `_is_active`} or {document that `validate()` assumes a defaults-merged config and let the core renderer PR own merging} for #1, and add a one-line docstring note for #2's string-comparison convention. #3 and #4 are fine as tracked follow-ups.

---

## Reviewer Verdicts (Round 2)

| Reviewer | Verdict | Note |
|----------|---------|------|
| Superpowers | Comment | H1/H2/H4/M1/M2 verified fixed; H3 ReDoS High but gated behind trusted manifests |
| Claude | Approve | All R1 items fixed; residuals are fail-open edges, none security-critical |
| Codex (gpt-5.4) | Comment | Confirmed all R1 fix-claims resolved; only 2×P2 (visibility-default, typed-enum) |
| Kamiwaza | N/A | Endpoint down (returned HTML) — excluded |
| Antigravity | Request Changes | Risk High; rates #1/#3 "Critical" (over-rated vs CX P2 / consensus Medium) |

**Vote tally (N=4)**: Approve 1 · Comment 2 · Request Changes 1. No approve-supermajority; no Critical/High survives synthesis (AG's two "Criticals" are fail-open/pre-existing Mediums per CX/CL/SP consensus).

## Overall Verdict

**Risk Level**: Low–Medium
**Recommendation**: Comment
**Reviewer Consensus**: Round-1 blockers fully resolved and test-verified; merge-readiness is high. Landing as **Comment** (not Approve) only because a genuinely convergent, still-open **Medium** — `_is_active` validating against raw rather than defaults-applied config (#1) — deserves an explicit decision (fix here vs. own it in the core renderer PR), and Kamiwaza being down leaves no approve-supermajority. Nice work turning the round-1 list around cleanly.

---

Legend: SP=Superpowers, CL=Claude, CX=Codex, KZ=Kamiwaza, AG=Antigravity, kbot=kbot-ships autoreview

shkevin:
Thanks for the round-2 re-review — addressed the two you flagged for this PR in `7a73aae` (100% module coverage, 60 tests).

- **Residual #1** (`_is_active` validates raw config) — fixed by resolving the controllerʼs declared **default** inside `_is_active`: when the controller is absent from `config`, it falls back to the value the form materializes, so a field gated on a defaulted controller is validated (fail-closed). New test `test_defaulted_controller_still_activates_a_gated_field`.
- **Residual #2** (string-comparison convention) — documented on `_is_active`: `visible_when` lists controlling values verbatim (a bool controller uses `("True",)`), and LIST/SCOPES controllers arenʼt supported as gates.

- **#3** (typed `enum` for INTEGER/BOOLEAN options) and **#4** (ReDoS — `re.error` guard done, length-cap mitigation accepted) — taking these as the tracked follow-ups you suggested; #3 is pre-existing + niche (no connector declares typed selects today) and #4ʼs real fix is a regex timeout if untrusted manifests ever ship patterns.

Round-1 + the round-2 actionable both cleared.

m9e (commented):
# PR Review Report (Round 3 — delta re-review): feat(connectors): generalize the declarative config-form contract (ENG-7810)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/196
**Branch**: feat/declarative-config-form → develop
**Scope**: 4 files, +801/-88 (delta since last review = **1 commit**, `7a73aae38b`, +29/-4 across 2 files)
**Reviewed HEAD**: `7a73aae38bb9b86ce5ebfeb87f1457b5cf8dc599`
**Last reviewed (R2)**: `00cffd88e9` → this round reviews the delta `00cffd88e9..HEAD`

---

## Delta summary — what changed since R2

One commit: **`7a73aae38b` "resolve controller default in _is_active (review round)"**. It targets the single substantive residual left open in Round 2:

> R2 residual #1 (convergent Medium, fail-OPEN): `_is_active()` read the controller's value straight from the supplied config. A conditionally-**required** field gated by a controller that has a **default** — but which the admin omits — slipped past `validate()` (controller read as `None` → `str(None)` ∉ `visible_when` → field deemed inactive → never required).

**The fix (verified):** `_is_active` is now an instance method; when the controller (`depends_on`) is absent from config it resolves the controller field's declared `default` before the `str(raw) in visible_when` comparison. This moves the behavior **fail-CLOSED** — it can only ever make *more* fields required, never fewer. Docstring updated to document the string-compare contract (`("True",)` for booleans) and the default-fallback semantics.

### Verification performed at HEAD
- **Tests: 36 passed** (`uv run --with pytest python -m pytest`), up from 35 at R2; the new `test_defaulted_controller_still_activates_a_gated_field` would FAIL on pre-delta code, so it genuinely exercises the fix.
- **No regression from `@staticmethod` → instance method:** repo-wide search finds exactly two references — the definition (L304) and the sole call site (L286), which already used `self._is_active(field, config)`. No `ConfigSchema._is_active(field, config)` static-style call exists anywhere. (`_is_active_deployment` in `kamiwaza_extensions_lib` is an unrelated function.)
- **Edge-case walkthrough (all sensible, none crash):** boolean default (`str(True)=="True"`, matches documented `("True",)`); explicit `{ctrl: null}` coalesces to default (fail-closed, safe); controller with no default → inactive (expected); missing/typo'd controller → graceful, no crash; LIST default → won't match (pre-existing documented string-compare contract).

**All eight R1 blockers (fixed + verified at R2) remain fixed at HEAD. The one R2 residual is now resolved.**

---

_None._ No reviewer raised a Critical at HEAD. The delta introduces no Critical/High finding.

## High Priority Issues

_None genuinely blocking in the delta._ Antigravity self-rated two items "High", but both are **standing pre-existing residuals on code untouched by this delta**, already adjudicated in R1/R2:

| # | File:Line (HEAD) | Issue | Flagged By | Synthesized severity |
|---|------------------|-------|------------|----------------------|
| 1 | config_schema.py:246 (`_pattern_errors`) | ReDoS: 1024-char cap ≠ regex timeout; `^(a+)+$` can still backtrack | AG | **Low / accepted** — patterns come from first-party reviewed manifests; documented in-code (`ponytail:` note). Not a delta item. |
| 2 | config_schema.py:~358 (`to_json_schema`) | INTEGER/BOOLEAN field carrying `options` emits `enum` as **strings** → type/enum mismatch for strict validators | AG (and CX-P2 in R2) | **Medium, pre-existing** — `to_json_schema` is untouched by this delta. Legitimate type-fidelity fast-follow. |

## Medium Priority (pre-existing / cross-PR boundary — not delta regressions)

| # | File:Line | Issue | Flagged By | Notes |
|---|-----------|-------|------------|-------|
| 1 | config_schema.py:147-155 (`must_supply`) & `_is_active` default-resolution | `default_template`-only fields are treated as satisfied (`must_supply=False`); `_is_active` resolves only static `default`, not `default_template`. A direct (non-UI) SDK/API caller could persist config missing a templated value, and a field gated on a *templated* controller would read inactive. | CX (P1), SP (Low) | **By-design boundary, pre-existing.** Template **materialization is the core renderer/platform's job** (the consuming PR), same seam as R2 residual #1. SP: "theoretical — `default_template` is used for URL/callback values, not gating selectors." Worth a cross-PR contract note; **not a blocking defect in this contract lib and not a delta regression.** |
| 2 | config_schema.py:326 (`_is_active`) | `str(raw)` couples gating to Python boolean capitalization (`"True"`/`"False"`); a JS renderer stringifies booleans to `"true"`/`"false"`. | AG (Medium) | Touches the delta, but the convention is now **explicitly documented** (`("True",)`). The renderer (separate core PR) must apply the same rule — a cross-component **contract-alignment** item, not a bug here. |
| 3 | config_schema.py (`from_json_schema`, `items` access) | `(prop.get("items") or {}).get("enum")` assumes `items` is a dict; JSON-Schema tuple-validation (`items: [...]`) would raise `AttributeError`. | AG (Medium) | Pre-existing; the SDK only ever emits `items: {type: string}`. Low probability. Not a delta item. |
| 4 | config_schema.py:263-279 (SCOPES) | `out=SCOPES` fields are excluded from `validate()`/`to_json_schema()`, so programmatic callers can submit arbitrary `scopes`. | CX (P1), AG (in R1) | **Refuted/by-design (R1):** scope-routed values are not part of the config blob; the external OAuth provider enforces the scope allowlist and admins are trusted. Pre-existing, not a delta item. |

## Suggestions (Nice to Have)
- [config_schema.py:319] Document that an explicit `null` controller value is intentionally coalesced to the controller's default (fail-closed). (SP, CL — Low)
- [config_schema.py:179] `from_spec` could wrap a bare `visible_when` string into a 1-tuple to prevent the char-splat authoring footgun. (AG — Low)

## Test Coverage Assessment
- **Coverage verdict: Adequate.** 36/36 pass at HEAD. The delta ships a targeted test that fails on pre-delta code and asserts both the fail-closed raise (`validate({})`) and the satisfied path (`validate({"advanced_token": "t"})`).
- Optional add: a test for a **boolean** defaulted controller (`default=True`, `visible_when=("True",)`) to lock the documented capitalization contract.

## Reviewer Verdicts

| Superpowers | Approve (Risk: Low) |
| Claude | Approve (Confidence: High) |
| Codex | Comment (two P1s, both pre-existing/by-design; nothing on the delta) |
| Kamiwaza | N/A — endpoint down (returned HTML, not JSON) |
| Antigravity | Request Changes (Risk: High) — but both "Highs" are pre-existing/accepted residuals on untouched code |

**Vote Tally**: Approve: 2 | Comment: 1 | Request Changes: 1 (out of N=4 successful reviewers; Kamiwaza excluded)

## Positive Observations
- Minimal, surgical fix (~25 lines) scoped exactly to the one residual raised last round; no collateral changes.
- The fix moves strictly in the **safe (fail-closed)** direction — it can only add required-field checks, never remove them.
- Docstring updated to document the string-compare contract and default-resolution semantics — pre-empts the next reviewer re-litigating the same edge.
- New test genuinely covers the fix (fails on pre-delta code) and asserts both directions.

## Overall Verdict

**Risk Level**: Low
**Recommendation**: Comment
**Reviewer Consensus**: No Critical/High defect remains in the delta; the R2 residual is resolved and verified (36/36 tests pass, no regression). This lands as **Comment** rather than Approve only because there is no approve-supermajority (Kamiwaza has been down every round, capping N at 4, and the remaining items are standing **pre-existing / cross-PR-boundary** Mediums — template materialization and the boolean gating contract — owned by the core renderer PR, not by this contract library).

**From this reviewer: no remaining blocking requests.** Every R1 blocker and the R2 residual are resolved and verified at HEAD. The open items are non-blocking, pre-existing, and belong to the consuming renderer PR; they should be tracked as fast-follows there rather than gating this foundational contract PR.

---

shkevin:
@m9e re-review please — develop ruleset needs an approving review and your round-3 delta re-review already found no remaining blocking requests (all R1 blockers + the R2 residual resolved at HEAD 7a73aae, 36/36 tests). You approved the sibling core PR #2186 at the same time; requesting the same here so this can merge and unblock the connector PRs.

m9e (commented):
# PR Review Report (Round 4 — re-review): generalize the declarative config-form contract (ENG-7810)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/196
**Branch**: feat/declarative-config-form → develop
**HEAD reviewed**: `6447e74e897b72406062ad90eec592f084736956`
**Delta since last review (R3 `7a73aae38b`)**: 1 commit, `kamiwaza_sdk/connectors/config_schema.py` +9/−1 and a new test, +9.

History: **R1 = Request Changes** (since dismissed) → **R2 = Comment** (all 8 R1 items fixed) → **R3 = Comment** (controller-`default` fail-closed fix) → **R4 = this review.**

---

## Delta under review

Single commit — `fix(connectors): wrap a bare visible_when string in from_spec`:

`ConfigField.from_spec` previously did `visible_when=tuple(data.get("visible_when") or ())`. For a **hand-authored manifest** that sets `visible_when="advanced"` (a bare string rather than a list), `tuple("advanced")` char-splatted into `("a","d","v","a","n","c","e","d")`. The fix wraps a bare string into a 1-tuple `("advanced",)`; non-string inputs still flow through `tuple(... or ())`.

```python
visible_when=(
    (data["visible_when"],)
    if isinstance(data.get("visible_when"), str)
    else tuple(data.get("visible_when") or ())
),
```

A new regression test, `test_from_form_spec_wraps_a_bare_visible_when_string`, asserts `field.visible_when == ("advanced",)`.

### Why this is a real (if minor) fix
The char-splat silently broke gated fields authored with a bare string. `_is_active` matches via `str(raw) in field.visible_when`; against `("a","d",...)` a real controller value like `"advanced"` is never a single character, so the gated field was **never active** → its `required` check was skipped (fail-OPEN) and the form never rendered it. The fix restores exact-match (fail-closed) semantics and aligns the single-string author case with the list case. `_is_active`'s own docstring already uses `("advanced",)` as the canonical form, so the delta makes the parser match the documented contract.

### Verification (run at HEAD in an isolated clone)
- **`37 passed`** for `tests/unit/test_connectors_config_schema.py` (R3 was 36; +1 new). 
- **Regression coverage proven**: restoring the pre-delta production code and running only the new test **fails** with `('a','d','v',...) != ('advanced',)`; restoring HEAD makes it pass. The test genuinely guards the bug.
- **Edge cases confirmed sound** (verified by SP + CL): `None`/missing → `()`; list/tuple → preserved; `int`/non-iterable → still raises the typed `InvalidConfigException` via the existing `except (KeyError, ValueError, TypeError)` (typed-exception guarantee preserved); empty string `""` → `("",)` (faithful and usable, since `_is_active` can match a `""` controller value).
- **Round-trip symmetry intact**: `to_spec`/`to_dict` emits `list(self.visible_when)`, so the serialize→deserialize path never feeds `from_spec` a bare string — the new branch is reached only by hand-authored manifests, exactly its intended audience.

**No Critical or High finding on the delta from any reviewer.**

---

## High Priority Issues (2+ reviewers or single high-confidence)

None on the delta.

## Medium / Pre-existing residuals (not introduced or changed by this delta — tracked, non-blocking)

| # | File | Issue | Flagged By | Status |
|---|------|-------|------------|--------|
| 1 | config_schema.py (`must_supply`/`to_json_schema`) | A `required=True` field backed only by `default_template` is dropped from JSON-Schema `required`; a consumer reconstructing validation from `config_schema` alone (not the rich `config_fields`) won't enforce it. | CX (P1), AG | **Pre-existing / by-design boundary.** SDK does not materialize templates; the **core renderer PR** owns template materialization. Convergent with R3's tracked residual. Theoretical for non-UI callers. Fast-follow in the renderer PR, not this SDK delta. |
| 2 | `_is_active` (controller default / visibility) | `_is_active` resolves an omitted controller's static `default` but not `default_selected` options or `default_template`; and it is non-recursive (a hidden controller with stale payload data can still gate a child). | CX (P2), AG | **Pre-existing.** Unchanged by the delta. Form renderer drives visibility; same renderer-PR boundary. Fail-closed for the static-default case fixed in R3. |
| 3 | `to_json_schema` enum / `ConfigOption.value` | `ConfigOption.value` is str-typed; an INTEGER/BOOLEAN field carrying `options` emits enum as strings → enum/type mismatch. | AG (rated Critical) | **Pre-existing Medium** (R2 residual #3; `to_json_schema:356-365` untouched by every delta in this PR). Not a delta regression. |
| 4 | `_is_active` matching | LIST/multi-select controllers (`str(["x"]) in visible_when`) and non-string `visible_when` elements (`visible_when=[True]`) don't match. | AG (High) | **Pre-existing / by-design.** Documented contract: controllers compared via `str(raw)`, authors write the string form (`("True",)`). The delta changed only bare-string *parsing*, not the matcher. |
| 5 | `from_json_schema` `items` handling | `(prop.get("items") or {}).get("enum")` assumes `items` is a dict; a JSON-Schema tuple-validation array would `AttributeError`. | AG (High) | **Pre-existing**, untouched by delta. Manifests are first-party/reviewed. Optional robustness fast-follow. |
| 6 | regex `_pattern_errors` | 1024-char cap mitigates but does not eliminate catastrophic-backtracking ReDoS. | AG (High) | **Accepted residual** (R1–R3). Patterns come from first-party reviewed manifests; admin-supplied values. Cap ≠ timeout, acknowledged in code. |

> All six were raised and adjudicated in R1–R3. None is a regression introduced by the R4 delta. Antigravity's line anchors (`:192`, `:236`, `:255`) do not match HEAD — it read the diff hunks, not the final file (consistent with prior rounds).

## Suggestions (nice-to-have, non-blocking)

- `config_schema.py:216-220` — `data.get("visible_when")` is read up to three times; hoisting to a local `vw = data.get("visible_when")` is marginally cleaner. The current form is correct (the `isinstance` guard guarantees the subscript can't `KeyError`). *(SP, CL — Low)*
- A `parametrize` adding `None` / list-passthrough / `""` cases would lock in the full branch behavior, though the targeted regression test is the load-bearing one. *(SP, CL — Low)*

## Test Coverage Assessment

- **Coverage verdict**: Adequate. The new test exercises the real public entry point (`from_form_spec` → `from_spec`) and provably catches the regression. 37/37 pass at HEAD.

## Reviewer Verdicts

| Reviewer | Individual Verdict | Notes |
|----------|--------------------|-------|
| Superpowers | **Approve** (Risk Low) | Minimal, surgical, correct; only cosmetic nice-to-haves. |
| Claude | **Approve** (Confidence High) | All edge cases verified sound; no regression vs R3. |
| Codex | **Comment** | Zero findings on the delta; 2 pre-existing P1/P2 boundary residuals owned by the renderer PR. |
| Kamiwaza | N/A — Error (endpoint down) | HTML not JSON; excluded from N (consistent with R1–R3). |
| Antigravity | **Request Changes** (Risk High) | Over-rated; every cited item is a pre-existing/by-design/accepted residual, none on the delta; line numbers off (read the diff, not HEAD). |

**Vote Tally** (N=4): Approve 2 | Comment 1 | Request Changes 1.

## Positive Observations

- Minimal, surgical, well-documented fix with a clear "why" comment naming the char-splat footgun.
- Genuine regression test that fails on the pre-delta code and passes at HEAD.
- Typed-exception, round-trip, and fail-closed guarantees all preserved.

## Overall Verdict

**Risk Level**: Low
**Recommendation**: Comment
**Reviewer Consensus**: No reviewer found a blocking defect in the delta. The R4 change is a correct, well-tested bug fix and I consider the PR **merge-ready**. This is a **Comment** rather than Approve purely on the no-approve-supermajority math — the Kamiwaza endpoint has been down across all four rounds (caps N at 4) and Antigravity has voted Request Changes every round on pre-existing/by-design residuals, so the two substantive Approves (Superpowers, Claude) land at 2/4 rather than a majority. I have **no remaining blocking requests**; the residuals above are pre-existing, by-design boundary items owned by the core renderer PR, suitable as fast-follows.

---

PR #230, opened 2026-07-13
fix(extensions): kz-ext dev builds owned images into the cluster registry (ENG-8626)

jonathanleff:
## Summary
`kz-ext dev` built, pushed, and deployed **owned** build images to their declared publish registry (GHCR) instead of the resolved cluster dev registry; this splits the shared canonical-ref policy with a required `purpose="dev" | "publish"` so dev targets the cluster registry while publish keeps preserving the declared namespace.

## Type
- [x] Bug Fix - Corrects existing behavior

## Change Flags
- [ ] API change
- [ ] Configuration change
- [ ] Requirements change
- [ ] Schema change

(Internal SDK signature change only — see Breaking Changes.)

## Breaking Changes
- [x] Internal interface - Other services/modules need coordinated updates

`compute_canonical_refs`, `_canonical_build_ref`, and `ComposeTransformer.transform` / `transform_service` now take a **required** `purpose` keyword, and `dev_env_image_refs.build_image_ref_map` drops four now-unused params. All are internal to `kamiwaza_extensions`; every in-repo call site is updated. `purpose` is deliberately required rather than defaulted — a call site that forgets it should fail loudly, because *silently inheriting publish semantics is exactly this bug*.

No change to `kz-ext publish` behavior (verified byte-identical, below).

## Motivation
`_canonical_build_ref`'s "declared ref is registry-qualified → preserve namespace, rewrite only the tag" early return **is** the entire destination policy, and it was shared by dev and publish.

That is correct for **publish** — an extension may publish beneath an unconventional namespace (Omniparse, ENG-4909), and catalog refs must match the pushed images. It is wrong for **dev**, where a qualified ref on a service with `build:` is the image's *published identity*, not where this cluster pulls from.

For Kaizen the resulting dev batch was mixed:

| Service | Before | After |
| -- | -- | -- |
| `sandbox-controller`, `backend`, `frontend` | `ghcr.io/.../images/kaizen-*:<dev-tag>` | dev registry |
| profile-gated `agent` | `host.docker.internal:5001/kaizen-agent:<dev-tag>` | dev registry (same repo path as siblings) |
| `postgres` (no `build:`) | declared ref, untouched | unchanged |

The controller pushes first, so the command died on a GHCR write **before ever deploying**, and local cluster development depended on an external private registry plus credentials. `build_push_ref_map` could not repair it: it only rewrites refs already prefixed with `image_registry`, so a `ghcr.io` ref never matched and was pushed verbatim.

Introduced by `c3ddc4c8` (ENG-4909), which made `compute_canonical_refs` the shared source of truth for both commands — unifying two genuinely different destination policies. Latent for Kaizen since its May 15 conversion to the flattened `kz-ext` layout.

## Source
- [x] Issue/Ticket: ENG-8626

## Alternatives Considered
- **Fix it in `build_push_ref_map`** (rebase any foreign ref onto the push registry). Rejected: the transformed compose / CR payload would still say `ghcr.io`, so the cluster would pull from GHCR — ImagePullBackOff. The deploy ref and the push ref must move together.
- **Relocate after `compute_canonical_refs`, inside `run_dev_remote` only.** Rejected: `transform_service` derives the CR's `image:` independently via the same helper, so dev would end up with two competing sources of truth for a service's ref — precisely the drift the existing docstrings were written to prevent.
- **Flatten dev repos to `{ext}-{svc}`.** Rejected: collides when two declared repos share a basename, and discards the declared repository identity. Host-replacement preserves the path, stays collision-free, and is idempotent.

## Changes Made
- `compose_transformer.py` — `Purpose` literal; required `purpose` on `compute_canonical_refs` / `_canonical_build_ref` / `transform` / `transform_service`. Dev swaps the registry host and **preserves the declared repository path** (reusing the existing `_split_image_ref`). Dev also **includes profile-gated build services** in the map; publish still excludes them (they publish via `extra_docker_images`).
- `commands/dev.py` — `purpose="dev"` at all three call sites (transform, canonical refs, resume probe). `--no-build` now **explicitly** filters profiled services out of the push list.
- `commands/publish.py` — `purpose="publish"` at all four call sites. No behavior change.
- `dev_env_image_refs.py` — `build_image_ref_map` reads the canonical map directly instead of re-deriving a legacy fallback for profiled services, removing the 4th site that had to stay in lockstep with `ImageBuilder`.

## Technical Notes
- **The `--no-build` exclusion became load-bearing.** Profiled services were kept out of the push list only *because* they were absent from `canonical_refs`. Now that dev includes them, that exclusion had to become an explicit filter — otherwise `--no-build` would try to push an agent image it never built, and a resumed run may no longer hold it locally. Pinned by a test.
- **The agent's repo path changes** from flat `.../kaizen-agent` to the nested declared path, alongside its siblings. `AGENT_SERVER_IMAGE` and the allowlist follow by construction; Kaizen's controller derives its allowlist from `AGENT_SERVER_IMAGE` by prefix (ENG-6663) and is explicitly depth-agnostic — verified in the live pod (below).
- **`build_push_ref_map` stays transport-only** — and now actually prefix-matches, so a split image/push (VM alias) registry works instead of silently no-op'ing.
- Unqualified / blank `image:` declarations keep the legacy `{registry}/{basename}-{svc}` fallback under **both** purposes, so VM-alias, insecure-registry, offline-relocation and `image_basename` behavior is untouched.

## Testing
- [x] Unit tests added/updated
- [x] Manual testing performed

## Verification
| Environment | Steps Performed | Result |
|-------------|-----------------|--------|
| Local | `uv run pytest -m "not integration and not live and not e2e"` | **2869 passed**, 1 skipped (extension baseline was 1649; +25 new) |
| Local | `ruff` + `mypy --follow-imports=silent` on all changed files | Clean (the 7 mypy errors in `dev.py`/`publish.py` are pre-existing on `develop`) |
| Local | Publish-parity: diffed `compute_canonical_refs(purpose="publish")` + transform output vs `develop` for Kaizen / omniparse / unqualified shapes | **Byte-identical** |
| Dev cluster (kamiwaza.test, Podman 5.7.1, registry `host.docker.internal:5001`) | `kz-ext dev` from `kaizen-v3` @ `493449f9` | **EXIT=0.** Zero `ghcr.io` references in the entire run |
| Dev cluster | All 4 owned images pushed at one revision | `{dev-registry}/kamiwaza-internal/kamiwaza-extensions-kaizen/images/kaizen-{controller,agent,backend,frontend}:2.0.2-dev-493449f9.…` |
| Dev cluster | Deployed pod images vs pushed refs | Match exactly; `postgres` still pulls `ghcr.io/.../postgres:v18.4` (external, untouched) |
| Dev cluster | Kaizen CR | `Running` / Ready=`True`, all 4 pods 2/2, no ImagePullBackOff |
| Dev cluster | Agent sandbox path: pulled `AGENT_SERVER_IMAGE` in `kamiwaza-sandboxes` | Pod reached `Running` — cluster pulls the relocated agent image |
| Dev cluster | `resolve_allowed_image_prefixes()` **inside the live controller pod** | Derives the nested repo prefix, **admits** the agent image, still **rejects** a same-prefix sibling (`kaizen-agent-evil`) |

## Test Coverage
Dev policy (qualified ref relocated beneath `image_registry`; repository path preserved; idempotent; digest-pinned refs; same-basename repos don't collide), publish policy unchanged, external `build:`-less services untouched, profile-gated agent included in the dev map, one-ref-everywhere (builder == pusher == deployed compose == CR), `AGENT_SERVER_IMAGE` + allowlist match the pushed agent ref, split image/push registry alters only the transport alias, `--no-build` still refuses to push profiled services, resume probes target the rebased refs.

Revised (these encoded the bug): `test_divergent_image_namespace_flows_through_to_builder`, the two "external refs are not retagged locally" preflight tests (they now exercise the alias-transport path), and `test_normal_service_keeps_declared_namespace`.

## Review Focus
1. **`_canonical_build_ref`'s dev branch** — is host-replacement-with-path-preservation the repository layout we want long-term? (Ticket called the layout an explicit design choice.)
2. **The `--no-build` profiled filter in `dev.py`** — the one prior behavior that had to be preserved by hand.
3. **`purpose` being required** — intentional, but it does touch every call site.

## Deployment Notes
N/A — SDK-internal. Developers on a dev cluster will see owned images move to the cluster registry on their next `kz-ext dev`; the first run after this lands rebuilds (prior dev-state artifacts were recorded under the old GHCR refs, and the resume probe correctly invalidates them).

## Checklist
- [x] Self-reviewed the code
- [x] Tests pass locally
- [x] Verified on live system (see Verification above)
- [ ] Documentation updated (if applicable) — none needed: `README.md` / `developer-guide.md` already documented the (previously false) cluster-registry contract this restores

### DIFF
--- a/kamiwaza_extensions/commands/dev.py
+++ b/kamiwaza_extensions/commands/dev.py
@@ -173,6 +173,7 @@ def _prior_artifacts_in_registry(
     try:
         canonical_refs = compute_canonical_refs(
             (info.compose_data or {}).get("services") or {},
+            purpose="dev",
             registry=registry,
             extension_name=info.name,
             revision_tag=prior_revision,
@@ -644,6 +645,7 @@ def run_dev_remote(
         extension_name=info.name,
         revision_tag=rev_tag,
         registry=registry,
+        purpose="dev",
         image_basename=info.image_basename,
     )
     # The catalog overlay (step 12) is a template destination: the platform
@@ -654,14 +656,16 @@ def run_dev_remote(
     catalog_compose = transformed
     transformed = transformer.resolve_env_placeholders(transformed)
 
-    # Canonical image refs for every build-context service. Single
-    # source of truth shared with the transformed compose so the image
-    # we build and push matches the ref the K8s payload will pull. The
-    # transformer honors a service's declared image namespace; without
-    # this map ImageBuilder would synthesize the legacy {ext}-{svc}
-    # form and ship a pod referencing an image that was never pushed.
+    # Canonical image refs for every build-context service, profiled ones
+    # included. Single source of truth shared with the transformed compose
+    # so the image we build and push matches the ref the K8s payload will
+    # pull, computed under the same ``purpose="dev"`` policy the transform
+    # above used: an owned build image lands in *this cluster's* registry
+    # even when its compose declares a qualified publish namespace
+    # (ENG-8626).
     canonical_refs: Dict[str, str] = compute_canonical_refs(
         info.compose_data.get("services") or {},
+        purpose="dev",
         registry=registry,
         extension_name=info.name,
         revision_tag=rev_tag,
@@ -675,10 +679,10 @@ def run_dev_remote(
     # ImagePullBackOff, chat 500. Rewrite that ref (and the sandbox
     # ``SANDBOX_ALLOWED_IMAGE_PREFIXES`` allowlist, in lockstep) to the
     # dev-built agent ref — the dev analog of ENG-5260's publish-side fix.
-    # ``build_image_ref_map`` mirrors ImageBuilder's per-service ref choice,
-    # so the env ref equals exactly what was built and pushed (the profiled
-    # ``image-only`` agent resolves to the ``{registry}/{ext}-agent`` fallback
-    # path). Applied to both the K8s payload (bare refs, post-resolve) and
+    # ``build_image_ref_map`` reads the canonical map directly, so the env ref
+    # equals exactly what was built and pushed — the profiled ``image-only``
+    # agent included, since ``purpose="dev"`` puts it in the map alongside its
+    # siblings. Applied to both the K8s payload (bare refs, post-resolve) and
     # the catalog overlay compose (``${VAR:-default}`` form).
     #
     # Caveat: env_ref_map spans ALL build-context services, so under
@@ -692,10 +696,6 @@ def run_dev_remote(
     env_ref_map = build_image_ref_map(
         info.compose_data.get("services") or {},
         canonical_refs,
-        registry=registry,
-        extension_name=info.name,
-        revision_tag=rev_tag,
-        image_basename=info.image_basename,
     )
     transformed = rewrite_env_image_refs(transformed, env_ref_map)
     catalog_compose = rewrite_env_image_refs(catalog_compose, env_ref_map)
@@ -788,18 +788,33 @@ def run_dev_remote(
         # payload will reference.
         #
         # ENG-7110: profiled ``image-only`` services (the agent, referenced
-        # via AGENT_SERVER_IMAGE — see the env rewrite above) are excluded
-        # from canonical_refs and so are intentionally NOT in this push list.
-        # A full run builds+pushes them via ImageBuilder (which iterates all
-        # build-context services); under --no-build they rely on a prior
-        # run's registry push (resume adopts that revision tag, so the agent
-        # ref the env rewrite points at already exists). Re-pushing them here
-        # would tag from the local engine store, which a resumed run may no
-        # longer hold — regressing the path that works.
+        # via AGENT_SERVER_IMAGE — see the env rewrite above) must NOT be in
+        # this push list. A full run builds+pushes them via ImageBuilder
+        # (which iterates all build-context services); under --no-build they
+        # rely on a prior run's registry push (resume adopts that revision
+        # tag, so the agent ref the env rewrite points at already exists).
+        # Re-pushing them here would tag from the local engine store, which a
+        # resumed run may no longer hold — regressing the path that works.
+        #
+        # ENG-8626: this exclusion is now explicit. It used to fall out of
+        # canonical_refs simply omitting profiled services, but the dev
+        # canonical map has to include them (that omission is what sent the
+        # agent to a different repository path than its siblings). Pushing is
+        # a separate question from identity, so the filter lives here.
+        profiled_names = {
+            name
+            for name, svc in (info.compose_data.get("services") or {}).items()
+            if isinstance(svc, dict) and svc.get("profiles")
+        }
+        pushable = {
+            name: ref
+            for name, ref in canonical_refs.items()
+            if name not in profiled_names
+        }
         if service:
-            image_refs = [canonical_refs[service]] if service in canonical_refs else []
+            image_refs = [pushable[service]] if service in pushable else []
         else:
-            image_refs = list(canonical_refs.values())
+            image_refs = list(pushable.values())
         console.print()
 
     # 7. Push images
--- a/kamiwaza_extensions/commands/publish.py
+++ b/kamiwaza_extensions/commands/publish.py
@@ -117,6 +117,7 @@ def _retag_appgarden_compose(
     # published_refs and the dev pipeline.
     canonical_by_name = compute_canonical_refs(
         source_services,
+        purpose="publish",
         registry=registry,
         extension_name=extension_name,
         revision_tag=image_tag,
@@ -146,7 +147,8 @@ def _retag_appgarden_compose(
             # Preserve declared namespace; we only rewrite the tag.
             svc["image"] = _canonical_build_ref(
                 svc, svc_name,
-                fallback_registry=registry,
+                purpose="publish",
+                registry=registry,
                 fallback_extension_name=extension_name,
                 revision_tag=image_tag,
                 fallback_image_basename=image_basename,
@@ -477,6 +479,7 @@ def _publish_one(
             extension_name=info.name,
             revision_tag=image_tag,
             registry=registry,
+            purpose="publish",
             image_basename=info.image_basename,
         )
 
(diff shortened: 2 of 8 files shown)


jxstanford (approved):
# Review Report: fix(extensions): kz-ext dev builds owned images into the cluster registry (ENG-8626)

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/230
**Branch**: `feature/eng-8626-kz-ext-dev-sends-qualified-owned-build-images-to-ghcr` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

CI Status: All checks passed / no conflicts. (Locally: 1681 extension tests pass, 1 unrelated skip.)

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 9/10 | Root-cause fix at the right altitude: splits publish/dev identity policy on an explicit required `purpose` rather than patching call sites. One under-covered resume interaction (see H1). |
| Security | 9/10 | No new attack surface; sandbox `SANDBOX_ALLOWED_IMAGE_PREFIXES` allowlist kept in lockstep with the relocated agent ref; net improvement — owned dev images no longer pushed to org GHCR. |
| Test Quality | 9/10 | Comprehensive unit + end-to-end coverage of dev relocation, allowlist lockstep, resume probe, and publish backward-compat; only the pathful-registry idempotency and pre-fix-state resume edges are untested. |
| Error Handling | 8/10 | Resume probe fails safe to rebuild; build failures raise cleanly; unmapped services skipped rather than invented. The same-revision pre-fix-state path (M... H1) bypasses the probe. |
| Code Quality | 10/10 | Clear naming, `Literal["dev","publish"]` type, four dead params dropped from `build_image_ref_map`, exceptional ticket-linked comments. |
| Refactoring | 9/10 | Ref logic consolidated into one source of truth; one missed comment/guard for the host-only-registry relocation assumption. |

## Critical Issues (Must Fix)
| — | — | None | — | — |

## High Priority Issues
| 1 | `kamiwaza_extensions/commands/dev.py:648` | Same-revision resume can skip build/push against stale pre-ENG-8626 state and deploy unpushed refs. When a workspace holds dev-state from a pre-fix run and the user reruns with the **same custom `--revision`**, `_resume_revision` returns `None` (because `last_revision == rev_tag`), so the new dev-registry probe (`_prior_artifacts_in_registry`) never runs. `_is_resumable` returns True for identical custom revisions, so build+push are skipped — but the old state only pushed the tag to `ghcr.io/...`, while the transform now deploys `{registry}/...` refs that were never pushed to the dev registry → ImagePullBackOff. Narrow (requires persisted pre-fix state + identical custom revision), but silent. Consider invalidating old state or verifying current canonical refs before any same-revision skip. | Codex (single-source) | Correctness / Error Handling |

| 1 | `kamiwaza_extensions/compose_transformer.py:246` | The dev relocation `f"{registry}/{repository}:{revision_tag}"` is idempotent only when the resolved dev `registry` has no path component. A pathful dev registry (e.g. `myreg.io/devspace`) would duplicate the path on re-relocation. Currently unreachable because `_canonical_build_ref` is always fed the original compose-declared ref (incl. on resume, which recomputes from `info.compose_data`). Worth a one-line comment documenting the host-only-registry assumption, or a guard/test for the pathful case. `test_relocation_is_idempotent` uses a path-less registry and gives false confidence about the general property. | Claude Code (single-source) | Correctness / robustness |
| 2 | `kamiwaza-ai-extensions-lib/package-lock.json` | Unrelated lockfile churn (removal of `libc` fields across many entries) is bundled into an otherwise tightly-scoped Python fix. Neither engine tied it to ENG-8626; confirm it's intentional or drop it to keep the diff focused. | Synthesis | Scope |

## Test Coverage Assessment
- **Verdict: Excellent.** New dev paths are covered end-to-end: `TestKaizenDevRegistryEndToEnd` asserts builder == pusher == deployed-compose refs, no-GHCR-push, external image untouched, and `AGENT_SERVER_IMAGE`/allowlist match the pushed agent ref; `TestResumeProbesDevRegistry` covers the resume probe hitting the dev registry (and correct rebuild when old-policy artifacts are absent); `TestCanonicalBuildRefDevPurpose`/`TestComputeCanonicalRefsDevPurpose` pin the relocation, path preservation, idempotency, and unqualified fallback; publish tests updated with `purpose="publish"` and verified unchanged.
- **Gaps:** (a) the pre-fix-state + identical-custom-revision resume path in H1; (b) pathful dev-registry idempotency (M1); (c) optional `--service <profiled-agent> --no-build` case.

## Positive Observations
- Root cause correctly targeted — one shared policy conflated "published identity" with "pull destination"; splitting on an explicit `purpose` at the identity layer is the right fix.
- `purpose` is required with no default, so an omitted call site fails loudly instead of silently inheriting publish semantics — exactly the regression being fixed.
- Push-vs-identity separation made explicit in `dev.py`; ENG-7110 `--no-build` profiled-agent exclusion preserved and re-tested.
- Security invariant maintained: sandbox controller admits exactly the image the backend launches; posture improved by no longer leaking owned dev images to org GHCR.
- Publish path reduces to byte-identical logic under `purpose="publish"`, confirmed by the full suite; comments tie every non-obvious decision to its ticket and failure mode.

## Overall Verdict
**Risk Level**: Low

Rationale: A well-scoped, root-cause bug fix with comprehensive tests and a valid Linear reference (ENG-8626). No Critical issues; all applicable dimensions ≥ 7. The single High finding (Codex) is a narrow, silent resume edge worth addressing in a follow-up but does not block — it requires persisted pre-fix dev-state combined with an identical custom `--revision`, and fails toward a redeploy rather than data loss. The Medium items are a documentation/guard nicety and unrelated lockfile churn.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>


**Reviewed-SHA**: 0a7bb34335659725dc61c4064e66bc2e461a9319

PR #237, opened 2026-07-20
ENG-8871: add NoAuth connector auth model (kind "none")

shkevin:
## Summary
- Adds `NoAuth` (kind `none`) to the connector auth-model union and exports it from `kamiwaza_sdk.connectors`: the model for public-content connectors that need no credential at all — nothing to configure, nothing to connect.
- Companion PRs: kamiwaza-internal/kamiwaza#2381 (core mint/proxy/visibility handling) and the kamiwaza-connector-website switch off its placeholder service_token.

## Testing
- `tests/unit/test_connectors_spi.py` extended (`auth_model_from_kind("none")` round-trip); 33 tests pass.

ENG-8871


kbot-ships (approved):
_Reviewing head `f6b44c7` against `develop`._

## Bottom line

This adds a third variant, `NoAuth` (`kind: "none"`), to the `AuthModel` discriminated union in `kamiwaza_sdk/connectors/provider.py`, registers it in `_AUTH_MODELS` so `auth_model_from_kind("none")` resolves, and re-exports it from `kamiwaza_sdk.connectors`. The union is marker-only — no fields, no behavior — and the reconstruction path is a plain dict lookup, so the change is additive with no branch logic to get wrong. I verified there is no exhaustive `match`/`if kind ==` dispatch anywhere in this repo that a new variant could fall through: the only consumers are `ConnectorSpec.to_manifest` (`provider.py:373`, serializes `self.auth_model.kind` generically) and `ConnectorSpec.from_manifest` (`provider.py:389`). `ruff check` is clean and `tests/unit/test_connectors_spi.py` + `test_connectors_service.py` pass (50 passed) locally. The real semantics — proxy not injecting an Authorization header, mint being rejected/skipped, visibility of a no-credential connector — live in the companion core PR (kamiwaza#2381), and none of that is enforceable from the SDK side. Merge risk here is low; the ordering risk is with core, not this PR. Verdict: APPROVE.

## Findings

### 1. `ConnectorSpec` manifest round-trip is not exercised with `NoAuth` — **(worth fixing)**

`tests/unit/test_connectors_spi.py:329-334` covers the `auth_model_from_kind("none")` lookup directly, but the full-envelope round-trip test (`test_connector_spec_manifest_roundtrip`, line ~294) still hardcodes `auth_model=ServiceToken()`. The narrow test would catch a missing `_AUTH_MODELS` entry, but not a regression in the envelope path — e.g. if someone later made `to_manifest` emit auth details conditionally on the variant, or if `from_manifest` grew a `data["auth_model"].get("kind", "service_token")` default. Smallest fix: parametrize the existing round-trip over the three variants, or add one case asserting `ConnectorSpec.from_manifest(spec.to_manifest()).auth_model == NoAuth()` (the dataclass is frozen so `==` works).

Confidence: high on the gap, moderate on how much it matters — the envelope is simple enough today that the failure mode is hypothetical.

### 2. Import ordering in the test file — **(nit)**

`tests/unit/test_connectors_spi.py:31` places `NoAuth` after `OpResult`, breaking the otherwise-alphabetical import block (`__init__.py` got it right, before `OAuthDescriptor`). `ruff check` passes, so this is cosmetic only, but it will look like drift to the next reader.

## Test gaps

Beyond finding 1, one scenario worth a line: `NoAuth()` is documented as "nothing to configure, nothing to connect," but nothing in this repo asserts that a `NoAuth` spec is valid with `oauth=None` and an empty `egress_allowlist`, or that `subscribe()` accepts `config=None`/`scopes=None` for it. That's arguably core's contract to test, not the SDK's — I'd only add it here if you want the SDK docstring at `services/connectors.py:130-136` ("Omit for connectors that need no credential") backed by an executable example. Not blocking.

I have nothing on concurrency, schema/migration, or performance — the change adds no state, no I/O, and no query. Security delta is neutral from the SDK's perspective: `NoAuth` is a declaration, and every enforcement decision (whether to inject an Authorization header, whether to allow minting for a `none` connector) is made in core. Worth confirming that core rejects a mint request from a `kind: "none"` connector rather than silently falling through to the service-token path — that's the one place this new kind could turn into an auth-boundary issue, and it's out of scope for this repo.

## What's good

The docstring on `NoAuth` (`provider.py:59-67`) explains the *why* and names the downstream behavior, matching the tone of the two existing variants rather than restating the class name. Registering via `NoAuth.kind` instead of a duplicated `"none"` literal keeps the discriminant single-sourced. And `auth_model_from_kind` still fails fast on unknown kinds — adding a variant did not soften that, which was the main thing to check.

## Verdict

**Verdict**: APPROVE

Additive, well-documented union variant with no exhaustive-dispatch consumers to break; the only gaps are a round-trip test case and an import-ordering nit, neither of which is worth blocking on.
---

PR #243, opened 2026-08-03
feat(extensions): scope Compose volumes per service and forward persistence [ENG-9604]

shkevin:
## Summary

Compose named volumes were translated into a single extension-level list of emptyDir volumes, with mounts distributed to services. Every service owns its own pod template, so that list belongs on the service.

The consequence was worse than a modelling wart: a service declaring persistence got an emptyDir mounted over its PVC path. The PVC was provisioned and bound, and nothing was ever written to it. Postgres in a Compose-authored extension came up green on an empty ephemeral volume and lost its data on every pod restart.

## What changed

- New `kamiwaza_extensions/compose_volumes.py` builds volumes and mounts per service. A Compose volume whose target matches the service's `x-kamiwaza.persistence` mountPath is skipped, leaving the operator's PVC mount in place. The operator-injected `tmp` and `data` names stay reserved so a user volume normalizing to either gets a collision suffix instead of producing a duplicate-name pod spec the API would reject.
- `persistence` is forwarded from `x-kamiwaza` to the CR alongside `volumes` and `volumeMounts`, so a Compose-authored extension can request a PVC. The operator has supported `PersistenceSpec` since its initial commit; the SDK simply never emitted it.
- `ExtensionServiceSpec` and `PatchServiceSpec` gain `persistence` / `volumes` / `volume_mounts` and serialize by alias, so `volumeMounts` survives the round-trip.

Two deploy-path fixes rode along, both in `commands/dev.py`:

- `--service` now filters the PATCH to the built service. The transformer assigns the new revision to every build-context service, but only the selected image is built and pushed, so PATCHing the siblings rolled them to nonexistent tags and produced `ImagePullBackOff`.
- The image PATCH carries a digest when the ref has one.

## Testing

`145 passed` across `test_payload_builder.py`, `test_dev_patch.py`, `test_dev_state.py`. `ruff` and `mypy` clean on the changed modules.

Verified end to end on a dev cluster: postgres and seaweedfs both reconcile with bound PVCs (40Gi / 30Gi, `ceph-block`) and the live pod spec shows `data` backed by the PVC rather than an emptyDir.

## Note

`payload_builder.py` shows -170/+72 because the volume helpers moved out to the new module. The remaining delta there is extraction of `_find_primary_service` and `_append_platform_env` from a function that had grown past readable size.

### DIFF
--- a/kamiwaza-ai-extensions-lib/package-lock.json
+++ b/kamiwaza-ai-extensions-lib/package-lock.json
@@ -746,6 +746,9 @@
                 "arm"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -763,6 +766,9 @@
                 "arm64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -780,6 +786,9 @@
                 "ppc64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -797,6 +806,9 @@
                 "riscv64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -814,6 +826,9 @@
                 "s390x"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -831,6 +846,9 @@
                 "x64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -848,6 +866,9 @@
                 "arm64"
             ],
             "dev": true,
+            "libc": [
+                "musl"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -865,6 +886,9 @@
                 "x64"
             ],
             "dev": true,
+            "libc": [
+                "musl"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -882,6 +906,9 @@
                 "arm"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -905,6 +932,9 @@
                 "arm64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -928,6 +958,9 @@
                 "ppc64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -951,6 +984,9 @@
                 "riscv64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -974,6 +1010,9 @@
                 "s390x"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -997,6 +1036,9 @@
                 "x64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1020,6 +1062,9 @@
                 "arm64"
             ],
             "dev": true,
+            "libc": [
+                "musl"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1043,6 +1088,9 @@
                 "x64"
             ],
             "dev": true,
+            "libc": [
+                "musl"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1239,6 +1287,9 @@
                 "arm64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "MIT",
             "optional": true,
             "os": [
@@ -1256,6 +1307,9 @@
                 "arm64"
             ],
             "dev": true,
+            "libc": [
+                "musl"
+            ],
             "license": "MIT",
             "optional": true,
             "os": [
@@ -1273,6 +1327,9 @@
                 "x64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "MIT",
             "optional": true,
             "os": [
@@ -1290,6 +1347,9 @@
                 "x64"
             ],
             "dev": true,
+            "libc": [
+                "musl"
+            ],
             "license": "MIT",
             "optional": true,
             "os": [
--- a/kamiwaza-ai-extensions-lib/package-lock.json
+++ b/kamiwaza-ai-extensions-lib/package-lock.json
@@ -746,9 +746,6 @@
                 "arm"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -766,9 +763,6 @@
                 "arm64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -786,9 +780,6 @@
                 "ppc64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -806,9 +797,6 @@
                 "riscv64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -826,9 +814,6 @@
                 "s390x"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -846,9 +831,6 @@
                 "x64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -866,9 +848,6 @@
                 "arm64"
             ],
             "dev": true,
-            "libc": [
-                "musl"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -886,9 +865,6 @@
                 "x64"
             ],
             "dev": true,
-            "libc": [
-                "musl"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -906,9 +882,6 @@
                 "arm"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -932,9 +905,6 @@
                 "arm64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -958,9 +928,6 @@
                 "ppc64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -984,9 +951,6 @@
                 "riscv64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1010,9 +974,6 @@
                 "s390x"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1036,9 +997,6 @@
                 "x64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1062,9 +1020,6 @@
                 "arm64"
             ],
             "dev": true,
-            "libc": [
-                "musl"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1088,9 +1043,6 @@
                 "x64"
             ],
             "dev": true,
-            "libc": [
-                "musl"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1287,9 +1239,6 @@
                 "arm64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "MIT",
             "optional": true,
             "os": [
@@ -1307,9 +1256,6 @@
                 "arm64"
             ],
             "dev": true,
-            "libc": [
-                "musl"
-            ],
             "license": "MIT",
             "optional": true,
             "os": [
@@ -1327,9 +1273,6 @@
                 "x64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "MIT",
             "optional": true,
             "os": [
@@ -1347,9 +1290,6 @@
                 "x64"
             ],
             "dev": true,
-            "libc": [
-                "musl"
-            ],
             "license": "MIT",
             "optional": true,
             "os": [
--- a/kamiwaza_extensions/commands/dev.py
+++ b/kamiwaza_extensions/commands/dev.py
@@ -96,24 +96,114 @@ def _build_patch_service_specs(
 def _validate_service_filter(
     service: Optional[str], compose_data: Dict[str, Any]
 ) -> None:
-    """Reject a ``--service`` name that no compose service answers to.
+    """Reject a ``--service`` name this run could not actually deploy.
 
-    Without this the filter yields an empty service list, and the failure
-    surfaces at the very end of the run as a raw pydantic ``min_length``
-    error that never mentions ``--service`` — after the whole build phase
-    has already run.
+    Raw Compose keys are the wrong set to check against. A profile-gated
+    service is stripped by the transformer, so the filter yields an empty
+    service list and ``PatchExtension`` raises a bare pydantic ``min_length``
+    error — after the whole build and push phase, naming neither the flag nor
+    the service. A service with no build context is not one ``--service`` can
+    build either: it would warn "No images to build" and PATCH the image the
+    author already declared, a silent no-op deploy.
     """
     if service is None:
         return
-    known = sorted((compose_data.get("services") or {}).keys())
-    if service in known:
-        return
-    console.print(
-        f"[red]Error:[/red] --service '{service}' is not a service in "
-        f"docker-compose.yml."
+    services = compose_data.get("services") or {}
+    if service not in services:
+        _fail_service_filter(
+            service, "is not a service in docker-compose.yml", services
+        )
+    config = services.get(service) or {}
+    if config.get("profiles"):
+        _fail_service_filter(
+            service,
+            "is profile-gated, so it is not part of a cluster deploy",
+            services,
+        )
+    if "build" not in config:
+        _fail_service_filter(
+            service,
+            "has no build context, so there is nothing to build or push for it",
+            services,
+        )
+
+
+def warn_orphaned_persistence(
+    prior_mounts: Dict[str, str], payload: Any
+) -> List[str]:
+    """Warn where a removed persistence block leaves the CR's PVC in place.
+
+    ``get_extension`` returns only ``ExtensionServiceStatus``, so the CR's own
+    spec cannot be read back to clear a block the extension has dropped. The
+    PVC therefore stays mounted while this PATCH sends a Compose emptyDir at
+    the same path: an identical target is a duplicate ``volumeMounts`` entry
+    the K8s API rejects, and a nested one shadows a claim still holding the
+    data the author believes they stopped writing.
+
+    Returns the messages emitted, so callers can assert on them.
+    """
+    warnings: List[str] = []
+    for service in payload.services:
+        prior = prior_mounts.get(service.name)
+        if not prior or service.persistence:
+            continue
+        collisions = [
+            mount["mountPath"]
+            for mount in (service.volume_mounts or [])
+            if _under_mount_path(mount.get("mountPath", ""), prior)
+        ]
+        if not collisions:
+            continue
+        message = (
+            f"Service '{service.name}' no longer declares persistence, but the "
+            f"deployed extension has a volume at '{prior}' that this deploy "
+            f"cannot clear. Compose still mounts {', '.join(collisions)} there, "
+            "which will either be rejected as a duplicate mount or silently "
+            "shadow the existing volume. Remove the Compose volume too, or "
+            "restore the persistence block."
+        )
+        warnings.append(message)
+        console.print(f"  [yellow]Warning:[/yellow] {message}")
+    return warnings
+
+
+def _deployed_persistence_mounts(payload: Any) -> Dict[str, str]:
+    """Per-service persistence mountPath this run is deploying."""
+    mounts: Dict[str, str] = {}
+    for service in payload.services:
+        persistence = service.persistence
+        if isinstance(persistence, dict) and persistence.get("enabled") is True:
+            mount_path = persistence.get("mountPath")
+            if isinstance(mount_path, str) and mount_path:
+                mounts[service.name] = mount_path
+    return mounts
+
+
+def _under_mount_path(target: str, mount_path: str) -> bool:
+    """Is *target* the persistence mountPath, or nested inside it?"""
+    if not target or not mount_path:
+        return False
+    normalized = target.rstrip("/") or "/"
+    base = mount_path.rstrip("/") or "/"
+    return normalized == base or normalized.startswith(base + "/")
+
+
+def _deployable_service_names(services: Dict[str, Any]) -> List[str]:
+    """Services ``--service`` can name: buildable and not profile-gated."""
+    return sorted(
+        name
+        for name, config in services.items()
+        if isinstance(config, dict) and "build" in config and not config.get("profiles")
     )
-    if known:
-        console.print(f"  [dim]Available: {', '.join(known)}[/dim]")
+
+
+def _fail_service_filter(
+    service: str, reason: str, services: Dict[str, Any]
+) -> None:
+    console.print(f"[red]Error:[/red] --service '{service}' {reason}.")
+    deployable = _deployable_service_names(services)
+    if deployable:
+        console.print(f"  [dim]Available: {', '.join(deployable)}[/dim]")
     raise typer.Exit(code=1)
 
 
@@ -964,6 +1054,10 @@ def run_dev_remote(
         console.print(f"[red]Error:[/red] invalid extension definition: {exc}")
         raise typer.Exit(code=1) from exc
 
+    warn_orphaned_persistence(
+        getattr(prior_state, "last_persistence_mounts", None) or {}, payload
+    )
+
     def _record(step: str) -> None:
         try:
             mark_step(
@@ -983,6 +1077,7 @@ def run_dev_remote(
                 registry=registry,
                 push_registry=push_registry,
                 image_basename=info.image_basename,
+                persistence_mounts=_deployed_persistence_mounts(payload),
                 build_engine=build_engine if step == "build" else "",
             )
         except OSError as state_exc:
--- a/kamiwaza_extensions/commands/dev.py
+++ b/kamiwaza_extensions/commands/dev.py
@@ -8,6 +8,7 @@ from typing import Any, Dict, List, Optional
 from urllib.parse import urlparse
 
 import typer
+from pydantic import ValidationError as PydanticValidationError
 from rich.console import Console
 
 console = Console(stderr=True)
@@ -66,7 +67,6 @@ def _build_patch_kwargs(
 def _build_patch_service_specs(
     payload: Any,
     service_filter: Optional[str] = None,
-    clear_persistence_for: Optional[frozenset] = None,
 ) -> List[Any]:
     """Build the per-service ``PatchServiceSpec`` list from a
     ``CreateExtension`` payload, forwarding the new ``x-kamiwaza``
@@ -86,39 +86,11 @@ def _build_patch_service_specs(
     leave the CR's image field at the original repository and produce
     ``ImagePullBackOff`` on the next pull.
     """
-    clear = clear_persistence_for or frozenset()
     return [
-        _build_patch_service_spec(service, clear_persistence=service.name in clear)
+        _build_patch_service_spec(service)
         for service in payload.services
         if service_filter is None or service.name == service_filter
     ]
-
-
-def persistent_service_names(extension: Any) -> frozenset:
-    """Names of services whose *current* CR has persistence enabled.
-
-    Used to decide where a removed compose persistence block must be
-    explicitly cleared. Sending a blanket disable to every service instead
-    would aim a PVC-affecting write at services that never asked for one.
-    """
-    services = getattr(getattr(extension, "spec", None), "services", None)
-    if services is None:
-        services = getattr(extension, "services", None) or []
-    return frozenset(
-        name
-        for service in services
-        if _persistence_is_enabled(service)
-        for name in [getattr(service, "name", None)]
-        if name
-    )
-
-
-def _persistence_is_enabled(service: Any) -> bool:
-    """True when a CR service carries persistence, dict- or model-shaped."""
-    persistence = getattr(service, "persistence", None)
-    if isinstance(persistence, dict):
-        return persistence.get("enabled") is True
-    return getattr(persistence, "enabled", False) is True
 
 
 def _validate_service_filter(
@@ -145,14 +117,7 @@ def _validate_service_filter(
     raise typer.Exit(code=1)
 
 
-def _patch_persistence(service: Any, clear: bool) -> Optional[dict]:
-    """Persistence value for one PATCH: declared, cleared, or untouched."""
-    if service.persistence:
-        return service.persistence
-    return {"enabled": False} if clear else None
-
-
-def _build_patch_service_spec(service: Any, *, clear_persistence: bool = False) -> Any:
+def _build_patch_service_spec(service: Any) -> Any:
     """Translate one create service into its complete PATCH contract."""
     from kamiwaza_extensions.compose_transformer import _split_image_ref
     from kamiwaza_sdk.schemas.extensions import ImagePatch, PatchServiceSpec
@@ -169,14 +134,12 @@ def _build_patch_service_spec(service: Any, *, clear_persistence: bool = False) 
         ),
         env=service.env or None,
         replicas=service.replicas,
-        # Removing an ``x-kamiwaza.persistence`` block must clear the PVC
-        # config on the CR: ``patch_extension`` dumps with
-        # ``exclude_none=True``, so None omits the field and leaves the old
-        # mount able to collide with a Compose volume reintroduced at the
-        # same path. The explicit disable goes *only* to services whose
-        # current CR actually has persistence on — a blanket disable would
-        # aim a PVC-affecting write at services that never requested one.
-        persistence=_patch_persistence(service, clear_persistence),
+        # Sent only when the extension declares it. Clearing a block the
+        # extension removed would need the CR's current spec, and
+        # ``get_extension`` returns a status projection
+        # (``ExtensionServiceStatus``) carrying no persistence — so a removed
+        # block stays on the CR rather than being silently mis-cleared.
+        persistence=service.persistence,
         # Volumes are service-scoped: every service owns a pod template.
         # Empty lists deliberately clear stale Compose volumes on PATCH.
         volumes=service.volumes or [],
@@ -985,14 +948,21 @@ def run_dev_remote(
         info.name, user_id=extract_user_id(token.access_token)
     )
     deployer_email = _decode_email(token.access_token)
-    payload = payload_builder.build(
-        metadata=info.metadata,
-        transformed_compose=transformed,
-        connection=connection,
-        dev_name=dev_name,
-        deployer=deployer_email,
-        revision=rev_tag,
-    )
+    try:
+        payload = payload_builder.build(
+            metadata=info.metadata,
+            transformed_compose=transformed,
+            connection=connection,
+            dev_name=dev_name,
+            deployer=deployer_email,
+            revision=rev_tag,
+        )
+    except (ValueError, PydanticValidationError) as exc:
+        # A malformed persistence or volume declaration would otherwise raise
+        # a raw traceback here — after the whole build and push phase has
+        # already run, with nothing naming the field at fault.
+        console.print(f"[red]Error:[/red] invalid extension definition: {exc}")
+        raise typer.Exit(code=1) from exc
 
     def _record(step: str) -> None:
         try:
@@ -1043,18 +1013,14 @@ def run_dev_remote(
         from kamiwaza_sdk.schemas.extensions import PatchExtension
 
         try:
-            # Check if extension already exists. The response is also the
-            # only view of which services currently hold a PVC, which is
-            # what decides where a removed persistence block gets cleared.
-            existing = client.extensions.get_extension(dev_name)
+            # Check if extension already exists
+            client.extensions.get_extension(dev_name)
 
             # Build patch from payload — extract image, env, replicas
             # plus the x-kamiwaza per-service overrides forwarded via
             # ``extra="allow"`` (jxstanford PR #97 review H2).
             patch_services = _build_patch_service_specs(
-                payload,
-                service_filter=service,
-                clear_persistence_for=persistent_service_names(existing),
+                payload, service_filter=service
             )
             # Carries the deployer/revision/deployed-at annotations on
             # every PATCH so `kz-ext status` reflects the current
--- a/kamiwaza_extensions/commands/dev.py
+++ b/kamiwaza_extensions/commands/dev.py
@@ -64,7 +64,9 @@ def _build_patch_kwargs(
 
 
 def _build_patch_service_specs(
-    payload: Any, service_filter: Optional[str] = None
+    payload: Any,
+    service_filter: Optional[str] = None,
+    clear_persistence_for: Optional[frozenset] = None,
 ) -> List[Any]:
     """Build the per-service ``PatchServiceSpec`` list from a
     ``CreateExtension`` payload, forwarding the new ``x-kamiwaza``
@@ -84,14 +86,73 @@ def _build_patch_service_specs(
     leave the CR's image field at the original repository and produce
     ``ImagePullBackOff`` on the next pull.
     """
+    clear = clear_persistence_for or frozenset()
     return [
-        _build_patch_service_spec(service)
+        _build_patch_service_spec(service, clear_persistence=service.name in clear)
         for service in payload.services
         if service_filter is None or service.name == service_filter
     ]
 
 
-def _build_patch_service_spec(service: Any) -> Any:
+def persistent_service_names(extension: Any) -> frozenset:
+    """Names of services whose *current* CR has persistence enabled.
+
+    Used to decide where a removed compose persistence block must be
+    explicitly cleared. Sending a blanket disable to every service instead
+    would aim a PVC-affecting write at services that never asked for one.
+    """
+    services = getattr(getattr(extension, "spec", None), "services", None)
+    if services is None:
+        services = getattr(extension, "services", None) or []
+    return frozenset(
+        name
+        for service in services
+        if _persistence_is_enabled(service)
+        for name in [getattr(service, "name", None)]
+        if name
+    )
+
+
+def _persistence_is_enabled(service: Any) -> bool:
+    """True when a CR service carries persistence, dict- or model-shaped."""
+    persistence = getattr(service, "persistence", None)
+    if isinstance(persistence, dict):
+        return persistence.get("enabled") is True
+    return getattr(persistence, "enabled", False) is True
+
+
+def _validate_service_filter(
+    service: Optional[str], compose_data: Dict[str, Any]
+) -> None:
+    """Reject a ``--service`` name that no compose service answers to.
+
+    Without this the filter yields an empty service list, and the failure
+    surfaces at the very end of the run as a raw pydantic ``min_length``
+    error that never mentions ``--service`` — after the whole build phase
+    has already run.
+    """
+    if service is None:
+        return
+    known = sorted((compose_data.get("services") or {}).keys())
+    if service in known:
+        return
+    console.print(
+        f"[red]Error:[/red] --service '{service}' is not a service in "
+        f"docker-compose.yml."
+    )
+    if known:
+        console.print(f"  [dim]Available: {', '.join(known)}[/dim]")
+    raise typer.Exit(code=1)
+
+
+def _patch_persistence(service: Any, clear: bool) -> Optional[dict]:
+    """Persistence value for one PATCH: declared, cleared, or untouched."""
+    if service.persistence:
+        return service.persistence
+    return {"enabled": False} if clear else None
+
+
+def _build_patch_service_spec(service: Any, *, clear_persistence: bool = False) -> Any:
     """Translate one create service into its complete PATCH contract."""
     from kamiwaza_extensions.compose_transformer import _split_image_ref
     from kamiwaza_sdk.schemas.extensions import ImagePatch, PatchServiceSpec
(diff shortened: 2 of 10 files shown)


jxstanford (changes requested):
# Review Report: feat(extensions): scope Compose volumes per service and forward persistence

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/243
**Branch**: `feat/compose-volume-persistence` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 7/10 | Per-service scoping is the correct model, but the persistence guard only handles exact-path collisions, removal of a persistence block never clears the CR, and `allowEmptyPersistenceInitialization` is undeclared scope. |
| Security | 8/10 | No injection, deserialization, or secret-exposure surface; host-path/`$`-interpolation rejection is preserved intact through the module move. Only concern is an unexplained flag now sent to the API on every PATCH. |
| Test Quality | 6/10 | Solid happy-path coverage and good regression tests for the deploy fixes, but the new `compose_volumes.py` module has no direct test file and none of the High findings are exercised. |
| Error Handling | 6/10 | Missing `mountPath` degrades to a silent `""` that disables the guard; an unmatched `--service` escapes as a raw pydantic traceback; persistence removal silently no-ops. |
| Code Quality | 8/10 | Clean small functions, consistent naming, ruff clean, comments explain rationale. `_first_value`'s truthiness test would drop a legitimately falsy source; operator volume-name knowledge is duplicated from the operator repo. |
| Refactoring | 8/10 | Well-executed extraction with behavior preserved; remaining opportunity is one shared persistence-path predicate for the skip and any future validator. |
| Documentation Accuracy | N/A | No Markdown files changed. |
| Documentation Standards | N/A | No Markdown files changed. |

## Critical Issues (Must Fix)
| 1 | PR metadata (branch / title / body / commits) | Missing Linear ticket reference — Kamiwaza repos require a related Linear ticket reference (e.g. ENG-1234) in the branch name, PR title, PR body, or commit messages. Note: `ENG-4834` appears only inside test docstrings carried over from prior work, which does not satisfy the policy gate. | Policy gate | Process |

## High Priority Issues
| 1 | `kamiwaza_extensions/compose_volumes.py:49` | The persistence guard is exact-match only (`target.rstrip("/") == persistence_mount`). A Compose volume targeting a path **nested inside** the persistence mountPath still emits an emptyDir mounted over the PVC subtree. Verified: `pgdata:/var/lib/postgresql/data` + `mountPath: /var/lib/postgresql` → `mounts=[{'name':'pgdata','mountPath':'/var/lib/postgresql/data'}]`. That is the stock Postgres layout (`PGDATA=/var/lib/postgresql/data`), so the exact data-loss scenario this PR exists to fix survives for the most likely authoring shape. The added test uses `target == mountPath`, the one case that works. Compare with `startswith(persistence_mount + "/")` as well as equality. | Claude Code | Business logic / data loss |
| 2 | `kamiwaza_extensions/compose_volumes.py:69` | `_persistence_mount_path` returns `""` when `persistence.enabled: true` but `mountPath` is absent, disabling the skip entirely rather than falling back to the operator default. Verified: `{'enabled': True, 'size': '5Gi'}` + `cache:/data` still emits an emptyDir at `/data`. Since the module reserves `data` as an operator-injected name, the operator's PVC default is evidently `/data` — yielding either a second mount shadowing the PVC or two `volumeMounts` at the same `mountPath`, which the K8s API rejects. Mirror the operator default or refuse to translate when `enabled` is true and `mountPath` is unset. | Claude Code | Business logic / edge case |
| 3 | `kamiwaza_sdk/schemas/extensions.py:189` | `allow_empty_persistence_initialization` is set by no code path, covered by no test, and unmentioned in the PR description. Being non-`Optional` with a `False` default while `patch_extension` serializes via `model_dump(exclude_none=True)`, every service in **every** dev PATCH now carries `"allowEmptyPersistenceInitialization": false`. Verified: `PatchServiceSpec(name='x').model_dump(exclude_none=True)` → `{'name': 'x', 'allowEmptyPersistenceInitialization': False}`. An unexplained blanket `false` against a persistence-initialization gate is not a safe silent default. Drop the field or make it `Optional[bool] = None` with a docstring and a test. | Claude Code | API contract / scope creep |
| 4 | `kamiwaza_extensions/commands/dev.py:111` | Removing an `x-kamiwaza.persistence` block from compose never clears it on the CR. The spec is built with `persistence=None`, and `patch_extension` dumps with `exclude_none=True`, so the field is omitted and the old PVC config persists. If a compose volume at the same mount path is then reintroduced as an emptyDir, the retained PVC mount can collide with the new `volumeMounts`. This directly contradicts the clear-on-removal contract the adjacent `volumes`/`volumeMounts` lines implement (`volumes=service.volumes or []`). Send an explicit disabled/cleared persistence value in the no-persistence case. | Codex (single-source) | Business logic / backward compatibility |

| 5 | `kamiwaza_extensions/commands/dev.py:982` | A `--service` value matching no payload service produces `services=[]`, and `PatchExtension` enforces `min_length=1` → uncaught `pydantic.ValidationError` and a raw traceback. `ImageBuilder` normally catches the bad name first, but `--no-build` skips that guard entirely (`dev.py:812`, which silently degrades to `image_refs = []`). Validate the filter against `payload.services` and exit with a message. | Claude Code | Error handling |
| 6 | `kamiwaza-ai-extensions-lib/package-lock.json:1275` | The diff deletes `libc` selectors (`glibc`/`musl`) from native optional packages, leaving only `os`/`cpu` constraints. `npm ci` on Linux can then no longer distinguish e.g. `@next/swc-linux-x64-gnu` from `@next/swc-linux-x64-musl` from the lockfile. This is unrelated churn from an older npm; regenerate with an npm that preserves `libc`, or revert the lockfile hunk. | Codex (single-source) | Build / dependency hygiene |
| 7 | `kamiwaza_extensions/commands/dev.py:45-63` | Dropping `volumes` from `_build_patch_kwargs` is right, but CRs created by prior SDK versions have a stale root-level `volumes` list persisted; nothing in this PR clears it. Worth either one explicit clearing PATCH or a note that the field is inert server-side. | Claude Code | Backward compatibility |
| 8 | `kamiwaza_extensions/commands/dev.py:99-107` | An untagged digest ref (`repo@sha256:...`) yields `tag="latest"` alongside the digest, so the operator reconstructs `repo:latest@sha256:...`. The digest wins at pull time so it resolves, but the fabricated tag will surface in `kz-ext status`. | Claude Code | Code quality |
| 9 | `tests/unit/extensions/` | No test module for `compose_volumes.py`; all coverage routes through `test_payload_builder.py`. See coverage section. | Claude Code | Test coverage |

## Test Coverage Assessment
- **Adequately covered**: `kamiwaza_sdk/schemas/extensions.py` (alias round-trip asserted in `test_dev_state.py`), `kamiwaza_extensions/commands/dev.py` (digest pin, persistence forwarding, service filter, clear-on-removal), `kamiwaza_extensions/payload_builder.py` (per-service volumes, shared volume, long-form read-only, host-path rejection, `tmp`/`data` collision, persistence enabled *and* disabled).
- **Gaps**: `kamiwaza_extensions/compose_volumes.py` is a new standalone parsing module with no direct test file. Untested branches: nested persistence target (H1), `enabled: true` with no `mountPath` (H2), long-form `type: bind` rejection, the `readOnly` camelCase key (only `read_only` is exercised), `dst`/`destination` aliases, `>63`-char name truncation, collision counter past `-2`, non-str/non-dict volume entries. Also untested: `allowEmptyPersistenceInitialization` (H3), persistence-removal clearing (H4), and the empty-service-filter path (M5).
- All 2942 unit tests pass and ruff is clean; no failures are attributable to this change.

## Positive Observations
- The core diagnosis is right and well argued: volumes belong on the pod template, and every service owns one. Moving them off the extension root is the correct model.
- Reserving `tmp`/`data` was carried through the refactor intact; `test_user_volume_named_tmp_avoids_operator_collision` still asserts collision-suffix behavior against the new per-service scope.
- `serialize_by_alias=True` is a genuine fix, not cosmetics — `patch_extension` dumps without `by_alias`, so `volume_mounts` would otherwise have shipped snake_case and been dropped by the API server. It correctly follows the precedent `ExtensionPort` set for `appProtocol`.
- The `--service` PATCH filter fixes a real `ImagePullBackOff`, and the docstring explains *why* rather than *what*.
- Decomposition of `_build_services` (`_find_primary_service`, `_append_platform_env`, `_add_service_overrides`) is a net readability win, and `_find_primary_service` quietly adds an `isinstance` guard the old loop lacked.
- The shared-named-volume semantic change (per-pod emptyDir, no sharing) is already surfaced to users by an existing validator warning in `validators/compose.py:218-225`, so the renamed test documents rather than hides it.

## Overall Verdict
**Risk Level**: High
**Verdict**: REQUEST_CHANGES

Rationale: the Linear ticket gate fails (hard blocker). Independently, findings H1/H2 leave the exact silent-data-loss failure this PR was written to eliminate reachable through the most common Postgres authoring shape (nested `PGDATA` target) and through `persistence.enabled` without an explicit `mountPath`; H4 leaves stale PVC config on the CR when a persistence block is removed, contradicting this PR's own clear-on-removal contract; H3 ships an undocumented, code-unreachable `allowEmptyPersistenceInitialization: false` on every service of every dev PATCH.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

**Reviewed-SHA**: e22137eb6802c5a321ec39b1562cbd76afecbb1b

jxstanford (changes requested):
# RE-REVIEW Report: feat(extensions): scope Compose volumes per service and forward persistence [ENG-9604]

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/243
**Branch**: `feat/compose-volume-persistence` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 8/10 | Scope matches the description; the primary-service and platform-env extractions are adjacent but defensible. Ruff-format churn in test files is drive-by noise (71 unformatted test files already in repo), not a new convention. |
| Security | 8/10 | No injection, secret, authz, or deserialization surface. One boundary gap: `x-kamiwaza.persistence` is forwarded verbatim into a K8s CR with only an emptiness check on `mountPath` (H1, M2). |
| Test Quality | 7/10 | The new `test_compose_volumes.py` is well-targeted and readable, but the two highest-blast-radius new behaviours (the `{"enabled": False}` PATCH default, the malformed-`mountPath` path) are untested. |
| Error Handling | 6/10 | Validation raised from the translation layer after build+push (M1), raw pydantic error for a typo'd `--service` (H3), misleading message for `mountPath: "/"` (M5). |
| Code Quality | 9/10 | Good decomposition, clear naming, no duplication. Minor: `_append_platform_env` compressed away the "empirical proof point" rationale for why explicit env beats configmap injection. |
| Refactoring | 8/10 | The extraction into `compose_volumes.py` is the right call and cleanly done; missed opportunity is centralizing persistence validation with the other compose validators. |
| Documentation Accuracy | N/A | No Markdown changed in this PR. |
| Documentation Standards | N/A | No Markdown changed in this PR. |

## Critical Issues (Must Fix)
| — | — | None | — | — |

## High Priority Issues
| 1 | `kamiwaza_extensions/compose_volumes.py:93` | **The re-review fix is incomplete — the original emptyDir-over-PVC data-loss bug is still reachable.** `_persistence_mount_path` now rejects an *empty* `mountPath` but accepts any other string, so two inputs slip through the guard and then never match a `/`-prefixed Compose target. (a) A non-absolute value: `{enabled: true, mountPath: "var/lib/postgresql"}` + `pg:/var/lib/postgresql/data` → `mounts=[{'name':'pg','mountPath':'/var/lib/postgresql/data'}]`, an emptyDir over the PVC (reproduced). Same for `mountPath: "data"` + `pg:/data`. (b) An explicit YAML null: `mountPath:` with no value parses to `None`, and `str(None)` → `"None"`, which is truthy, so the missing-path guard never fires and the volume is translated to an emptyDir with the PVC left unmounted (reproduced: `ServiceVolumeSpec(volumes=[{'name': 'data-2', ...}], mounts=[{'name': 'data-2', 'mountPath': '/data'}])`). Fix both beside the existing check: reject `None` before stringifying, and require `mount_path.startswith("/")`. | Claude Code, Codex | correctness / data-loss |
| 2 | `kamiwaza_extensions/commands/dev.py:116` | `persistence=service.persistence or {"enabled": False}` sends an explicit **disable on every service of every PATCH**, including services that never mentioned persistence. Verified body: `{'name':'backend', ..., 'persistence': {'enabled': False}, ...}`. This is the mirror of the bug the PR exists to fix — whether it merely unmounts or reclaims the PVC depends on operator behaviour that isn't stated anywhere in the diff. Also note persistence is read *only* from compose `x-kamiwaza`, unlike `healthCheck` which has a `kamiwaza.json` precedence path (`_metadata_service_field`), so a metadata-authored persistence block gets disabled on every redeploy. No test covers this default. | Claude Code (single-source) | correctness / data-loss |
| 3 | `kamiwaza_extensions/commands/dev.py:88` | `--service` is never validated against the compose services. With a typo the new filter yields `[]`, and `PatchExtension.services` carries `min_length=1` → `ValidationError: List should have at least 1 item after validation, not 0`, surfaced by `_handle_exception` as a raw pydantic string that never mentions `--service`. This fires *after* the whole build phase. Validate the flag against `info.compose_data["services"]` at entry. | Claude Code (single-source) | error-handling / UX |

| 1 | `kamiwaza_extensions/compose_volumes.py:95` | The new `ValueError` is raised from the translation layer at deploy step 8 — *after* images are built and pushed. `validators/compose.py` doesn't check persistence at all, so `kz-ext validate` passes and the author burns a full build cycle before the error. Move (or mirror) the check into the compose validator. | Claude Code | architecture |
| 2 | `kamiwaza_extensions/compose_volumes.py:91` | `persistence.get("enabled") is not True` — a non-bool truthy value (`"true"` from quoted YAML/JSON) silently disables the guard while the same dict is still forwarded verbatim to the CR. Same failure class as H1. | Claude Code | data-validation |
| 3 | `kamiwaza_extensions/validators/compose.py:139` | The one user-facing message about named volumes still reads "named volumes deploy as emptyDir, so data is lost on pod restart" with no pointer to the new `x-kamiwaza.persistence` escape hatch. This PR ships the capability with zero discoverability (`x-kamiwaza.persistence` appears in no doc in the repo — a repo-wide grep finds it only in the two schema files). | Claude Code | docs / UX |
| 4 | `kamiwaza_extensions/commands/dev.py:694-696` | Stale comment: "That's the same accepted partial-deploy sharp edge as the service `image:` fields (a `--service backend` run already deploys un-rebuilt siblings)". This PR is exactly what stops that from happening. | Claude Code | comment-rot |
| 5 | `kamiwaza_extensions/compose_volumes.py:93` | `mountPath: "/"` normalizes to `""` and produces "mountPath is missing", which is misleading for a path that was supplied. | Claude Code | error-message |
| 6 | `kamiwaza_extensions/commands/dev.py:113` | `env=service.env or None` doesn't follow the clear-on-removal contract the adjacent `volumes` / `volumeMounts` / `persistence` lines all now implement — emptying env in compose leaves stale env on the CR. Pre-existing, but the surrounding lines make the inconsistency conspicuous. | Claude Code | consistency |
| 7 | `kamiwaza_extensions/commands/dev.py:1005` | The 405 `_delete_and_recreate` fallback ignores `service_filter` and recreates from the full payload, so a `--service` run that hits it deploys siblings at tags that were never pushed — the exact failure the PATCH filter was added to prevent. | Claude Code | correctness |

## Test Coverage Assessment
- **Local verification**: `171 passed` across `test_payload_builder.py`, `test_dev_patch.py`, `test_dev_state.py`, `test_compose_volumes.py`. `ruff check` clean on all changed modules. (`mypy` not installed in the review environment — not verified.)
- **Adequately covered**: `compose_volumes.py` — the new direct test module is genuinely good (nested/sibling-prefix, both read-only spellings, long-form aliases, malformed entries, 63-char truncation, operator-name collisions). `payload_builder.py` — per-service volumes, shared source, enabled/disabled persistence. `schemas/extensions.py` — alias round-trip; nested `PatchExtension`/`CreateExtension` serialization was independently confirmed to honour `serialize_by_alias` (no test covers it directly), and the pydantic `>=2.11` floor is correct for that config key.
- **Gaps**:
  - `compose_volumes.py` — non-absolute `mountPath` and explicit-null `mountPath` (H1), non-bool `enabled` (M2), same source mounted at both a covered and an uncovered path in one service.
  - `commands/dev.py` — the `{"enabled": False}` PATCH default when no persistence is declared (H2); `--service` naming a non-existent service (H3); `_delete_and_recreate` under `--service` (M7).
- Test churn is substantive, not cosmetic — old assertions against `payload.model_extra["volumes"]` were rewritten to the new per-service contract rather than deleted.

## Positive Observations
- The core modelling fix is right and the reasoning in the PR body is accurate: volumes are pod-scoped, every service owns a pod template, so a single extension-level list was wrong.
- All six items from the previous round were addressed in good faith — nested-path guard (`_covered_by_persistence` + tests), missing-`mountPath` rejection, `allowEmptyPersistenceInitialization` dropped, persistence clear-on-removal, new direct test module, package-lock churn reverted in `c1f10ac`.
- `compose_volumes.py` is a clean extraction — small single-purpose functions, `frozen=True` dataclass, and the `_OPERATOR_VOLUME_NAMES` reservation correctly moved to per-service scope.
- Comments consistently explain *why* (the `exclude_none=True` interaction, the `tmp`/`data` collision rationale, the `--service`/`ImagePullBackOff` chain) rather than restating the code.
- `_find_primary_service` is not just tidier — the new `isinstance(frontend, dict)` guard fixes a latent crash the original loop had.
- Verified as non-issues: nested alias serialization works through both `PatchExtension` and `CreateExtension`; the three new `null` keys in the CREATE body match the pre-existing pattern (`env`/`resources`/`command`/`args` already serialize as `null`), so no API contract change.

## Overall Verdict
**Risk Level**: Medium
**Verdict**: REQUEST_CHANGES

Both engines independently landed on the same line — `compose_volumes.py:93` — from different angles (non-absolute path vs. explicit YAML null), which is strong evidence the guard is under-specified rather than merely incomplete for one input. The previous round's feedback was addressed thoroughly, but the headline fix still lets the original emptyDir-over-PVC data-loss bug through (H1, a two-line fix). Paired with the untested unconditional `persistence: {"enabled": false}` on every PATCH (H2) and Error Handling at 6/10, the two highest-risk paths in a PR specifically about preventing silent data loss still need work.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

**Reviewed-SHA**: c1f10ac4d2ec54925c50bf9a91fe012185455ae3

jxstanford (changes requested):
# RE-REVIEW Report: feat(extensions): scope Compose volumes per service and forward persistence [ENG-9604]

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/243
**Branch**: `feat/compose-volume-persistence` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

Local verification (Claude Code): `188 passed` across the four touched test modules; `ruff check` clean on all four changed source files. Prior-round findings are all addressed — the nested-mountPath guard, mountPath validation, targeted persistence clear, kamiwaza.json→compose precedence, and early `--service` validation all landed, and the unrelated `package-lock.json` churn was reverted. Findings below are in the new delta.

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 7/10 | Scope matches the stated goal, but the kamiwaza.json persistence precedence added last round was wired into the payload without being wired into the volume guard it must feed. |
| Security | 8/10 | No injection, secret, or authz surface; volume names sanitized to DNS-1123 labels. Gaps are boundary validation, not exploitable. |
| Test Quality | 6/10 | Excellent parametrized parser coverage, but the newest code path (`_resolve_persistence`) is entirely untested and one test's assertion doesn't match its name. |
| Error Handling | 5/10 | Two new `raise` sites with no handler at the CLI boundary, surfacing post-build as tracebacks — inconsistent with the early-validation pattern the same commit establishes for `--service`. |
| Code Quality | 9/10 | Clean decomposition, precise naming, comments that carry the reasoning rather than restating the code. |
| Refactoring | 8/10 | Good extraction; persistence resolution should be hoisted to a single source feeding both the guard and the spec, which would eliminate Critical #1 structurally. |
| Documentation Accuracy | N/A | No Markdown changed in this PR. |
| Documentation Standards | N/A | No Markdown changed in this PR. |

## Critical Issues (Must Fix)
| 1 | `kamiwaza_extensions/payload_builder.py:252` | `build_service_volume_specs(transformed)` reads persistence **only** from compose `x-kamiwaza`, but the new `_resolve_persistence` (line 291) lets `kamiwaza.json` declare it. A kamiwaza.json-declared PVC never arms the volume guard: the service ships `persistence.enabled=true` **and** an `emptyDir` mounted at the same path — the exact ENG-9604 data-loss bug this PR exists to fix, reintroduced through the newest commit. All new `mountPath` validation (missing / relative) is bypassed on this path too. Reproduced live: with persistence declared in `metadata` and `volumes: ["pg:/var/lib/postgresql/data"]`, the payload emits `persistence: {enabled: True, mountPath: /var/lib/postgresql}` alongside `volumes: [{name: pg, emptyDir: {}}]` and `volumeMounts: [{name: pg, mountPath: /var/lib/postgresql/data}]`. Fix: resolve persistence once (metadata-then-compose) and thread the resolved mountPath into `build_service_volume_specs` / `_build_service_volume_spec`. | Claude Code, Codex | correctness |

## High Priority Issues
| 2 | `kamiwaza_extensions/commands/dev.py:97-106` | `persistent_service_names` assumes `get_extension`'s per-service objects carry `persistence`. `Extension.services` is typed `List[ExtensionServiceStatus]` — a status projection (`name`/`ready`/`replicas`/`message`). Codex confirmed by construction: `Extension(..., spec={"services": [{"name": "pg", "persistence": {"enabled": True}}]})` returns `frozenset()`, because a dict-shaped extra `spec` makes `getattr(spec, "services")` fail and the fallback reads status-only data. Clear-on-removal then silently never fires — the regression `d97951b` set out to fix, now failing quietly. Only tested against a hand-built `SimpleNamespace`. | Claude Code, Codex | correctness |
| 3 | `kamiwaza_extensions/compose_volumes.py:100` | The new `ValueError`s have no handler. `payload_builder.build()` at `dev.py:988` is unwrapped, so a missing or relative `mountPath` prints a raw traceback **after** the whole build+push phase — verbatim the failure mode the same commit's `_validate_service_filter` docstring calls out and fixes for `--service`. Same spot: a non-dict `persistence` from kamiwaza.json (`persistence: true`) raises an uncaught pydantic `ValidationError`. | Claude Code (single-source) | error-handling |

| 4 | `kamiwaza_extensions/commands/dev.py:1056`, `:1076` | The `--service` filter guards only the PATCH branch. First deploy (POST) and the 405 `_delete_and_recreate` fallback still send every build-context service at the new revision while only one image was pushed — the same `ImagePullBackOff` remains reachable on those two paths. | Claude Code (single-source) | correctness |
| 5 | `kamiwaza_extensions/commands/dev.py:1056` | With `--service` active, a *sibling* that dropped its persistence block is filtered out of the PATCH and never cleared, so the stale PVC mount survives until a full redeploy. Probably acceptable, but undocumented. | Claude Code (single-source) | edge-case |
| 6 | `tests/unit/extensions/test_dev_patch.py:350` | `test_unknown_service_exits_with_a_message_naming_the_flag` takes `capsys` and never reads it — it only asserts `typer.Exit`. The message content the test name promises (and that `_validate_service_filter` exists to produce) is unasserted. | Claude Code (single-source) | test-coverage |
| 7 | `kamiwaza_extensions/compose_volumes.py:99` | `mountPath: "/"` normalizes to `""` and raises *"mountPath is missing"* — misleading for a value that was present. Distinguish "absent" from "not a usable mount root". | Claude Code (single-source) | error-message |
| 8 | `kamiwaza_extensions/commands/dev.py:161` (`_build_patch_service_spec`) | `service.image.partition("@")` on a ref ending in a bare `@` yields `digest=""`, which survives `exclude_none=True` and reaches the operator as an empty digest. Guard with `digest or None`. | Claude Code (single-source) | edge-case |

## Test Coverage Assessment
- **Adequately covered**: `compose_volumes.py` (parser branches, both read-only spellings, long-form aliases, malformed entries, truncation, operator-name collisions, nested/sibling-prefix persistence, absolute/relative/non-string mountPath — genuinely thorough); `schemas/extensions.py` alias round-trip; the `--service` filter and digest-pin paths in `test_dev_patch.py`.
- **Gaps**:
  - `payload_builder.py` — **no test anywhere for kamiwaza.json-declared persistence**, neither the precedence itself nor its interaction with the volume guard. A test mirroring `test_explicit_persistence_replaces_matching_empty_dir` but declaring persistence in `metadata` would have caught Critical #1 immediately.
  - `commands/dev.py` — `persistent_service_names` needs a case built from a real `Extension.model_validate({...})` of a representative GET body, not `SimpleNamespace` (High #2).
  - No test at the CLI boundary asserting the `ValueError` becomes a usable message rather than a traceback (High #3).
  - Medium #6: assert the `--service` error text.

## Positive Observations
- Every finding from the prior round is genuinely fixed, not papered over — the nested-path guard, `mountPath` validation, and the targeted (rather than blanket) PVC clear each land with a test pinning the specific failure.
- Targeting the persistence clear via the `get_extension` response the deploy path already fetched and discarded is the right instinct: no extra round-trip, and it avoids aiming a PVC-affecting write at services that never asked for one.
- The refactor is a real improvement: `_find_primary_service`, `_append_platform_env`, `_add_service_overrides`, and the parse-helper split in `compose_volumes.py` reduce nesting without changing behavior, and the docstrings consistently explain *why* (the sibling-prefix and `str(None)` notes are exactly what a future reader needs).
- Reserving `tmp`/`data` per-service and suffixing collisions correctly follows the operator's `[tmp] + [data PVC] + svc.Volumes` assembly.
- The `package-lock.json` churn was diagnosed precisely (npm dropping libc selectors, breaking `npm ci` musl/gnu resolution) and reverted rather than hand-waved.

## Overall Verdict
**Risk Level**: High
**Verdict**: REQUEST_CHANGES

Both engines independently landed on the same two defects. The kamiwaza.json persistence precedence added in the most recent commit bypasses the volume guard entirely, reproducibly re-emitting an emptyDir over the PVC mount — the exact silent data-loss failure this PR exists to eliminate. Codex additionally confirmed by construction that the persistence-clear path returns an empty set against a realistic CR response shape, so removing a persistence block silently leaves the stale PVC config in place. Those two plus the unhandled `ValueError` at the CLI boundary should land before merge; the rest is straightforward follow-up.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

**Reviewed-SHA**: 55501007b552ad6ab881a8a45e8d31846f9a4d0e

jxstanford (changes requested):
# RE-REVIEW Report: feat(extensions): scope Compose volumes per service and forward persistence [ENG-9604]

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/243
**Branch**: `feat/compose-volume-persistence` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

Claude Code verified locally: `2984 passed, 2 skipped` on `tests/unit` (both skips environment-related), `ruff check` clean on all four changed modules. No failures attributable to this change.

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 9/10 | Service-scoped volumes is the right model; the two deploy-path riders (`--service` PATCH filter, digest pin) are disclosed and genuinely adjacent. |
| Security | 8/10 | No injection/secret/access-control surface; `looks_like_host_path` still blocks host-path escape into `emptyDir`. C1/M1 are integrity (volume shadowing) rather than classic OWASP findings. |
| Test Quality | 7/10 | Strong direct coverage of the new module (217 lines), undercut by an assertion-free test (M2) and the untested payload-builder precedence seam (M3) where a regression already landed once. |
| Error Handling | 7/10 | New build-time `ValueError`/`ValidationError` wrapper and `--service` validation are real improvements; C1 leaves a validation hole in the same function that validates `mountPath`, and H1/H2 surface only as opaque late failures. |
| Code Quality | 9/10 | Small focused functions, ruff-clean, comments explain why not what. M7/M9 are minor. |
| Refactoring | 8/10 | Extraction of `compose_volumes.py` and the `_find_primary_service`/`_append_platform_env` simplifications are behaviour-preserving and net-positive; M7 leaves ref parsing split across two modules. |

## Critical Issues (Must Fix)
| C1 | `kamiwaza_extensions/compose_volumes.py:98` | `persistence.get("enabled") is not True` is a strict identity check, so `enabled: "true"`, `enabled: 1`, or `enabled: "yes"` disarm the guard entirely and `persistence_mount_path` returns `""`. Verified by execution: `{"enabled": "true", "mountPath": "/var/lib/postgresql"}` emits the emptyDir. The platform layer coerces truthy scalars (pydantic lax mode), so the PVC is still provisioned and, for the stock nested Postgres layout, an emptyDir shadows the PVC subtree — re-opening the exact silent data loss this PR exists to remove. `_build_sandbox_spec` in the same feature area already accepts `"true"/"1"/"yes"/"on"` for its own `persistence` key, so a user copying that idiom lands here. Fix: reject non-bool `enabled` with a `ValueError` alongside the existing two — symmetric with the `mountPath` validation accepted in round 2 for the identical reason. **Severity elevated from Claude's High**: the failure is confirmed by execution, silent, and destroys user data in the feature's headline scenario. | Claude Code (single-source) | Correctness / data loss |

## High Priority Issues
| H1 | `kamiwaza_extensions/commands/dev.py:108-110` | `_validate_service_filter` accepts any raw Compose service name. Two distinct failure modes: (a) Codex demonstrated by execution that a **profile-gated** service (e.g. Kaizen's `image-only` `agent`) passes validation but is stripped by `ComposeTransformer`, so `_build_patch_service_specs(..., service_filter=...)` returns `[]` and `PatchExtension` raises a raw pydantic `min_length` error — the precise late, unnamed failure this function was added to prevent, now after the whole build+push phase; (b) a service with no build context passes validation, warns "No images to build", and PATCHes to its authored image — a silent no-op deploy. Validate against deployable/build-context services instead of raw Compose keys. | Claude Code (M4), Codex (P2) — higher severity used | Correctness / validation |
| H2 | `kamiwaza_extensions/commands/dev.py:142` | Dropping clear-on-removal is the right call given `get_extension` returns only `ExtensionServiceStatus`, but the consequence is unguarded and unsurfaced. Remove `x-kamiwaza.persistence` while keeping the Compose volume and the CR keeps `persistence.enabled=true` while the PATCH now sends the emptyDir back: identical `mountPath` → duplicate `volumeMounts[].mountPath`, rejected by the K8s API, so the Deployment stops reconciling with an opaque error; nested target → emptyDir shadows the still-bound PVC and the user believes they went ephemeral while data goes nowhere. `dev_state.py:47` already persists per-run inputs (`last_service`), so recording last-deployed persistence per service would allow either an explicit disable or an actionable error. At minimum warn when a Compose volume target falls under a previously-deployed `mountPath`. | Claude Code (single-source) | Correctness / error handling |

| M1 | `kamiwaza_extensions/compose_volumes.py:83` | Targets are only `rstrip("/")`-normalized, so a non-canonical but path-equivalent target escapes the guard. Verified: `pg://var/lib/postgresql/data` under `mountPath: /var/lib/postgresql` emits the emptyDir (`/./data` and `/../` variants are caught by luck of string prefixing; `//` is not). `posixpath.normpath` on both sides closes it. | Claude Code | Correctness |
| M2 | `tests/unit/extensions/test_dev_patch.py:263` | `test_unknown_service_exits_with_a_message_naming_the_flag` requests `capsys` and never reads it — the only assertion is that `typer.Exit` raised. The behaviour the name promises (message names `--service`, lists available services) is unverified, so the `console.print` block at `dev.py:111-116` could be deleted with the suite green. | Claude Code | Test quality |
| M3 | `kamiwaza_extensions/payload_builder.py:255` | No test drives `_resolve_persistence`'s kamiwaza.json-over-compose precedence through `PayloadBuilder.build`. `test_persistence_from_metadata_arms_the_guard` passes the resolved dict straight into `build_service_volume_specs`, testing the module contract, not the wiring — and that wiring seam is exactly where d97951b reintroduced the data loss and 8abb59a fixed it. | Claude Code | Test coverage |
| M4 | `kamiwaza_extensions/commands/dev.py:952-963` | The new `except (ValueError, PydanticValidationError)` handler has no test. It is the only thing between a malformed `persistence` block and a raw traceback after the entire build-and-push phase, and both raise sites (`persistence_mount_path`, `_parse_resources`) are reachable from user input. | Claude Code | Test coverage |
| M5 | `kamiwaza_sdk/schemas/extensions.py:169` | `ImagePatch.digest` is a new wire field; if the platform's own `ImagePatch` forbids extras, digest-pinned refs would now 422 where they previously succeeded. Also, for a digest-only ref (`repo@sha256:…`) `_split_image_ref` fabricates `tag="latest"` and pairs it with the real digest — if the operator prefers tag over digest, the wrong image is pulled. Confirm against the operator's reconstruction before merge. | Claude Code | API contract |
| M6 | `kamiwaza_extensions/commands/dev.py:125-126` | `_split_image_ref` already strips the digest internally; partitioning `service.image` on `"@"` again duplicates ref parsing across two modules. Returning the digest from `_split_image_ref` keeps ref handling in one place, matching that helper's stated role as single source of truth. | Claude Code | Refactoring |
| M7 | `kamiwaza_extensions/compose_volumes.py:107` | `mountPath: /` reaches the "mountPath is missing" branch (`"/".rstrip("/") == ""`), so the error misdescribes the input. Reject root explicitly. | Claude Code | Error message |
| M8 | `kamiwaza_extensions/payload_builder.py:255-259` vs `267` | The new comprehension guards `isinstance(svc, dict)` while the loop immediately below calls `svc.get(...)` unguarded, as does `_find_primary_service`. Either non-dict services are possible (then the loop raises `AttributeError`) or they aren't (then the guard is dead). Pre-existing in the loop; the new guard makes the inconsistency visible. | Claude Code | Code quality |

## Test Coverage Assessment
**Verdict: adequate for the new module, with two named gaps and one assertion-free test.**

- **Well covered**: `compose_volumes.py` (217 lines of direct tests — parsing branches, both read-only spellings, long-form key aliases, malformed entries, DNS truncation, operator-name collision, and the exact/nested/trailing-slash/sibling-prefix persistence matrix); `kamiwaza_sdk/schemas/extensions.py` alias round-trip (`test_dev_state.py:770-781`); `_build_patch_service_spec` (digest pin, persistence forwarding, service filter, empty-list clear).
- **Gaps**: kamiwaza.json persistence precedence through `PayloadBuilder.build` (M3 — the seam where a regression already occurred); the `ValueError`/`ValidationError` handler (M4); the `--service` error message (M2, assertion-free); non-bool `enabled` (C1) and non-canonical targets (M1) once fixed; a profile-gated `--service` target reaching PATCH (H1 — Codex reproduced the `min_length` crash by execution, nothing pins it).

## Positive Observations
- The core modelling fix is right and the reasoning is load-bearing: volumes belong on the service because every service owns a pod template, and the emptyDir-over-PVC failure mode is explained where the decision lives rather than only in the commit message.
- Extracting `compose_volumes.py` out of `payload_builder.py` earned a real test module; parsing branches are now pinned directly instead of only through the payload builder, which is what made the nested-path and truncation cases testable at all.
- H2's underlying decision is the disciplined one. Discovering that `persistent_service_names` could never work (`Extension.services` is `List[ExtensionServiceStatus]`) and deleting the code rather than patching it, then stating the limitation at the decision point, beats shipping a clear that only passes against a hand-built `SimpleNamespace`.
- The `--service` PATCH filter fixes a genuine `ImagePullBackOff`, and the docstring explains why siblings must be excluded; it composes correctly with the existing catalog-overlay skip.
- `_OPERATOR_VOLUME_NAMES` pre-seeding survived the refactor with its rationale intact, and the collision-suffix behaviour is pinned by test.
- Test names carry the failure they prevent (`test_sibling_prefix_is_still_translated`, `test_enabled_without_mount_path_is_rejected`) rather than restating the method under test.
- All items from the three prior fix commits landed and are verifiable in the tree — nested-path guard, `mountPath` validation, `allowEmptyPersistenceInitialization` removal, persistence resolved once before volumes, `--service` validation at entry, and the payload-build exception wrapper.

## Overall Verdict
**Risk Level**: Medium
**Verdict**: REQUEST_CHANGES (inferred — no structured verdict line found)

The core fix is correct, well-factored, and three rounds of prior feedback are genuinely addressed. Blocking on C1: the guard this PR exists to build still fails open on a truthy non-bool `enabled` (confirmed by execution), re-opening the exact silent Postgres data loss for a plausible authoring shape — a three-line fix symmetric with the `mountPath` validation already accepted. H1 is close behind: both engines independently hit `--service` validation gaps, and Codex reproduced a post-build pydantic crash for a profile-gated target. Fix C1, tighten `--service` to deployable services, decide on a warning or state-file guard for H2, and close M2/M3, and this is ready.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>


> ℹ️ Verdict **REQUEST_CHANGES** was inferred from review content (no structured verdict line found).

**Reviewed-SHA**: 8abb59a55a55997b99637b3337a498b55e0f28f5

jxstanford (changes requested):
I'll synthesize the two engine reports into a single review.

# RE-REVIEW Report: feat(extensions): scope Compose volumes per service and forward persistence [ENG-9604]

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/243
**Branch**: `feat/compose-volume-persistence` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 9/10 | Per-service volume scoping matches the operator's pod-template ownership model; the two drive-by dev.py fixes are coupled to the same PATCH path and disclosed |
| Security | 9/10 | No injection/secret/authz surface; volume sources sanitized to DNS-1123, mountPaths forced absolute |
| Test Quality | 8/10 | `test_compose_volumes.py` is exemplary (261 lines, full persistence-guard matrix); `dev_state` persistence round-trip has zero tests and the state-ordering bug slips through helper-only coverage |
| Error Handling | 8/10 | `ValidationError`/`ValueError` caught with field-naming message instead of raw traceback; docked for validating after build+push and the state-write ordering |
| Code Quality | 8/10 | Small single-purpose functions, frozen dataclass, high-value comments; the duplicated path-containment predicate and the unrelated lockfile churn stand out |
| Refactoring | 8/10 | Extension-level `_build_volume_specs` properly replaced rather than patched around; the two copies of the containment predicate are the missed consolidation |
| Documentation Accuracy | N/A | No Markdown changed in this PR |
| Documentation Standards | N/A | No Markdown changed in this PR |

## Critical Issues (Must Fix)

None.

## High Priority Issues
| H1 | `kamiwaza_extensions/commands/dev.py:1080` (via `_record("build")` at :1091-1094) | `persistence_mounts=_deployed_persistence_mounts(payload)` is written **before** the PATCH runs, so orphan-persistence memory is overwritten with this run's *intent* rather than applied state. Three failure paths: (a) the warning is one-shot — run N warns correctly then erases `/data` from state, run N+1 sees `prior_mounts == {}` and is silent while the CR's PVC is still mounted; (b) a run that stops after build/push but before apply records a layout never applied; (c) under `--service X`, siblings that were never PATCHed get their mounts recorded as deployed. Record should be sticky until the CR is known to have moved, or written only on `apply`. | Claude Code, Codex (both engines, independently) | correctness |
| H2 | `kamiwaza-ai-extensions-lib/package-lock.json:746-1290` | 60 lines of `libc: ["glibc"]` / `libc: ["musl"]` selectors deleted from native optional deps (`@img/sharp-*`, `@next/swc-linux-*`). Without them npm cannot distinguish glibc from musl variants for the same CPU/OS and can reify an incompatible native package on Alpine or glibc runners. This is unrelated lockfile churn from a differing local npm version — regenerate with an npm that preserves `libc`, or revert the file from the PR. | Codex (single-source) | dependency-hygiene |

| M1 | `kamiwaza_sdk/schemas/extensions.py:169` + `dev.py:216-223` | `ImagePatch.digest` is a new wire field on the PATCH contract. The PR body justifies `persistence` against the operator but says nothing about `digest`. If the platform's request model rejects unknown fields, every dev PATCH 422s. Unit-covered, unverified against a deployed operator. | Claude Code | api-contract |
| M2 | `dev.py:216` | For a digest-only ref (`repo@sha256:…`, no tag) the PATCH sends `tag="latest"` alongside the digest — the `latest` is fabricated. Pre-existing, but this PR touches digest handling and adds no tag-less test. An operator reconstructing from `(registry, repository, tag)` would deploy a different image. | Claude Code | correctness |
| M3 | `kamiwaza_extensions/compose_volumes.py:83` | `target.rstrip("/")` is the only normalization. `//var/lib/postgresql/data` escapes the guard under a `/var/lib/postgresql` mountPath and emits an emptyDir over the PVC subtree. (`/x//y` and `/x/./y` are correctly caught.) | Claude Code | correctness |
| M4 | `compose_volumes.py:73-86` vs `dev.py:182-188` | `_covered_by_persistence` and `_under_mount_path` are the same predicate with the same trailing-slash/prefix rules, duplicated across modules. Drift would make the emptyDir guard and the collision warning disagree. | Claude Code | duplication |
| M5 | `compose_volumes.py:106-110` | `persistence.enabled: true` with no `mountPath` is a hard `ValueError`. If the operator's `PersistenceSpec` defaults `mountPath`, this rejects a config that works today. No `PersistenceSpec` reference exists in this repo to read back. | Claude Code | validation |
| M6 | `kamiwaza_sdk/services/extensions.py:79` (unchanged) | `create_extension` dumps without `exclude_none`, so the POST body now carries `persistence: null, volumes: null, volumeMounts: null` per service. Existing `env`/`command` already send nulls, but this is new wire surface with no test. | Claude Code | api-contract |
| M7 | repo-wide | `x-kamiwaza.persistence` is a new author-facing contract with a hard-fail validator, documented only in code comments — no `.md` or schema mention, and `validators/compose.py` doesn't know the key, so malformed blocks surface only after build+push. | Claude Code | documentation |

## Test Coverage Assessment
- **Well covered**: `compose_volumes.py` — parametrized tests pin every parse branch, alias spelling, read-only form, DNS truncation, operator-name collision, and the full persistence-guard matrix. Schema alias round-trip asserted. `_validate_service_filter` and service-filter PATCH exclusion covered. `warn_orphaned_persistence` / `_deployed_persistence_mounts` covered as isolated helpers.
- **Gaps**:
  - `dev_state.py` — zero `persistence` references in `test_dev_state.py`. The whole orphan-warning chain depends on `mark_step(persistence_mounts=…)` → `write_state` → `read_state` round-tripping, including the old-state-file path (`last_persistence_mounts` absent) that `dev.py:1058`'s `getattr(..., None) or {}` defends.
  - `dev.py:1057` wiring — H1 is invisible to current tests precisely because `warn_orphaned_persistence` is only exercised in isolation, never against a `_record`-mutated state or a partial (`--service`) run.
  - `_build_patch_service_spec` digest path for a tag-less `repo@sha256:…` ref.
- Full extensions unit suite: 1760 passed, 1 skipped (BSD-sed host issue, unrelated). Ruff clean.

## Positive Observations
- **Every finding from the previous rounds is demonstrably addressed.** The `persistence.get('enabled') is not True` identity check became `_persistence_enabled`, which raises on non-bool with a message naming the sibling `SANDBOX_PERSISTENCE` flag that makes the mistake reachable. The dual-source bypass is closed structurally: persistence resolves once in `PayloadBuilder._build_services` (kamiwaza.json ahead of Compose) and is *passed into* `build_service_volume_specs`.
- The unworkable "clear removed persistence" idea was dropped rather than half-implemented, with a docstring stating plainly why (`get_extension` returns only `ExtensionServiceStatus`).
- Comments explain *why*: the operator's `[tmp] + (data PVC) + svc.Volumes` reconcile model justifies both the reserved-name set and the empty-list-clears semantics; the `mountPath:` → `None` → truthy `"None"` stringify trap is called out at the guard.
- The nested-path case (`pgdata:/var/lib/postgresql/data` under a `/var/lib/postgresql` mountPath — the literal bug in the PR description) is caught, tested, and distinguished from the sibling-prefix case.
- `_split_image_ref` already stripped digests, so digest extraction composes cleanly rather than duplicating ref parsing.

## Overall Verdict
**Risk Level**: Medium
**Verdict**: REQUEST_CHANGES

Both engines independently landed on the same defect in `_record`'s ordering (H1), which converts the orphan-persistence warning from a durable safety net into a one-shot that erases its own evidence — and Codex extended it to the partial-apply and `--service` paths. H2 is a distinct, mechanically-fixable problem: the lockfile change strips glibc/musl selectors that npm needs to pick the right native binary, and it is unrelated to this PR's purpose. The core fix — per-service volumes with a persistence-aware guard — is correct and well-tested; these two items are what stand between it and merge.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

**Reviewed-SHA**: aac13d0f5b243ad5b8d99a24794e7469b98c4a9a

jxstanford (approved):
# RE-REVIEW Report: feat(extensions): scope Compose volumes per service and forward persistence [ENG-9604]

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/243
**Branch**: `feat/compose-volume-persistence` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 9/10 | Volumes correctly re-scoped from extension root to per-service pod templates; persistence precedence (kamiwaza.json ahead of compose) mirrors the established ENG-4832 healthCheck pattern; resolved-persistence is threaded into the volume builder rather than re-read, so a metadata-declared PVC still arms the emptyDir guard |
| Security | 9/10 | No new auth/secret surface; host-path and `${VAR}` interpolation rejection preserved via `looks_like_host_path`; operator-reserved `tmp`/`data` volume names still pre-seeded so a user volume cannot collide with operator-injected ones |
| Test Quality | 9/10 | 1768 passed / 1 skipped (skip is environmental — BSD vs GNU `sed`); new `test_compose_volumes.py` pins every parse branch, plus persistence-guard, mount-recording, and service-filter validation suites |
| Error Handling | 8/10 | `payload_builder.build` wrapped for `ValueError`/`ValidationError` with a named-field message instead of a post-push traceback; `--service` validated up front against profile-gated and build-less services; enabled-but-invalid `mountPath`/non-bool `enabled` fail loudly rather than silently disarming the guard |
| Code Quality | 8/10 | Good decomposition (`compose_volumes.py` extraction, `_find_primary_service`, `_append_platform_env`, `_add_service_overrides`); docstrings explain the *why*. Docked for the unrelated lockfile churn (H2) riding along in the diff |
| Refactoring | 9/10 | The payload-builder split is the right cut — parsing/translation now lives in one testable module and the builder only orchestrates |
| Documentation Accuracy | N/A | No Markdown changed in this PR |
| Documentation Standards | N/A | No Markdown changed in this PR |

## Critical Issues (Must Fix)
| — | — | None | — | — |

## High Priority Issues
| H1 | `kamiwaza_extensions/commands/dev.py:1051`, `:1126` | `--service` filters the PATCH path only. The create path (`create_extension(payload)`) and the `_delete_and_recreate` 405 fallback both send the full unfiltered payload, so a `--service backend` run against a missing/PATCH-less extension deploys siblings at revision tags that were never built or pushed → `ImagePullBackOff`. The new `_validate_service_filter` narrows *which* name is accepted but does not scope those two payloads | Claude Code (single-source) | correctness |
| H2 | `kamiwaza-ai-extensions-lib/package-lock.json:748` | 60 lines of `libc: ["glibc"]` / `["musl"]` selectors were dropped from Sharp/libvips and Next SWC optional native packages. npm uses these to disambiguate glibc vs musl variants that share the same `os`/`cpu`; without them an `npm ci` on Alpine (or Debian) can resolve the wrong native artifact. This is unrelated lockfile churn in a Python-scoped PR — regenerate with an npm version that preserves `libc`, or revert the file | Codex (single-source) | dependency-hygiene |

| M1 | `kamiwaza_extensions/compose_volumes.py:83`, `kamiwaza_extensions/commands/dev.py:195` | Target-path normalization is `rstrip("/")` only. `/var/lib/postgresql/./data`, `//var/lib/postgresql`, or a trailing-dot form will not match the persistence mountPath, so the guard silently passes and an emptyDir lands over the PVC subtree. A `posixpath.normpath` pass before comparison would close it. Note both call sites duplicate the same prefix logic (`_covered_by_persistence` and `_under_mount_path`) — worth unifying so they can't drift | Claude Code (single-source) | correctness |
| M2 | `kamiwaza_extensions/dev_state.py:210` | `last_persistence_mounts` is sticky by design (merge-only, never cleared), which is correct for keeping the orphan warning firing — but there is no path to reset it once the author has actually removed the Compose volume. The warning will keep firing on every subsequent deploy with no way to acknowledge it. Consider dropping the entry when the current payload declares no colliding mount | Claude Code (single-source) | usability |

## Test Coverage Assessment
Strong. The full extensions unit suite is green (1768 passed, 1 skipped; the skip is a BSD-`sed` host limitation in `test_sdk_override.py:1382`, unrelated to this change). New coverage is well-targeted at the risky paths: every short/long-form volume parse branch, read-only in all three spellings, DNS-label truncation and operator-name collision, the exact/nested/sibling-prefix persistence guard, non-bool `enabled` and non-absolute `mountPath` rejection, `--service` validation against profile-gated and image-only services, and the apply-only/sticky/service-scoped semantics of `last_persistence_mounts`.

Gaps matching the findings above: no test exercises the create or `_delete_and_recreate` path under `--service` (H1), and no test covers non-canonical path forms (`//`, `/./`) against the persistence guard (M1).

## Positive Observations
- The core fix is right and well-motivated: volumes belong to a service's pod template, not the extension root, and empty lists on PATCH deliberately clear stale Compose volumes.
- The persistence↔emptyDir interaction is handled at the nested level, not just exact equality — the stock Postgres `pgdata:/var/lib/postgresql/data` under a `/var/lib/postgresql` mountPath is exactly the case that would have silently shadowed a live PVC.
- Failing loudly on `enabled: "true"` is the correct call: the platform would coerce it and provision the PVC while this code's guard stayed disarmed. The docstring names the sibling `SANDBOX_PERSISTENCE` flag that makes the mistake reachable.
- Resolving persistence once in the caller and passing it down (rather than re-reading Compose inside `compose_volumes`) closes the kamiwaza.json-declared-PVC hole before it could exist.
- Digest preservation on `ImagePatch` and the `serialize_by_alias` config on both service specs fix real serialization drift (`volume_mounts` → `volumeMounts`) that the K8s API would have dropped.
- Error handling moved ahead of the traceback: a malformed persistence block now names the field instead of blowing up after the whole build-and-push phase.

## Overall Verdict
**Risk Level**: Low
**Verdict**: APPROVE

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

**Reviewed-SHA**: 7a8a69b16c4cc602f2559f0fc96a398dd923de2f

PR #245, opened 2026-08-03
refactor(sdk-override): split runtime classification into named steps [ENG-9606]

shkevin:
## Summary

`sdk_override/classification.py` resolved a service's runtime by walking Dockerfile stages, image tokens and `PYTHONPATH` inline. The functions were large enough that the precedence between those signals — which wins when a `python:` base image has a Node build stage on top — was hard to establish by reading.

Behavior is unchanged. This is a readability pass plus two small fixes that were sitting in the same tree.

## What changed

- The walk is split into named steps: `_classify_runtime_image`, `_parse_dockerfile_stages`, `_resolve_stage_base`, `_final_stage_body_start`, `_last_runtime_pythonpath`. Image tokens are lifted to `_PYTHON_IMAGE_TOKENS` / `_NODE_IMAGE_TOKENS` / `_STATIC_IMAGE_TOKENS` rather than being matched inline.
- New `test_sdk_override_runtime_classification.py` pins the classification outcomes, which were previously only covered indirectly through the override tests.
- `sdk_override/__init__.py` stops re-exporting about twenty private helpers that existed only so the old tests could reach them.

Two unrelated one-liners:

- `_replace_image_tag` stops at a `:${...}` Compose default-value placeholder instead of the last colon. `foo:${IMAGE_TAG:-local}` retagged to `foo:${IMAGE_TAG:2.0.0` — the `:-` inside the placeholder was read as the tag separator.
- `catalog_overlay` passes `preview_image` through to the overlay entry.

## Testing

`183 passed` across the override, classification, catalog-overlay and publish suites. `ruff` and `mypy` clean.

### DIFF
--- a/kamiwaza-ai-extensions-lib/package-lock.json
+++ b/kamiwaza-ai-extensions-lib/package-lock.json
@@ -746,6 +746,9 @@
                 "arm"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -763,6 +766,9 @@
                 "arm64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -780,6 +786,9 @@
                 "ppc64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -797,6 +806,9 @@
                 "riscv64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -814,6 +826,9 @@
                 "s390x"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -831,6 +846,9 @@
                 "x64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -848,6 +866,9 @@
                 "arm64"
             ],
             "dev": true,
+            "libc": [
+                "musl"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -865,6 +886,9 @@
                 "x64"
             ],
             "dev": true,
+            "libc": [
+                "musl"
+            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -882,6 +906,9 @@
                 "arm"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -905,6 +932,9 @@
                 "arm64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -928,6 +958,9 @@
                 "ppc64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -951,6 +984,9 @@
                 "riscv64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -974,6 +1010,9 @@
                 "s390x"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -997,6 +1036,9 @@
                 "x64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1020,6 +1062,9 @@
                 "arm64"
             ],
             "dev": true,
+            "libc": [
+                "musl"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1043,6 +1088,9 @@
                 "x64"
             ],
             "dev": true,
+            "libc": [
+                "musl"
+            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1239,6 +1287,9 @@
                 "arm64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "MIT",
             "optional": true,
             "os": [
@@ -1256,6 +1307,9 @@
                 "arm64"
             ],
             "dev": true,
+            "libc": [
+                "musl"
+            ],
             "license": "MIT",
             "optional": true,
             "os": [
@@ -1273,6 +1327,9 @@
                 "x64"
             ],
             "dev": true,
+            "libc": [
+                "glibc"
+            ],
             "license": "MIT",
             "optional": true,
             "os": [
@@ -1290,6 +1347,9 @@
                 "x64"
             ],
             "dev": true,
+            "libc": [
+                "musl"
+            ],
             "license": "MIT",
             "optional": true,
             "os": [
--- a/kamiwaza-ai-extensions-lib/package-lock.json
+++ b/kamiwaza-ai-extensions-lib/package-lock.json
@@ -746,9 +746,6 @@
                 "arm"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -766,9 +763,6 @@
                 "arm64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -786,9 +780,6 @@
                 "ppc64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -806,9 +797,6 @@
                 "riscv64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -826,9 +814,6 @@
                 "s390x"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -846,9 +831,6 @@
                 "x64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -866,9 +848,6 @@
                 "arm64"
             ],
             "dev": true,
-            "libc": [
-                "musl"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -886,9 +865,6 @@
                 "x64"
             ],
             "dev": true,
-            "libc": [
-                "musl"
-            ],
             "license": "LGPL-3.0-or-later",
             "optional": true,
             "os": [
@@ -906,9 +882,6 @@
                 "arm"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -932,9 +905,6 @@
                 "arm64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -958,9 +928,6 @@
                 "ppc64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -984,9 +951,6 @@
                 "riscv64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1010,9 +974,6 @@
                 "s390x"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1036,9 +997,6 @@
                 "x64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1062,9 +1020,6 @@
                 "arm64"
             ],
             "dev": true,
-            "libc": [
-                "musl"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1088,9 +1043,6 @@
                 "x64"
             ],
             "dev": true,
-            "libc": [
-                "musl"
-            ],
             "license": "Apache-2.0",
             "optional": true,
             "os": [
@@ -1287,9 +1239,6 @@
                 "arm64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "MIT",
             "optional": true,
             "os": [
@@ -1307,9 +1256,6 @@
                 "arm64"
             ],
             "dev": true,
-            "libc": [
-                "musl"
-            ],
             "license": "MIT",
             "optional": true,
             "os": [
@@ -1327,9 +1273,6 @@
                 "x64"
             ],
             "dev": true,
-            "libc": [
-                "glibc"
-            ],
             "license": "MIT",
             "optional": true,
             "os": [
@@ -1347,9 +1290,6 @@
                 "x64"
             ],
             "dev": true,
-            "libc": [
-                "musl"
-            ],
             "license": "MIT",
             "optional": true,
             "os": [
--- a/kamiwaza_extensions/catalog_overlay.py
+++ b/kamiwaza_extensions/catalog_overlay.py
@@ -26,6 +26,10 @@ from urllib.parse import quote
 
 import yaml
 
+# Safe for the dev path: registry_builder imports only stdlib, yaml and
+# packaging — none of the forbidden publishing machinery (see TestImportInvariant).
+from kamiwaza_extensions.registry_builder import _normalize_preview_image
+
 OVERLAY_PATH = "/apps/app_templates/catalog/overlay"
 
 # Matches the platform's app_templates.version column (String(40)).
@@ -174,6 +178,12 @@ def build_overlay_entry(
         value = metadata.get(key)
         if isinstance(value, expected_type):
             entry[key] = value
+    if isinstance(entry.get("preview_image"), str):
+        # The publish path stores this normalized, so the shorthand
+        # ``preview_image: "screenshot.png"`` becomes ``images/screenshot.png``
+        # in the catalog. Writing the raw value here would overwrite that row
+        # with a key the asset was never uploaded under.
+        entry["preview_image"] = _normalize_preview_image(entry["preview_image"])
     ext_type = metadata.get("type") or metadata.get("template_type")
     if isinstance(ext_type, str) and ext_type in ("app", "tool", "service"):
         entry["template_type"] = ext_type
--- a/kamiwaza_extensions/catalog_overlay.py
+++ b/kamiwaza_extensions/catalog_overlay.py
@@ -160,6 +160,7 @@ def build_overlay_entry(
         ("license", str),
         ("homepage", str),
         ("image", str),
+        ("preview_image", str),
         ("kamiwaza_version", str),
         ("preferred_model_type", str),
         ("preferred_model_name", str),
--- a/kamiwaza_extensions/compose_transformer.py
+++ b/kamiwaza_extensions/compose_transformer.py
@@ -33,24 +33,39 @@ _IMAGE_BASENAME_RE = re.compile(r"^[a-z0-9]([-a-z0-9._/]{0,126}[a-z0-9])?$")
 Purpose = Literal["dev", "publish"]
 
 
+def _tag_separator_index(ref: str) -> int:
+    """Return the index of *ref*'s tag-separating colon, or -1 if untagged.
+
+    Two colons can appear before the tag: a registry port
+    (``localhost:5000/foo``) and a Compose default-value placeholder
+    (``foo:${IMAGE_TAG:-local}``). The port is disambiguated by the last
+    ``/``; the placeholder has to win over ``rfind`` because the ``:`` in
+    ``:-`` sits *after* the real tag separator.
+
+    Single source of truth for ``_replace_image_tag`` and
+    ``_split_image_ref`` — they disagreed once, and the malformed ref that
+    produced (``foo:${IMAGE_TAG:2.0.0``) reached the K8s PATCH path.
+    """
+    last_slash = ref.rfind("/")
+    placeholder = ref.find(":${", last_slash + 1)
+    if placeholder >= 0:
+        return placeholder
+    last_colon = ref.rfind(":")
+    return last_colon if last_colon > last_slash else -1
+
+
 def _replace_image_tag(image_ref: str, new_tag: str) -> str:
     """Return *image_ref* with its tag (and any digest) replaced by *new_tag*.
 
     The namespace (registry + repo path) is preserved verbatim. Handles
-    refs that include a registry port (``localhost:5000/foo:tag``) by
-    using the position of the last ``/`` to disambiguate the port colon
-    from the tag colon, and strips any ``@sha256:...`` suffix before
-    re-tagging.
+    refs that include a registry port (``localhost:5000/foo:tag``) or a
+    Compose default-value placeholder (``foo:${IMAGE_TAG:-local}``), and
+    strips any ``@sha256:...`` suffix before re-tagging.
     """
     ref = image_ref.split("@", 1)[0]
-    last_slash = ref.rfind("/")
-    placeholder_tag = ref.find(":${", last_slash + 1)
-    if placeholder_tag >= 0:
-        ref = ref[:placeholder_tag]
-    else:
-        last_colon = ref.rfind(":")
-        if last_colon > last_slash:
-            ref = ref[:last_colon]
+    separator = _tag_separator_index(ref)
+    if separator >= 0:
+        ref = ref[:separator]
     return f"{ref}:{new_tag}"
 
 
@@ -91,11 +106,10 @@ def _split_image_ref(image_ref: str) -> Tuple[Optional[str], str, str]:
     repository, that mismatch would produce ImagePullBackOff.
     """
     ref = image_ref.split("@", 1)[0]
-    last_slash = ref.rfind("/")
-    last_colon = ref.rfind(":")
-    if last_colon > last_slash:
-        namespace = ref[:last_colon]
-        tag = ref[last_colon + 1 :]
+    separator = _tag_separator_index(ref)
+    if separator >= 0:
+        namespace = ref[:separator]
+        tag = ref[separator + 1 :]
     else:
         namespace = ref
         tag = "latest"
--- a/kamiwaza_extensions/compose_transformer.py
+++ b/kamiwaza_extensions/compose_transformer.py
@@ -44,9 +44,13 @@ def _replace_image_tag(image_ref: str, new_tag: str) -> str:
     """
     ref = image_ref.split("@", 1)[0]
     last_slash = ref.rfind("/")
-    last_colon = ref.rfind(":")
-    if last_colon > last_slash:
-        ref = ref[:last_colon]
+    placeholder_tag = ref.find(":${", last_slash + 1)
+    if placeholder_tag >= 0:
+        ref = ref[:placeholder_tag]
+    else:
+        last_colon = ref.rfind(":")
+        if last_colon > last_slash:
+            ref = ref[:last_colon]
     return f"{ref}:{new_tag}"
 
 
--- a/kamiwaza_extensions/sdk_override/__init__.py
+++ b/kamiwaza_extensions/sdk_override/__init__.py
@@ -25,7 +25,10 @@ from kamiwaza_extensions.sdk_override.build import (
     generate_build_overrides,
     generate_local_build_dockerfile_patches,
 )
-from kamiwaza_extensions.sdk_override.classification import detect_service_runtime
+from kamiwaza_extensions.sdk_override.classification import (
+    detect_service_runtime,
+    detect_service_type,
+)
 from kamiwaza_extensions.sdk_override.compose import (
     generate_compose_override,
 )
@@ -52,6 +55,7 @@ __all__ = [
     "validate_sdk_override",
     # classification
     "detect_service_runtime",
+    "detect_service_type",
     # compose
     "generate_compose_override",
     # build
--- a/kamiwaza_extensions/sdk_override/__init__.py
+++ b/kamiwaza_extensions/sdk_override/__init__.py
@@ -21,45 +21,19 @@ from __future__ import annotations
 
 from kamiwaza_extensions.sdk_override.build import (
     BuildOverride,
-    _PYTHON_OVERLAY,
-    _PYTHON_PIP_INSTALL_PATTERN,
-    _PYTHON_PRE_INSTALL_STRIP,
-    _TS_BUILD_PATTERNS,
-    _TS_NPM_CI_LINE_PATTERN,
-    _TS_NPM_INSTALL_PATTERN,
-    _TS_OVERLAY,
-    _TS_PRE_INSTALL_STRIP,
-    _find_active_user,
-    _insert_before_install_pattern,
-    _restore_user_block,
     apply_build_overlay,
     generate_build_overrides,
     generate_local_build_dockerfile_patches,
 )
-from kamiwaza_extensions.sdk_override.classification import (
-    _detect_build_service_runtime,
-    _image_basename,
-    _read_dockerfile_stage_bases,
-    _read_final_base_image,
-    _resolve_dockerfile,
-    detect_service_runtime,
-    detect_service_type,
-)
+from kamiwaza_extensions.sdk_override.classification import detect_service_runtime
 from kamiwaza_extensions.sdk_override.compose import (
-    _SDK_BIND_TARGET,
-    _TS_LIB_NODE_MODULES_TARGET,
-    _TS_LIB_PACKAGE_DIR,
     generate_compose_override,
 )
 from kamiwaza_extensions.sdk_override.spec import (
     SdkOverrideSpec,
     ValidationResult,
-    _CONFIG_DIR,
-    _CONFIG_FILE,
-    _newest_mtime,
     build_typescript_lib,
     check_buildkit_available,
-    console,
     is_typescript_dist_stale,
     print_override_diagnostics,
     resolve_sdk_override,
@@ -78,7 +52,6 @@ __all__ = [
     "validate_sdk_override",
     # classification
     "detect_service_runtime",
-    "detect_service_type",
     # compose
     "generate_compose_override",
     # build
--- a/kamiwaza_extensions/sdk_override/classification.py
+++ b/kamiwaza_extensions/sdk_override/classification.py
@@ -20,6 +20,8 @@ import re
 from pathlib import Path
 from typing import Any, List, Literal, Optional
 
+from kamiwaza_extensions.compose_ports import extract_container_port
+
 ServiceRuntime = Literal["frontend", "backend", "static", "other"]
 
 _PYTHON_IMAGE_TOKENS = ("python",)
@@ -28,15 +30,47 @@ _STATIC_IMAGE_TOKENS = ("nginx", "caddy", "httpd", "apache")
 _PYTHONPATH_ENV_RE = re.compile(
     r"""^\s*ENV\s+
         (?:.*?\s+)?              # allow other ENV pairs before ours
-        PYTHONPATH\s*=\s*
+        PYTHONPATH
+        (?:\s*=\s*|\s+)          # = or space (legacy ``ENV PYTHONPATH /val``)
         ("?)([^"\s]+)\1          # quoted-or-bare value
     """,
     re.IGNORECASE | re.VERBOSE,
 )
 
 
+def detect_service_type(
+    svc_name: str,
+    svc_config: dict,
+) -> ServiceRuntime:
+    """Heuristic: classify a compose service as 'frontend' or 'backend'.
+
+    Uses service name, port, and Dockerfile path as signals. This is the
+    fallback whenever no Dockerfile base image is readable, so a service
+    that only declares ``build: ./backend`` still classifies.
+    """
+    name_lower = svc_name.lower()
+    if any(token in name_lower for token in ("frontend", "ui", "web")):
+        return "frontend"
+
+    build = svc_config.get("build", {})
+    if isinstance(build, dict):
+        dockerfile = str(build.get("dockerfile", "")).lower()
+        context = str(build.get("context", "")).lower()
+        if "frontend" in dockerfile or "frontend" in context:
+            return "frontend"
+
+    # Uses the shared compose-port parser so all three spec shapes (bare,
+    # host:container, long-form dict) classify the same way. Mirrors
+    # ``_should_use_node_frontend_probe``.
+    for port_spec in svc_config.get("ports", []):
+        if extract_container_port(port_spec) in (3000, 3001):
+            return "frontend"
+
+    return "backend"
+
+
 def detect_service_runtime(
-    _svc_name: str,
+    svc_name: str,
     svc_config: dict,
     *,
     extension_dir: Optional[Path] = None,
@@ -47,18 +81,23 @@ def detect_service_runtime(
     - ``frontend`` for likely Node/Next-style services
     - ``backend`` for likely Python/backend services
     - ``static`` for generic web servers such as nginx/caddy/httpd
-    - ``other`` for a readable Dockerfile with a non-SDK runtime
 
     When the Dockerfile is available, prefer its final base image over naming
-    heuristics so converted static apps do not get a Python SDK overlay.
+    heuristics so converted static apps do not get a Python SDK overlay. An
+    unreadable Dockerfile, or a base image matching no known runtime token,
+    falls back to ``detect_service_type``.
     """
     dockerfile = _service_dockerfile(svc_config, extension_dir)
     base_image = _read_final_base_image(dockerfile)
-    return _classify_runtime_image(base_image) if base_image else "other"
+    if base_image:
+        runtime = _classify_runtime_image(base_image)
+        if runtime != "other":
+            return runtime
+    return detect_service_type(svc_name, svc_config)
 
 
 def _detect_build_service_runtime(
-    _svc_name: str,
+    svc_name: str,
     svc_config: dict,
     *,
     extension_dir: Optional[Path] = None,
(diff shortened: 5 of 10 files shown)


jxstanford (changes requested):
# Review Report: refactor(sdk-override): split runtime classification into named steps

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/245
**Branch**: `refactor/sdk-override-classification` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

> Note: reported CI is green, but Claude Code reproduced **5 failures locally** in `tests/unit/extensions/test_convert_agent.py` at `ec1b7f6` (`5 failed, 1681 passed, 1 skipped`) against **62/62 passing** for the same file at base `d0f9aa0`. See Test Coverage Assessment.

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 4/10 | PR is labelled a "readability pass" with "behavior is unchanged", but deletes the name/port classification fallback and a documented public symbol — deliberate semantic changes buried in a refactor. |
| Security | 8/10 | No injection, secret, authz, or deserialization surface touched; the pre-existing absolute-`dockerfile` read in `_resolve_dockerfile` is unchanged by this PR. |
| Test Quality | 3/10 | New test file pins only the two negative (`"other"`) outcomes; 5 existing tests left failing; the reported pass count came from four hand-picked suites that exclude the breaking file. |
| Error Handling | 5/10 | `_read_dockerfile`'s None-on-OSError is clean, but the widened `"other"` bucket turns a previously-working overlay into a silent no-op with zero diagnostic output. |
| Code Quality | 8/10 | Decomposition, naming, token-constant extraction, and the `ServiceRuntime` `Literal` are genuine improvements; minor stale docstring and a redundant `.lower()`. |
| Refactoring | 6/10 | Good structural work, but the `:${...}` fix landed in only one of two functions explicitly documented as mirrors, leaving the `dev` path emitting the malformed ref. |

## Critical Issues (Must Fix)
| 1 | *(repo policy — branch/title/body/commits)* | **Missing Linear ticket reference** — Kamiwaza repos require a related Linear ticket reference (e.g. ENG-1234) in the branch name, PR title, PR body, or commit messages. | Policy gate | Process |
| 2 | `kamiwaza_extensions/sdk_override/classification.py:57` | `detect_service_runtime` no longer falls back to `detect_service_type` (name/port/path heuristics) and returns `"other"` when no Dockerfile resolves — including the common `extension_dir=None` default. Semantic change, not a refactor. Breaks 4 tests with `KeyError: 'backend'`. In production, any service whose Dockerfile isn't at the conventional `context/Dockerfile` path (`dockerfile_inline`, a Containerfile named elsewhere) silently loses its SDK overlay under `kz-ext dev --sdk-repo`. | Claude Code, Codex | Correctness / Regression |
| 3 | `kamiwaza_extensions/sdk_override/classification.py:31` | `_PYTHONPATH_ENV_RE` dropped the `(?:\s*=\s*\|\s+)` alternation, so Docker's valid legacy `ENV PYTHONPATH /val` space-separated form no longer matches — the removed comment explicitly called this form out as supported. Breaks `test_pythonpath_legacy_env_form_in_dockerfile`; src-layout images using the legacy form silently lose `/app/src` from `PYTHONPATH`. | Claude Code, Codex | Correctness / Regression |

## High Priority Issues
| 4 | `kamiwaza_extensions/compose_transformer.py:95` | The `:${...}` fix was applied to `_replace_image_tag` only. `_split_image_ref` — whose docstring says it "Mirrors `_replace_image_tag`'s tag detection" — still uses bare `rfind(":")`. Reproduced: `_canonical_build_ref(..., purpose="dev")` on `ghcr.io/my-org/foo:${IMAGE_TAG:-local}` yields `localhost:5000/my-org/foo:${IMAGE_TAG:2.0.0-dev` — exactly the malformed ref the PR claims to fix. The K8s PATCH path (`commands/dev.py:97`) and `validators/metadata.py` share the helper. | Claude Code (single-source, reproduced) | Correctness |
| 5 | `kamiwaza_extensions/sdk_override/classification.py:76` | `_detect_build_service_runtime` likewise returns `"other"` instead of delegating, so build-time overlay insertion is silently skipped for the same unresolvable-Dockerfile cases — on the cluster-deploy path. | Claude Code (single-source) | Correctness / Regression |
| 6 | `kamiwaza_extensions/sdk_override/classification.py:57`, `compose.py:68`, `build.py:275` | A service with a `build` key that classifies as `"other"` is dropped with no diagnostic. `dev_local.py` / `dev.py` already print override diagnostics; the developer gets a silent no-op with no signal that classification failed. Given #2 widens the `"other"` set considerably, a warning is now load-bearing. | Claude Code (single-source) | Silent failure / Error handling |
| 7 | `kamiwaza_extensions/sdk_override/__init__.py:54` | `detect_service_type` was a public name in `__all__` and is deleted outright with no deprecation shim. Dropping the ~20 underscore-prefixed re-exports is fine; removing a documented public symbol from a published SDK package in a "readability pass" is not. | Claude Code (single-source) | API compatibility |

| 8 | `kamiwaza_extensions/sdk_override/classification.py:52` | Docstring still reads "prefer its final base image over naming heuristics" — there are no naming heuristics left. Stale after the deletion. | Claude Code | Docs |
| 9 | `kamiwaza_extensions/sdk_override/classification.py:39,61` | Renaming the first positional parameter of the public `detect_service_runtime` to `_svc_name` breaks keyword callers and is unusual for an exported function. Keep `svc_name`, or drop it in a deliberate signature change. | Claude Code | API / Convention |
| 10 | `kamiwaza_extensions/catalog_overlay.py:163` | `preview_image` is forwarded verbatim, but the published-catalog path applies `_normalize_preview_image` (`registry_builder.py:772`) to prefix `images/`. A `kamiwaza.json` declaring `preview_image: "screenshot.png"` yields `screenshot.png` in the overlay vs `images/screenshot.png` when published — contradicting the block's own "behaves like a published catalog entry" comment. The new test uses an already-prefixed path, so it doesn't catch this. | Claude Code | Consistency |
| 11 | `kamiwaza_extensions/compose_transformer.py:47` | `find(":${", last_slash + 1)` misses a whole-ref placeholder with no registry: `${IMAGE:-foo}` still retags to `${IMAGE:2.0.0`. Same failure shape as the bug being fixed. | Claude Code | Edge case |
| 12 | `kamiwaza_extensions/sdk_override/classification.py:89` | `_contains_token` lowercases its input, but every caller passes the output of `_image_basename`, which already lowercases. Harmless, redundant. | Claude Code | Code quality |

## Test Coverage Assessment
- **Verdict**: REQUEST_CHANGES (inferred — no structured verdict line found)
- **Reproduced failures** (base green, PR red) in `tests/unit/extensions/test_convert_agent.py`: `test_pythonpath_prepend_env_var_appends_to_sdk`, `test_pythonpath_default_when_env_not_set`, `test_pythonpath_preserves_compose_environment_value`, `test_pythonpath_preserves_compose_env_list_form` (all `KeyError: 'backend'`, from #2), plus `test_pythonpath_legacy_env_form_in_dockerfile` (from #3).
- **Adequately covered**: `catalog_overlay.py` forwarding; `_replace_image_tag` for the `:-` default form.
- **Gaps**:
  - `test_sdk_override_runtime_classification.py` asserts only the two `"other"` outcomes. The `backend` / `frontend` / `static` branches of `_classify_runtime_image` and the multi-stage Node→nginx promotion in `_detect_build_service_runtime` remain covered only indirectly — the exact gap the PR set out to close.
  - `_resolve_stage_base` — the alias-cycle guard (`base_ref in seen`) has no direct test.
  - `read_runtime_pythonpath` — legacy `ENV PYTHONPATH /val` coverage exists only in the suite that wasn't run.
  - `_split_image_ref` — no placeholder-tag case (#4).
  - `_replace_image_tag` — no `foo:${TAG}` (no default) or whole-ref placeholder case (#11).
- **Process gap**: 20 existing call sites in `test_sdk_override.py` were edited to add `extension_dir=_write_runtime_dockerfiles(...)`. Tests needing new inputs to keep passing *is* the signal that behavior changed; that should have triggered a full-suite sweep rather than a targeted four-suite run.

## Positive Observations
- The decomposition is genuinely good: `_read_dockerfile` / `_parse_dockerfile_stages` / `_resolve_stage_base` / `_final_stage_body_start` / `_last_runtime_pythonpath` are well-named, single-purpose, and each fits on a screen. The precedence question the PR set out to answer is now readable.
- Lifting image tokens into `_PYTHON_IMAGE_TOKENS` / `_NODE_IMAGE_TOKENS` / `_STATIC_IMAGE_TOKENS` removes three duplicated inline matches and makes ordering explicit.
- The `ServiceRuntime` `Literal` alias is a real typing improvement over bare `str`.
- Hoisting the `PYTHONPATH` regex to module scope removes a per-call `re.compile` and a function-local `import re`.
- The `_replace_image_tag` fix is correct for every realistic ref exercised (registry-with-port, digest-suffixed, `${REG:-localhost:5000}/foo:1.0`, `foo:${A}-${B}`).
- Converting private helpers from package re-exports to direct module imports in tests is the right direction.
- `ruff check` is clean on all changed files.

## Overall Verdict
**Risk Level**: High
**Verdict**: REQUEST_CHANGES

Both engines independently identified the same two regressions in `classification.py`: the removed name/port fallback (now returning `"other"`, including for the public `extension_dir=None` default) and the lost legacy `ENV PYTHONPATH /val` parse. Both degrade to silent no-ops in production rather than errors, and together they break 5 tests that pass on base. Separately, the `:${...}` retag fix was applied to `_replace_image_tag` but not to `_split_image_ref`, so the `dev` build path still emits the malformed ref the PR claims to have fixed. The missing Linear ticket reference is an independent hard gate.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>


> ℹ️ Verdict **REQUEST_CHANGES** was inferred from review content (no structured verdict line found).

**Reviewed-SHA**: ec1b7f668fa38a0056aa13beec6e26ebb9adc275

jxstanford (changes requested):
# RE-REVIEW Report: refactor(sdk-override): split runtime classification into named steps [ENG-9606]

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/245
**Branch**: `refactor/sdk-override-classification` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

**Previous-round feedback — all addressed:** `detect_service_type` restored (`classification.py:41-69`) with the service-heuristic fallback rewired at `:96` and pinned by two new tests; `_tag_separator_index` unifies the tag-boundary rule for `_replace_image_tag` / `_split_image_ref` with an agreement test; the legacy `ENV PYTHONPATH /val` space form is back in `_PYTHONPATH_ENV_RE`; `detect_service_type` re-added to `__all__` with a test pinning it; package-lock churn dropped. Local run: `188 passed, 1 skipped`, `ruff check` clean.

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 7/10 | Refactor is faithful and behavior-preserving, but `_warn_no_overlay` is an undocumented third behavior change in a PR described as "behavior unchanged… plus two small fixes" |
| Security | 9/10 | No new attack surface; all inputs are the developer's own compose/Dockerfile. Pre-existing absolute-`dockerfile`-path passthrough at `classification.py:167` untouched |
| Test Quality | 7/10 | New classification tests pin real precedence, but two new branches are uncovered, the warning is unasserted, and the consolidated heuristic test loses named cases |
| Error Handling | 8/10 | `_read_dockerfile` centralizes the `OSError`/missing-file path cleanly; the weak spot is misleading warning text, not a swallowed error |
| Code Quality | 8/10 | Good decomposition and naming; minor inconsistencies (inline name tokens, `str` where `ServiceRuntime` now exists, stale docstring) |
| Refactoring | 8/10 | Duplication genuinely removed on both the classification and tag-boundary axes; one placeholder gap remains in the new shared helper |
| Documentation Accuracy | N/A | No Markdown changed in this PR |
| Documentation Standards | N/A | No Markdown changed in this PR |

## Critical Issues (Must Fix)
| — | — | None | — | — |

## High Priority Issues
| 1 | `kamiwaza_extensions/sdk_override/compose.py:93,106` | New `_warn_no_overlay` prints `No SDK overlay for <svc> (runtime classified as 'static'…)` for multi-stage Node→nginx frontends — which **do** receive the TS overlay via the build-time path. Reproduced: `generate_compose_override` warns while `generate_local_build_dockerfile_patches` returns `['frontend']` for the same service; both run in the same `dev local --sdk-repo` flow (`dev_local.py:253,275`). The warning's own docstring says silence would make a dev "debug a container still running the published one" — as written it causes exactly that misdirection for a shape the module docstring and `test_sdk_override.py:1261` explicitly support. Fix: consult `_detect_build_service_runtime` before warning, or drop the warning from this readability PR. | Claude Code (single-source) | correctness / UX |

| 2 | `compose_transformer.py:50` | `_tag_separator_index` is incomplete for the defect class it names: an **untagged** ref whose last path segment is a placeholder still mis-splits — `${IMAGE_NAME:-myapp}` retags to `${IMAGE_NAME:2.0.0`, and `localhost:5000/${IMG:-foo}` yields repository `${IMG`. Reproduced; pre-existing, not a regression, but the helper is billed "single source of truth". | Claude Code | correctness |
| 3 | `classification.py:115,127` | The two `_detect_build_service_runtime` fallbacks added this round are the only uncovered branches (`--cov` misses 115, 127). Their twins in `detect_service_runtime` got tests; the build-path ones didn't — so the exact regression class under re-review can silently return. | Claude Code | test-coverage |
| 4 | `tests/unit/extensions/test_sdk_override_runtime_classification.py:64` | 12 named `TestDetectServiceType` cases collapsed into one 5-assert function. Drops the ENG-5954 long-form dict-port assertion (`{"target": 3000}`) and the `build.dockerfile` signal case; failures now report `assert 'backend' == 'frontend'` with no name, and the first failure masks the rest. Partly mitigated by `test_payload_builder.py:107`. | Claude Code | test-quality |
| 5 | `catalog_overlay.py:163` | New `preview_image` forwarding copies the raw metadata value, while the publish path normalizes the supported shorthand (`screenshot.png` → `images/screenshot.png`) via `registry_builder._normalize_preview_image`. Dev shadow rows can therefore point at a path published entries never use. Normalize before including it. | Codex (single-source) | consistency / correctness |
| 6 | `compose.py:93` | Console I/O introduced inside `generate_compose_override`, otherwise a pure dict builder called from several paths. `_warn_no_overlay` has no assertion-level test — lines execute incidentally; nothing pins the message or the quiet-vs-loud decision. | Claude Code | architecture |
| 7 | `compose_transformer.py:97` | `_split_image_ref` docstring still claims it "Mirrors `_replace_image_tag`'s tag detection (last `:` past the last `/`)" — inaccurate after the shared-helper extraction. | Claude Code | doc-drift |
| 8 | `classification.py:52`, `compose.py:98,139` | Inconsistent with the PR's own pattern: name tokens `("frontend","ui","web")` left inline while image tokens were lifted to `_*_IMAGE_TOKENS`; `_warn_no_overlay` / `_override_for` take `svc_type: str` rather than the new `ServiceRuntime` alias. | Claude Code | consistency |

## Test Coverage Assessment
- **Adequate**: `classification.py` (95% — `detect_service_runtime` precedence, base-image-wins-over-name, unrecognized-base and missing-Dockerfile fallbacks all pinned); `compose_transformer.py` (placeholder retag, split, and the both-helpers-agree invariant); `catalog_overlay.py` (`preview_image` forwarded and asserted).
- **Gaps**: `classification.py:115,127` — `_detect_build_service_runtime` fallbacks uncovered (#3); `compose.py:98-113` — `_warn_no_overlay` has no assertion on message or on the frontend/backend quiet path (#1/#6); the deleted ENG-5954 dict-port and `build.dockerfile` assertions on `detect_service_type` should be restored (#4); no test pins `preview_image` shorthand normalization through the overlay path (#5).

## Positive Observations
- Round-2 fixes are each pinned by a test rather than just patched: `test_unrecognized_runtime_base_falls_back_to_service_heuristics` and `test_missing_dockerfile_falls_back_to_service_heuristics` lock the regressed fallback; `test_detect_service_type_stays_publicly_exported` asserts `__all__` membership so the symbol can't be dropped again.
- `_tag_separator_index` is the right response to "two helpers disagreed" — one rule, both call sites converted, and `test_both_helpers_agree_on_the_namespace` encodes the invariant rather than the two outputs separately. The docstring records the failure mode that motivated it.
- The `_split_image_ref` change improves beyond the stated fix: `_repo_part` now yields `ghcr.io/my-org/foo` instead of `ghcr.io/my-org/foo:${IMAGE_TAG`, so sibling-service matching in `publish.py:143` and `dev_env_image_refs.py` works on placeholder-tagged compose refs.
- Decomposition is genuinely clearer — `_read_dockerfile` / `_parse_dockerfile_stages` / `_resolve_stage_base` untangle the nested closure, and lifting `_PYTHONPATH_ENV_RE` to module scope removes a per-call `re.compile` and a function-local `import re`.
- Dropping ~20 private re-exports from `__init__.py` is safe: no non-test consumer imports them, and `console` (the one non-underscore removal) is imported from `spec` directly.

## Overall Verdict
**Risk Level**: Low

Every previous comment was addressed properly and pinned with tests, and the refactor itself is clean and behavior-preserving. No Critical issues, no quality-gate failures, the Linear ticket (ENG-9606) is present, and every applicable dimension scores ≥ 7 — so this clears the approval gate. Note that Claude Code's own verdict was REQUEST_CHANGES on the strength of High #1; that finding is real and worth fixing before merge (a single condition plus a test), but under the stated decision boundaries a High-severity issue does not block. Medium #5 (Codex) is the only finding not in Claude's set and is a cheap parity fix in the same area.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

**Reviewed-SHA**: 2a8092e794aa82960c7813b7344e58a5ce14cdf6

jxstanford (changes requested):
# RE-REVIEW Report: refactor(sdk-override): split runtime classification into named steps [ENG-9606]

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/245
**Branch**: `refactor/sdk-override-classification` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

Locally: `1692 passed, 1 skipped` across `tests/unit/extensions/` (the skip is a BSD-sed host guard). `ruff` clean on all changed files; the 54 repo-wide ruff errors are pre-existing in untouched files. `mypy` unverified locally (not installed); CI reports clean.

**Delta since last review — all three fix commits verified:**

| Commit | Prior finding | Status |
|---|---|---|
| `d0091e6` | `detect_service_type` + both fallback call sites dropped; legacy `ENV PYTHONPATH /val` broken | **Fixed** — fallback restored at `classification.py:96`, `:126-129`; regex accepts `(?:\s*=\s*|\s+)` at `:34`, covered by `test_convert_agent.py:1469` |
| `a3b8a57` | `_split_image_ref` still used bare `rfind(":")` while `_replace_image_tag` was fixed | **Fixed** — both route through `_tag_separator_index` (`compose_transformer.py:66,109`), verified empirically |
| `a3b8a57` | `detect_service_type` removed from `__all__` (breaking change) | **Fixed** — re-exported and pinned by `test_detect_service_type_stays_publicly_exported` |
| `a002a48` | Warning added in `a3b8a57` misfired on Node→nginx frontends | **Fixed** — removed; `build.py` is net-unchanged vs. develop |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 9/10 | Scope matches the stated readability goal; the two declared "unrelated one-liners" are genuinely small, and the three fix commits stayed inside scope |
| Security | 9/10 | Nothing introduced; the tag-boundary fix closes a malformed-ref path that reached the K8s PATCH. Pre-existing unconfined Dockerfile path resolution left untouched (byte-identical to develop) |
| Test Quality | 7/10 | New file pins classification precedence well and adds a public-symbol regression guard, but `_detect_build_service_runtime`'s new `other` branch and the alias-cycle guard are unpinned |
| Error Handling | 9/10 | OSError paths preserved and centralized in one `_read_dockerfile`; fallbacks made explicit; warning removal well-reasoned |
| Code Quality | 8/10 | Clean decomposition and naming; weak spots are substring token semantics, the redundant re-parse, and the un-normalized `preview_image` passthrough |
| Refactoring | 8/10 | `_tag_separator_index` is the standout — it removes the drift class, not just the instance. Missed: the double-parse and the token matching the extracted constants were meant to clarify |
| Documentation Accuracy | N/A | No Markdown changed |
| Documentation Standards | N/A | No Markdown changed |

## Critical Issues (Must Fix)

None.

## High Priority Issues
| 1 | `kamiwaza_extensions/catalog_overlay.py:163` | `preview_image` is forwarded verbatim into the dev overlay entry, but the publish path normalizes it via `_normalize_preview_image` (`registry_builder.py:185-186`) — so the supported shorthand `preview_image: "screenshot.png"` publishes as `images/screenshot.png` but overlays as `screenshot.png`, overwriting the correct catalog row value. Workrooms launched from the overlay then reference the preview asset under the wrong key. Normalize before adding to the entry. | Codex (single-source) | Correctness / Consistency |

| 1 | `sdk_override/classification.py:28`, `:134` | `"bun" in "ubuntu"` is `True`, so `FROM ubuntu:22.04` classifies as `frontend`. A Python service on a plain `ubuntu` base gets the **TypeScript** overlay under `--sdk-repo` instead of falling through to the name/port heuristic (which would say `backend`). Pre-existing — the old inline `"bun" in base` had the same flaw — so "behavior unchanged" holds, but the tokens are now named constants whose stated purpose is to make these signals legible, and this contradicts `a002a48`'s verification note. Suggest exact-or-boundary matching for `bun`. | Claude Code (single-source; verified by executing `_classify_runtime_image`) | Correctness |
| 2 | `compose_transformer.py:50` | `_tag_separator_index` detects `:${` but not a `:` *inside* an already-open `${…}` span. `${IMAGE_NAME:-myapp}` (placeholder in the **name** position, untagged) retags to `${IMAGE_NAME:2.0.0` and splits to namespace `${IMAGE_NAME` / tag `-myapp}` — the exact malformed shape this helper exists to prevent. Pre-existing and rarer than the tag-position case, but the docstring bills this as the "single source of truth" for the tag-boundary rule. | Claude Code (single-source; verified empirically) | Correctness |
| 3 | `sdk_override/classification.py:126-129` | When `final_runtime == "other"`, delegating to `detect_service_runtime` re-reads and re-parses the same Dockerfile, re-derives the same final base, re-classifies it as `other`, then falls to `detect_service_type` — two redundant file reads plus regex passes per service. Calling `detect_service_type(svc_name, svc_config)` directly is equivalent and clearer. | Claude Code (single-source) | Efficiency / Clarity |
| 4 | `kamiwaza-ai-extensions-lib/package-lock.json:749` (and ~19 other blocks) | The diff vs. `develop` drops `libc: ["glibc"]` / `libc: ["musl"]` selectors from optional native packages (`@img/sharp-*`, `@next/swc-*`) — 60 lines of removal in a PR whose `package.json` is unchanged and whose stated scope is classification refactoring. This makes `npm ci` less reproducible across Alpine vs. glibc environments. Revert the lockfile churn, or regenerate with an npm version that preserves `libc`. | Codex (single-source) | Reproducibility / Scope |

## Test Coverage Assessment

**Adequately covered:**
- `sdk_override/classification.py` — `detect_service_runtime` precedence (base image wins / unrecognized falls back / missing Dockerfile falls back) and `detect_service_type` name→build-path→port ordering are all pinned by the new `test_sdk_override_runtime_classification.py`. `read_runtime_pythonpath`, including the restored legacy space-separated form, is covered in `test_convert_agent.py:1469`.
- `compose_transformer.py` — the `:${…}` placeholder fix is pinned for both functions (`test_publish_cmd.py:2900`, `:4021`, `:4042`), including a cross-function agreement test.
- `catalog_overlay.py` — `preview_image` passthrough is covered (`test_catalog_overlay.py:183,193`), but only for the already-normalized `images/…` shape; the shorthand `screenshot.png` case that exposes High #1 is untested.

**Gaps:**
- `sdk_override/classification.py:126-129` — the `final_runtime == "other"` fallback branch in `_detect_build_service_runtime` is new in `a3b8a57` and has no direct test. Given a previous round's regression came from exactly this fallback being dropped, it should be pinned directly.
- `sdk_override/classification.py:215-227` — `_resolve_stage_base`'s alias-chain resolution and `seen` cycle guard are untested. The recursion moved from a closure to a module-level function with explicit params; a broken guard on `FROM builder AS builder` would recurse infinitely.
- `test_sdk_override_runtime_classification.py` has no `pytestmark = pytest.mark.unit`, unlike its sibling suites. With `make test-unit` running `pytest -m "unit"`, this new file is silently excluded from that target.
- Process note: the legacy-`PYTHONPATH` regression escaped because `test_convert_agent.py` was outside the four suites listed in the PR description. Now that the parser lives in `classification.py`, that coverage would be better owned by the new classification test file.

## Positive Observations
- **Previous feedback was addressed properly, not papered over.** `_tag_separator_index` doesn't just fix `_split_image_ref` — it removes the possibility of the two functions drifting again, which is the actual defect class. That is the right shape of fix, and a test now pins the agreement invariant directly.
- **`a002a48` is good judgment.** Removing the warning rather than patching it, with a written rationale for why the gap it guarded no longer exists (the restored `detect_service_type` fallback means `detect_service_runtime` never returns `other` for a service with a `build` key).
- `ServiceRuntime = Literal[...]` replaces bare `str` returns, making the four-way contract checkable.
- `_PYTHONPATH_ENV_RE` lifted to module scope — compiled once instead of per call, and the function-local `import re` is gone.
- OSError handling preserved and centralized in one `_read_dockerfile` instead of duplicated at two call sites.
- `__all__` and the import list stayed consistent through the re-export pruning, with a test pinning the public symbol.
- Test-side change is substantive, not cosmetic: `_write_runtime_dockerfiles` gives the override tests explicit runtime evidence rather than relying on name-heuristic fallthrough.
- Commit messages state the failure mode, blast radius, and what was verified.

## Overall Verdict
**Risk Level**: Low

The refactor is behavior-preserving with two well-scoped bug fixes, every item from the previous round is genuinely fixed and verified, the full 1692-test extensions suite passes, and ruff is clean on all changed files. No Critical issues and all applicable dimensions score ≥7, so the gates pass. Linear ticket ENG-9606 is present.

Two items warrant a follow-up commit but do not block: the `preview_image` normalization asymmetry (High #1 — a one-line fix mirroring `registry_builder.py:185`) and the unrelated `package-lock.json` churn (Medium #4). The three classification findings are all pre-existing behavior faithfully carried through the refactor rather than regressions introduced by it.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

**Reviewed-SHA**: a002a4815dbe48900a5d02e0a9428170e8a2b69b

jxstanford (changes requested):
# RE-REVIEW Report: refactor(sdk-override): split runtime classification into named steps [ENG-9606]

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/245
**Branch**: `refactor/sdk-override-classification` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 8/10 | Scope held; the two "unrelated one-liners" are small and each carries a test, but the PR description no longer matches the code on two points (M7) |
| Security | 8/10 | No injection, secrets, deserialization, or access-control surface; parsing is local-file read-only. Only note is the unbounded `preview_image` forwarded to the overlay PUT without publish's escape guard (M6), dev-only |
| Test Quality | 6/10 | Strong new coverage for the tag-boundary fix and the runtime classifier, but the *other* regression fix from this same review round ships unpinned and mutation-survives (H1) |
| Error Handling | 9/10 | `_read_dockerfile` centralizes the `OSError`/missing-file handling previously duplicated in two places; every fallback path returns a defined classification rather than propagating `None` |
| Code Quality | 8/10 | Named steps are a real readability win; deductions for the stale docstring (M2), cross-module private import (M4), redundant re-parse (M5) |
| Refactoring | 8/10 | Right decomposition; the private re-export cleanup in `__init__.py` is a genuine API-surface reduction. `_contains_token` over a 1-tuple is mild ceremony, and M5's double file read was left on the table |
| Documentation Accuracy | N/A | No Markdown changed |
| Documentation Standards | N/A | No Markdown changed |

## Critical Issues (Must Fix)
| — | — | None | — | — |

## High Priority Issues
| H1 | `kamiwaza_extensions/sdk_override/classification.py:126` | The `final_runtime == "other"` fallback — the fix committed in d0091e6 in response to the previous review — has **no test**. Claude mutation-verified it: deleting those 4 lines leaves all 214 tests in `test_sdk_override.py`, `test_sdk_override_runtime_classification.py`, `test_convert_agent.py`, `test_dev_local.py` green; coverage confirms 126–128 unhit. Codex independently observed a working-tree state with exactly those 4 lines absent and diagnosed the resulting defect from source — the mutant is not hypothetical. Without the branch, a build service whose final base matches no token (`FROM debian:12`, `FROM postgres:17`) returns `"other"`, matches neither branch in `build.py:229/275`, and silently receives **no SDK overlay** — the exact regression the last round caught. The new test file exercises only `detect_service_runtime`, never `_detect_build_service_runtime` | Claude Code, Codex | test-coverage |

| M1 | `kamiwaza_extensions/catalog_overlay.py:181` | Guard is `isinstance(..., str)` where `RegistryBuilder.build_entry:185` uses truthiness. Verified: `preview_image: ""` → overlay writes `"images/"`, publish leaves `""`. A deliberate blank clear becomes a broken path — the same publish/dev divergence this commit set out to close. Use `if entry.get("preview_image"):` | Claude Code | correctness |
| M2 | `kamiwaza_extensions/compose_transformer.py:97` | `_split_image_ref`'s docstring still says "Mirrors `_replace_image_tag`'s tag detection (last `:` past the last `/`)" — that rule was replaced by `_tag_separator_index`. Stale doc on the very helper the PR unified | Claude Code | documentation |
| M3 | `kamiwaza_extensions/compose_transformer.py:36` | `_tag_separator_index` handles `foo:${TAG:-x}` but not a placeholder spanning the whole ref: `${FULL_IMAGE:-ghcr.io/org/api:v1}` retags to `${FULL_IMAGE:-ghcr.io/org/api:2.0` (unclosed brace). Not a regression — old `rfind` did the same — but the docstring now claims single-source-of-truth for the tag boundary, so the residual gap deserves a note or a test | Claude Code | edge-case |
| M4 | `kamiwaza_extensions/catalog_overlay.py:31` | Cross-module import of a private symbol (`_normalize_preview_image`). Two modules now depend on it; promote to a public helper so the coupling is declared rather than implied by a comment | Claude Code | code-quality |
| M5 | `kamiwaza_extensions/sdk_override/classification.py:126` | The `"other"` fallback calls `detect_service_runtime`, which re-reads and re-parses the same Dockerfile `_detect_build_service_runtime` already parsed at line 113. Correct, but a redundant stat+read per service | Claude Code | efficiency |
| M6 | `kamiwaza_extensions/catalog_overlay.py:181` | Unlike `publish.py:168 _resolve_preview_image`, the dev overlay has no escape guard: `../secrets/a.png` → `images/../secrets/a.png` in the PUT body. Low impact (dev-only, self-authored metadata, user's own cluster) and matches pre-existing publish-entry behavior, but worth a bounded check now that the field is forwarded | Claude Code | input-validation |
| M7 | PR description | Stale on two points: "Behavior is unchanged" — `detect_service_runtime` now routes through `_image_basename`, so `localhost:5000/python-app:3.11` classifies as `backend` (old inline `split(":")[0]` yielded `localhost` → no match). That's a genuine fix aligning it with `_detect_build_service_runtime` and deserves a line. Likewise "`catalog_overlay` passes `preview_image` through" — it normalizes now | Claude Code | documentation |
| M8 | `kamiwaza_extensions/sdk_override/classification.py:130` | `_contains_token` is an unanchored substring match, so `_NODE_IMAGE_TOKENS = ("node", "bun")` matches **`ubuntu`**. Codex verified at runtime: a service whose final stage is `FROM ubuntu:22.04` classifies as `frontend` and receives the TypeScript overlay instead of the Python one. Pre-existing (old code had the same `"bun" in final_base`), but this refactor centralizes the rule in one function — the cheapest place it will ever be to fix. Suggest matching on the basename's leading component or a boundary-aware check | Codex | correctness |

## Test Coverage Assessment
- **Adequate**: `compose_transformer.py` — `_tag_separator_index` is covered by 4 new tests (placeholder retag, placeholder split, registry-port regression, cross-helper agreement), good shape for a bug born from two helpers drifting. `catalog_overlay.py` — shorthand + already-prefixed both pinned; import-invariant tests still pass with the new `registry_builder` import (Codex confirmed at runtime that importing `catalog_overlay` pulls in none of `boto3` / `catalog_publisher` / `profile_manager`, so the ENG-6802 hard invariant holds). `classification.py` `detect_service_runtime` / `detect_service_type` — 95% statement coverage; base-image-wins, static-over-name, unrecognized-base fallback, missing-Dockerfile fallback, name/port heuristics, and `__all__` export all pinned. Alias-cycle resolution in `_resolve_stage_base` still covered via `test_sdk_override.py:626`; `read_runtime_pythonpath` covered indirectly via `test_convert_agent.py:1405-1490`.
- **Gaps**: `classification.py:126-128` — `_detect_build_service_runtime` has **no direct tests at all**. Add one with `extension_dir` + `FROM postgres:17` (or `FROM debian:12`) asserting `backend`, and ideally one for the empty-Dockerfile path. `catalog_overlay.py:181` — no test for `preview_image: ""` (M1). Neither engine found a test that would fail on M8's `ubuntu` → `frontend` misclassification.
- **Test hygiene (pre-existing, informational)**: running `tests/unit/extensions/` mutates the tracked `kamiwaza-ai-extensions-lib/package-lock.json` (−60 lines, stripping `libc` platform filters from Sharp/Next SWC optional deps). Codex saw this as uncommitted working-tree noise and flagged it as a lockfile-reproducibility concern; it is **not part of the final PR diff** — Claude verified the diff is net-clean after 91623d6 and 2a8092e. It is, however, exactly what forced those two cleanup commits. A `tmp_path` copy or a lockfile-restore fixture would stop it costing every PR two cleanup commits.

## Positive Observations
- The `_tag_separator_index` extraction is the right fix, not the expedient one: the previous round's comment could have been satisfied by patching `_replace_image_tag` alone. Instead both helpers share one rule and `test_both_helpers_agree_on_the_namespace` keeps them from drifting again. Claude probed 16 ref shapes (registry ports, placeholder registries, placeholder ports, digests, untagged, nested placeholders) and found no case where the new rule is worse than the old one; Codex confirmed the two helpers agree at runtime.
- `a002a48` reverting the speculative warning added in `a3b8a57` — `compose.py` is byte-identical to base in the final diff. Clean response to review rather than defensive accretion.
- The `ServiceRuntime` Literal turns the four-way classification into something the type checker enforces at every call site; `_classify_runtime_image` returning `"other"` rather than a bare fallback is what made the precedence explicit enough for the previous review to spot the missing branch at all.
- Test edits are honest: `test_skips_services_without_build` was duplicated at two class scopes with one silently shadowing the other — the rename to `test_skips_prebuilt_redis_service` recovers a test that had never run.
- Comments explain *why*, anchored to the incident ("they disagreed once, and the malformed ref that produced … reached the K8s PATCH path"), which is what makes the residual gap in M3 visible rather than hidden.
- The `__init__.py` private re-export cleanup is a real API-surface reduction: Codex grepped the tree and confirmed nothing outside tests imported the removed symbols, and the affected tests were updated to import from the owning module.

## Overall Verdict
**Risk Level**: Low
**Verdict**: REQUEST_CHANGES

No correctness regression was found, the full extensions suite passes locally (1694 passed, 1 skipped), and every item from the previous round is addressed in the code — the one behavior change (`_image_basename` in `detect_service_runtime`) is a fix that makes the two classifiers agree. Both engines' independent Codex-side probing corroborated the tag-boundary fix and the ENG-6802 import invariant.

The gate trips on **Test Quality 6/10**: the `"other"` fallback restored in d0091e6 survives deletion with zero failing tests, and Codex demonstrated that concretely by encountering a tree with those four lines gone. Landing a direct `_detect_build_service_runtime` test (plus the one-word `isinstance` → truthiness change in M1) clears this. M8 is worth folding in at the same time, since the refactor just made it a one-line fix in one place.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

**Reviewed-SHA**: 8cd917984636d4272492d36717221d1fdf2ca35e

jxstanford (approved):
I'll synthesize the two engine reports into a single review.

# RE-REVIEW Report: refactor(sdk-override): split runtime classification into named steps [ENG-9606]

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/245
**Branch**: `refactor/sdk-override-classification` → `develop`

## CI & Merge Status
| Check | Status |
|-------|--------|
| Mergeable | UNKNOWN |
| Merge state | UNKNOWN |
| Failing checks | None |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 9/10 | Scope matches the stated readability pass plus two named one-liners; only drift is the undocumented `_image_basename` behavior change (M2). |
| Security | 9/10 | No new attack surface; the new `registry_builder` import edge on the dev path is covered by the ENG-6802 invariant test (subprocess import-graph probe + static source check), confirmed passing. |
| Test Quality | 8/10 | New file pins classification precedence directly and the prior untested-fallback comment is fully addressed; gaps are the behavior change (M2), the whole-value placeholder (M1), and the blank `preview_image` case (M3). |
| Error Handling | 9/10 | `_read_dockerfile` now distinguishes unreadable (`None`) from empty and every caller handles both; the `seen` cycle guard in `_resolve_stage_base` survives the extraction intact. |
| Code Quality | 8/10 | Small named steps, tokens lifted to constants, regex hoisted out of the hot path; nits are the over-wide `ServiceRuntime` (M4), the private cross-module import (M5a), and the `isinstance` vs truthiness mismatch with the publish path (M3). |
| Refactoring | 8/10 | Clean, behavior-preserving extraction except M2; `_tag_separator_index` stopped one duplication short of the `${VAR:-...}` unwrap already implemented twice elsewhere (M1). |
| Documentation Accuracy | N/A | No Markdown changed in this PR. |
| Documentation Standards | N/A | No Markdown changed in this PR. |

## Critical Issues (Must Fix)
| — | — | None | — | — |

## High Priority Issues
| — | — | None | — | — |

| M1 | `kamiwaza_extensions/compose_transformer.py:49-54` | The placeholder rule only fires for `:${` *after the last slash*, so a whole-value placeholder still mis-splits: `${FULL_IMAGE:-ghcr.io/org/api:v1}` retags to `${FULL_IMAGE:-ghcr.io/org/api:2.0.0` (unbalanced brace), and `_split_image_ref` yields registry `${FULL_IMAGE:-ghcr.io`. Verified empirically. Pre-existing, not a regression — but the new docstring claims single-source-of-truth for tag boundaries, and `registry_builder._ENV_DEFAULT_SUB_RE` / `dev_env_image_refs._DEFAULT_SUB_RE` both already unwrap this shape. Reuse the `\A\$\{VAR:?-...\}\Z` unwrap or document the limit. | Claude Code | correctness |
| M2 | `kamiwaza_extensions/sdk_override/classification.py:148-156` | Behavior is *not* unchanged for port-qualified base images. Old: `base_image.split(":")[0].rsplit("/",1)[-1]` turned `localhost:5000/nginx:1.27` into `"localhost"` → no token → name/port heuristic → `backend` (Python overlay injected into an nginx runtime). New `_image_basename` → `nginx` → `static` (overlay skipped). Confirmed by running both revisions side by side. A genuine fix, but the PR body says "Behavior is unchanged" and no test pins it. Add a test and a line in the description. | Claude Code | behavior-change |
| M3 | `kamiwaza_extensions/catalog_overlay.py:181` | The `isinstance(entry.get("preview_image"), str)` guard admits the empty string, so `preview_image: ""` in `kamiwaza.json` normalizes to the literal `"images/"` and the dev overlay overwrites the catalog row with an invalid key. Confirmed empirically (`build_overlay_entry(..., metadata={"preview_image": ""})` → `images/`). The publish path (`registry_builder.py:185`) guards on truthiness instead. Match it. | Codex | correctness |
| M4 | `kamiwaza_extensions/sdk_override/classification.py:25,77,104` | `ServiceRuntime` includes `"other"`, but neither `detect_service_runtime` nor `_detect_build_service_runtime` can ever return it — both funnel through `detect_service_type`, which returns only `frontend`/`backend`. Callers get a return type wider than reality, so `if runtime == "other"` in `build.py` would type-check as live code while being dead. Consider a narrower alias for the detectors, keeping `ServiceRuntime` for `_classify_runtime_image`. | Claude Code | typing |
| M5 | `kamiwaza_extensions/catalog_overlay.py:31,181-186` | (a) The cross-module `_normalize_preview_image` import is still private-named; the ENG-6802 invariant test does cover the new `registry_builder` edge (verified passing), so the risk is convention, not safety — a public alias would read better. (b) `kz-ext dev` never uploads a preview asset (`_upload_preview_image` lives in `catalog_publisher`, forbidden on the dev path), so for a never-published extension the overlay row now carries `images/<file>` pointing at a key that was never uploaded, where previously the key was absent and the pre-shadow row's value survived. | Claude Code | design |
| M6 | `kamiwaza_extensions/sdk_override/__init__.py:35-44` | `console` was dropped from the re-exports alongside the private helpers. It isn't underscore-prefixed and the PR body describes the removal as "about twenty private helpers", so `from kamiwaza_extensions.sdk_override import console` breaks for any out-of-tree importer. Never in `__all__` and no in-repo caller uses it, so impact is likely nil — worth a line in the description. | Claude Code | api-compat |

## Test Coverage Assessment
- **Adequately covered**: `sdk_override/classification.py` (new `test_sdk_override_runtime_classification.py` pins base-image-wins, unrecognized-token fallback, missing-Dockerfile fallback, name/port heuristic, build-time `"other"` fallback including the name-heuristic variant, node→nginx multi-stage, and the `__all__` export); `compose_transformer.py` tag helpers (`TestTagSeparatorSharedByBothHelpers` pins both helpers *and* their mutual agreement — good regression shape); `catalog_overlay.py` preview_image (shorthand + already-prefixed); `read_runtime_pythonpath` indirectly via `test_convert_agent.py` (62 passed, all four ENV forms). Full run: `tests/unit/extensions` → 1701 passed, 1 skipped (skip is a BSD-sed host guard, unrelated). `ruff` clean on changed files; `mypy` on the three changed modules reports only missing `types-PyYAML` stubs.
- **Gaps**:
  - `classification.py` — the port-qualified base image path (M2); a `FROM python AS build → FROM nginx` case asserting `static` (only the node→nginx positive is pinned in the new file); `_resolve_stage_base` cycle / self-alias (`FROM a AS a`) — pre-existing, guard is correct but unpinned.
  - `catalog_overlay.py` — blank `preview_image` (M3) is unspecified in either direction.
  - `compose_transformer.py` — whole-value placeholder refs (M1), currently unspecified.
- Note the deliberate tradeoff in `test_sdk_override.py`: `_write_runtime_dockerfiles` makes those tests exercise the base-image path rather than the heuristic path. That's the right call (they were passing for the wrong reason before), and the heuristic is now covered directly in the new file.

## Positive Observations
- Every previous-round comment is addressed with a targeted commit: service-heuristic fallback restored (`d0091e6`), `"other"` branch covered (`9d0eb77`), tag-boundary helpers consolidated (`a3b8a57`), unreachable warning reverted so `compose.py` nets to zero diff (`a002a48`), package-lock churn dropped (`2a8092e`).
- The decomposition makes precedence legible: `_classify_runtime_image` returning `"other"` instead of falling off the end of a token chain is what made the "static final + node builder ⇒ frontend" rule readable, and it removed the duplicated token-matching between the two detectors.
- `_tag_separator_index` is the right consolidation — the two helpers had genuinely diverged, and `test_both_helpers_agree_on_the_namespace` pins the invariant rather than the two outputs separately, so they can't silently re-diverge.
- Docstrings carry the *why* (which colon wins and what broke when it didn't), not just the what — same for the `preview_image` comment explaining the overwrite hazard.
- Test names got sharper in the rewrite: `test_skips_services_without_build` → `test_skips_prebuilt_redis_service` describes what's actually asserted.
- Correctly resisted scope creep — `compose.py` nets to a zero diff and the lock-file churn was cleaned up rather than left in.

## Overall Verdict
**Risk Level**: Low
**Verdict**: APPROVE

Rationale: no Critical or High issues from either engine, all applicable dimensions ≥ 8, the Linear ticket (ENG-9606) is present in branch/title/commits, and the full 1701-test extensions suite plus lint/type checks pass on the changed files. The six remaining items are pre-existing edge cases, typing polish, and two small correctness nits (M2 behavior change, M3 blank `preview_image`) — none block merge, though M2 deserves a test plus a line in the PR description and M3 is a one-word guard change worth folding in before landing.

---

---
<details>

| Engine | Status |
|--------|--------|
| Claude Code | ✅ OK |
| Codex | ✅ OK |

**Engines contributing**: 2/3
</details>

**Reviewed-SHA**: 9d0eb7795b2d6f1e88e3592a537f24b2c4f12670

PR #261, opened 2026-08-13
ENG-10260: Resolve compound Compose defaults in kz-ext dev

m9e:
## Summary

- Resolve exported host values for direct `kz-ext dev` deployments using supported braced Docker Compose substitution semantics.
- Support embedded substitutions, nested defaults, `$$` escapes, literal braces in defaults, and the unset-versus-empty behavior of `:-`, `-`, `:?`, `?`, `:+`, and `+`.
- Preserve tolerated environment-list mappings and Kubernetes `name`/`value`/`valueFrom` entries through resolution and payload serialization; apply image-reference and service-reference rewriting consistently to their literal values.
- Continue dropping unresolved/unsatisfied substitutions and placeholder-bearing entries whose keys start with `KAMIWAZA_` so platform ConfigMap injection wins; excessively deep nesting now fails closed rather than raising `RecursionError`.

## Customer impact

Direct developer deployments can now carry values such as `${SECRET_ENCRYPTION_KEY}` from the invoking shell and resolve compound forms such as `neo4j/${NEO4J_PASSWORD:-changeme}`. Catalog templates remain unresolved for platform install-time substitution.

## Security note

This behavior applies only to the direct `kz-ext dev` path. Values resolved from the developer process environment are included in the submitted extension environment, matching Docker Compose behavior; callers should treat the deployed extension resource as secret-bearing configuration. Compose `$$` escapes remain literal and never read the escaped host variable. Unbraced `$VAR` remains literal by design; after `$$` collapse, a resulting `$(VAR)` token can still be expanded by Kubernetes container-env handling.

## Validation

- `make test`: 3,396 passed, 2 skipped, 260 deselected
- Changed files pass Ruff and isort
- Targeted mypy passes for `compose_transformer.py`, `dev_env_image_refs.py`, and `payload_builder.py`
- Transformer suite: 137 passed
- Focused transformer/payload/image-rewrite suites: 233 passed
- Docker Compose 5.1 runtime parity checked for escaped dollars, all supported operator branches, nested/multiple substitutions, selected nested alternates, balanced JSON defaults, asymmetric literal braces, and host-set/unset/empty values
- End-to-end regression coverage verifies `name`/`value` resolution → payload env → `service-ref-rewrites` annotation consistency
- Regression coverage verifies tolerated mapping-fragment resolution and rewrite parity, `valueFrom` preservation, and `name`/`value` image-reference rewriting
- Repository-wide `make lint` remains red on unrelated pre-existing unused-import findings

Linear: https://linear.app/kamiwaza/issue/ENG-10260

### DIFF
--- a/kamiwaza_extensions/compose_transformer.py
+++ b/kamiwaza_extensions/compose_transformer.py
@@ -297,7 +297,7 @@ def _fallback_image_basename(
 # A single Compose substitution. The fallback is greedy so nested defaults
 # are parsed after ``_compose_substitution_end`` finds the balanced expression.
 _COMPOSE_SUB_RE = re.compile(
-    r"^\$\{([A-Za-z_][A-Za-z0-9_]*)(?:(:?[-?])(.*))?\}$",
+    r"^\$\{([A-Za-z_][A-Za-z0-9_]*)(?:(:?[-?+])(.*))?\}$",
     re.DOTALL,
 )
 _COMPOSE_BRACE_RE = re.compile(r"[{}]")
@@ -342,9 +342,10 @@ _DEFAULT_LIMITS: Dict[str, Dict[str, str]] = {
 class ComposeTransformer:
     """Transform a local-dev compose dict into a deployment-ready dict.
 
-    All operations are pure (no I/O).  The caller provides the parsed
-    compose dict and receives a new dict suitable for building a
-    ``CreateExtension`` payload.
+    The caller provides the parsed compose dict and receives a new dict
+    suitable for building a ``CreateExtension`` payload. Transformations are
+    pure; ``resolve_env_placeholders`` additionally reads the process
+    environment without modifying it.
     """
 
     def transform(
@@ -487,6 +488,8 @@ class ComposeTransformer:
           substitution semantics.
         - ``${VAR:-default}`` / ``${VAR-default}`` for non-platform
           keys → recursively collapsed to a host value or literal default.
+        - ``${VAR:+alternate}`` / ``${VAR+alternate}`` → the alternate when
+          the corresponding non-empty/set condition is satisfied, else empty.
         - Placeholder-bearing entries whose key starts with ``KAMIWAZA_`` →
           dropped. The kamiwaza-extension operator injects these via ConfigMap
           envFrom; an explicit resolved entry would shadow the cluster-internal
@@ -513,12 +516,13 @@ class ComposeTransformer:
 def _resolve_shell_refs(env: Any) -> Any:
     """Resolve compose ``${VAR:-default}`` substitutions, drop unresolvable.
 
-    Two rules:
+    Three rules:
 
     1. Non-platform placeholders use the current process environment and
-       Compose unset/empty semantics for ``:-``, ``-``, ``:?``, and ``?``.
-       Embedded substitutions and nested defaults are resolved recursively.
-       The cluster deployment carries the resolved value through to the pod
+       Compose unset/empty semantics for ``:-``, ``-``, ``:?``, ``?``,
+       ``:+``, and ``+``. Embedded substitutions and nested defaults are
+       resolved recursively. The cluster deployment carries the resolved
+       value through to the pod
        — and ``detect_service_url_rewrites`` (called by
        ``PayloadBuilder``) emits a ``service-ref-rewrites`` annotation
        so the operator can swap cross-service hostnames at deploy time.
@@ -531,12 +535,12 @@ def _resolve_shell_refs(env: Any) -> Any:
 
     3. Unset bare substitutions and unsatisfied required forms are dropped.
 
-    Plain values without ``${`` pass through unchanged.
+    Values without supported substitutions or ``$$`` escapes pass unchanged.
     """
     if isinstance(env, dict):
         out: Dict[str, Any] = {}
         for k, v in env.items():
-            if not isinstance(v, str) or "${" not in v:
+            if not isinstance(v, str) or "$" not in v:
                 out[k] = v
                 continue
             resolved = _resolve_default_substitution(k, v)
@@ -557,10 +561,24 @@ def _resolve_shell_refs(env: Any) -> Any:
 
 
 def _resolve_default_substitution(key: str, value: str) -> Optional[str]:
-    """Resolve Compose substitutions in one non-platform environment value."""
-    if key.startswith(_PLATFORM_INJECTED_PREFIX):
+    """Resolve Compose substitutions in one environment value."""
+    is_platform_key = key.startswith(_PLATFORM_INJECTED_PREFIX)
+    if is_platform_key and _has_braced_substitution(value):
         return None
     return _resolve_compose_value(value)
+
+
+def _has_braced_substitution(value: str) -> bool:
+    """Whether *value* contains an unescaped ``${...}`` opener."""
+    cursor = 0
+    while (dollar := value.find("$", cursor)) >= 0:
+        if value.startswith("$$", dollar):
+            cursor = dollar + 2
+            continue
+        if value.startswith("${", dollar):
+            return True
+        cursor = dollar + 1
+    return False
 
 
 def _resolve_compose_value(value: str) -> Optional[str]:
@@ -619,6 +637,10 @@ def _resolve_compose_substitution(value: str) -> Optional[str]:
         return host_value
     if operator in ("-", "?") and is_set:
         return host_value
+    if operator == ":+":
+        return _resolve_compose_value(fallback) if host_value else ""
+    if operator == "+":
+        return _resolve_compose_value(fallback) if is_set else ""
     if operator in (":?", "?"):
         return None
     return _resolve_compose_value(fallback)
@@ -637,9 +659,9 @@ def _resolve_list_entry(entry: Any) -> Optional[str]:
 
 def _entry_has_shell_ref(entry: Any) -> bool:
     if isinstance(entry, str) and "=" in entry:
-        return "${" in entry.split("=", 1)[1]
+        return "$" in entry.split("=", 1)[1]
     if isinstance(entry, dict):
-        return "${" in str(entry.get("value", ""))
+        return "$" in str(entry.get("value", ""))
     return False
 
 
--- a/kamiwaza_extensions/compose_transformer.py
+++ b/kamiwaza_extensions/compose_transformer.py
@@ -300,6 +300,7 @@ _COMPOSE_SUB_RE = re.compile(
     r"^\$\{([A-Za-z_][A-Za-z0-9_]*)(?:(:?[-?])(.*))?\}$",
     re.DOTALL,
 )
+_COMPOSE_SUB_TOKEN_RE = re.compile(r"\$\{|}")
 
 # Env var names that should be left to the platform's ConfigMap envFrom
 # injection (operator writes the cluster-internal value; an explicit
@@ -581,17 +582,10 @@ def _resolve_compose_value(value: str) -> Optional[str]:
 def _compose_substitution_end(value: str, start: int) -> Optional[int]:
     """Find the closing brace paired with the ``${`` at *start*."""
     depth = 1
-    cursor = start + 2
-    while cursor < len(value):
-        if value.startswith("${", cursor):
-            depth += 1
-            cursor += 2
-            continue
-        if value[cursor] == "}":
-            depth -= 1
-            if depth == 0:
-                return cursor
-        cursor += 1
+    for token in _COMPOSE_SUB_TOKEN_RE.finditer(value, start + 2):
+        depth += 1 if token.group() == "${" else -1
+        if depth == 0:
+            return token.start()
     return None
 
 
--- a/kamiwaza_extensions/compose_transformer.py
+++ b/kamiwaza_extensions/compose_transformer.py
@@ -300,7 +300,7 @@ _COMPOSE_SUB_RE = re.compile(
     r"^\$\{([A-Za-z_][A-Za-z0-9_]*)(?:(:?[-?])(.*))?\}$",
     re.DOTALL,
 )
-_COMPOSE_SUB_TOKEN_RE = re.compile(r"\$\{|}")
+_COMPOSE_BRACE_RE = re.compile(r"[{}]")
 
 # Env var names that should be left to the platform's ConfigMap envFrom
 # injection (operator writes the cluster-internal value; an explicit
@@ -483,12 +483,14 @@ class ComposeTransformer:
         the pod verbatim.
 
         Rules (per env var):
-        - Host environment values are used with normal Compose semantics.
+        - Host environment values are used with Compose's supported braced
+          substitution semantics.
         - ``${VAR:-default}`` / ``${VAR-default}`` for non-platform
           keys → recursively collapsed to a host value or literal default.
-        - ``${KAMIWAZA_*:-default}`` → dropped. The kamiwaza-extension
-          operator injects these via ConfigMap envFrom; an explicit
-          env entry would shadow the cluster-internal value.
+        - Placeholder-bearing entries whose key starts with ``KAMIWAZA_`` →
+          dropped. The kamiwaza-extension operator injects these via ConfigMap
+          envFrom; an explicit resolved entry would shadow the cluster-internal
+          value.
         - Unset ``${VAR}`` and unsatisfied required forms are dropped.
         - Plain values pass through unchanged.
 
@@ -514,18 +516,18 @@ def _resolve_shell_refs(env: Any) -> Any:
     Two rules:
 
     1. Non-platform placeholders use the current process environment and
-       normal Compose unset/empty semantics. Embedded substitutions and nested
-       defaults are resolved recursively. The cluster deployment carries the
-       resolved value through to the pod
+       Compose unset/empty semantics for ``:-``, ``-``, ``:?``, and ``?``.
+       Embedded substitutions and nested defaults are resolved recursively.
+       The cluster deployment carries the resolved value through to the pod
        — and ``detect_service_url_rewrites`` (called by
        ``PayloadBuilder``) emits a ``service-ref-rewrites`` annotation
        so the operator can swap cross-service hostnames at deploy time.
 
-    2. ``${KAMIWAZA_*:-default}`` → drop. The kamiwaza-extension
-       operator injects these via ConfigMap envFrom; an explicit env
-       entry would shadow the cluster-internal value. Compose defaults
-       point to laptop-only addresses (``host.docker.internal:7777``)
-       that don't resolve in-cluster anyway.
+    2. Placeholder-bearing entries whose key starts with ``KAMIWAZA_`` → drop.
+       The kamiwaza-extension operator injects these via ConfigMap envFrom; an
+       explicit resolved entry would shadow the cluster-internal value.
+       Compose defaults point to laptop-only addresses
+       (``host.docker.internal:7777``) that don't resolve in-cluster anyway.
 
     3. Unset bare substitutions and unsatisfied required forms are dropped.
 
@@ -562,31 +564,45 @@ def _resolve_default_substitution(key: str, value: str) -> Optional[str]:
 
 
 def _resolve_compose_value(value: str) -> Optional[str]:
-    """Resolve every balanced ``${...}`` expression embedded in *value*."""
+    """Resolve braced expressions while honoring Compose's ``$$`` escape."""
     resolved: List[str] = []
     cursor = 0
-    while (start := value.find("${", cursor)) >= 0:
-        end = _compose_substitution_end(value, start)
+    while cursor < len(value):
+        dollar = value.find("$", cursor)
+        if dollar < 0:
+            resolved.append(value[cursor:])
+            break
+        resolved.append(value[cursor:dollar])
+        if value.startswith("$$", dollar):
+            resolved.append("$")
+            cursor = dollar + 2
+            continue
+        if not value.startswith("${", dollar):
+            resolved.append("$")
+            cursor = dollar + 1
+            continue
+        end = _compose_substitution_end(value, dollar)
         if end is None:
             return None
-        resolved.append(value[cursor:start])
-        substitution = _resolve_compose_substitution(value[start : end + 1])
+        substitution = _resolve_compose_substitution(value[dollar : end + 1])
         if substitution is None:
             return None
         resolved.append(substitution)
         cursor = end + 1
-    resolved.append(value[cursor:])
     return "".join(resolved)
 
 
 def _compose_substitution_end(value: str, start: int) -> Optional[int]:
     """Find the closing brace paired with the ``${`` at *start*."""
     depth = 1
-    for token in _COMPOSE_SUB_TOKEN_RE.finditer(value, start + 2):
-        depth += 1 if token.group() == "${" else -1
+    for brace in _COMPOSE_BRACE_RE.finditer(value, start + 2):
+        depth += 1 if brace.group() == "{" else -1
         if depth == 0:
-            return token.start()
-    return None
+            return brace.start()
+    # Compose's greedy fallback accepts an unmatched literal ``{`` inside a
+    # default. In that case its final ``}`` still closes the substitution.
+    final_brace = value.rfind("}", start + 2)
+    return final_brace if final_brace >= 0 else None
 
 
 def _resolve_compose_substitution(value: str) -> Optional[str]:
--- a/kamiwaza_extensions/compose_transformer.py
+++ b/kamiwaza_extensions/compose_transformer.py
@@ -496,7 +496,12 @@ class ComposeTransformer:
           envFrom; an explicit resolved entry would shadow the cluster-internal
           value.
         - Unset ``${VAR}`` and unsatisfied required forms are dropped.
+        - Unbraced ``$VAR`` references stay literal; this resolver intentionally
+          handles only braced Compose substitutions.
         - Plain values pass through unchanged.
+
+        ``$$`` escapes collapse to a literal dollar. A resulting ``$(VAR)``
+        token can still be expanded later by Kubernetes container-env handling.
 
         Skip this step when the destination DOES perform install-time
         substitution (e.g. a catalog template consumed by the platform
@@ -536,7 +541,8 @@ def _resolve_shell_refs(env: Any) -> Any:
 
     3. Unset bare substitutions and unsatisfied required forms are dropped.
 
-    Values without supported substitutions or ``$$`` escapes pass unchanged.
+    Values without supported substitutions or ``$$`` escapes pass unchanged;
+    unbraced ``$VAR`` references are outside this resolver's scope.
     """
     if isinstance(env, dict):
         out: Dict[str, Any] = {}
@@ -663,7 +669,9 @@ def _resolve_list_entry(entry: Any) -> Optional[Any]:
 
 
 def _resolve_dict_list_entry(entry: Dict[str, Any]) -> Optional[Dict[str, Any]]:
-    """Resolve a Kubernetes-style ``name``/``value`` environment entry."""
+    """Resolve a tolerated environment-list mapping."""
+    if "name" not in entry:
+        return _resolve_mapping_list_entry(entry)
     name = entry.get("name")
     value = entry.get("value")
     if name is None or not isinstance(value, str):
@@ -676,13 +684,28 @@ def _resolve_dict_list_entry(entry: Dict[str, Any]) -> Optional[Dict[str, Any]]:
     return out
 
 
+def _resolve_mapping_list_entry(entry: Dict[str, Any]) -> Optional[Dict[str, Any]]:
+    """Resolve a mapping-fragment list entry one environment key at a time."""
+    out: Dict[str, Any] = {}
+    for key, value in entry.items():
+        if not isinstance(value, str) or "$" not in value:
+            out[key] = value
+            continue
+        resolved = _resolve_env_value(str(key), value)
+        if resolved is not None:
+            out[key] = resolved
+    return out or None
+
+
 def _entry_has_shell_ref(entry: Any) -> bool:
     if isinstance(entry, str) and "=" in entry:
         return "$" in entry.split("=", 1)[1]
-    if isinstance(entry, dict) and "name" in entry:
+    if not isinstance(entry, dict):
+        return False
+    if "name" in entry:
         value = entry.get("value")
         return isinstance(value, str) and "$" in value
-    return False
+    return any(isinstance(value, str) and "$" in value for value in entry.values())
 
 
 # ------------------------------------------------------------------
--- a/kamiwaza_extensions/compose_transformer.py
+++ b/kamiwaza_extensions/compose_transformer.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import copy
+import os
 import re
 from typing import Any, Dict, List, Literal, Optional, Tuple
 
@@ -293,14 +294,12 @@ def _fallback_image_basename(
     return basename or "extension"
 
 
-# Compose ``${VAR:-default}`` (use default if unset OR empty) and the
-# ``${VAR-default}`` form (use default only if unset). For our purposes
-# both collapse to the literal default — there's no host process between
-# us and Kubernetes, so the var is always "unset" by the time the pod
-# starts. The expression may be embedded in a larger value, such as
-# ``neo4j/${NEO4J_PASSWORD:-changeme}``. ``${VAR:?error}`` and bare
-# ``${VAR}`` aren't matched (no safe default → drop downstream).
-_DEFAULT_SUB_RE = re.compile(r"\$\{([A-Za-z_][A-Za-z0-9_]*):?-([^}]*)\}")
+# A single Compose substitution. The fallback is greedy so nested defaults
+# are parsed after ``_compose_substitution_end`` finds the balanced expression.
+_COMPOSE_SUB_RE = re.compile(
+    r"^\$\{([A-Za-z_][A-Za-z0-9_]*)(?:(:?[-?])(.*))?\}$",
+    re.DOTALL,
+)
 
 # Env var names that should be left to the platform's ConfigMap envFrom
 # injection (operator writes the cluster-internal value; an explicit
@@ -483,13 +482,13 @@ class ComposeTransformer:
         the pod verbatim.
 
         Rules (per env var):
+        - Host environment values are used with normal Compose semantics.
         - ``${VAR:-default}`` / ``${VAR-default}`` for non-platform
-          keys → collapsed to the literal ``default``.
+          keys → recursively collapsed to a host value or literal default.
         - ``${KAMIWAZA_*:-default}`` → dropped. The kamiwaza-extension
           operator injects these via ConfigMap envFrom; an explicit
           env entry would shadow the cluster-internal value.
-        - ``${VAR}`` (no default) and ``${VAR:?error}`` (required) →
-          dropped. No safe value to ship.
+        - Unset ``${VAR}`` and unsatisfied required forms are dropped.
         - Plain values pass through unchanged.
 
         Skip this step when the destination DOES perform install-time
@@ -513,12 +512,10 @@ def _resolve_shell_refs(env: Any) -> Any:
 
     Two rules:
 
-    1. ``${VAR:-default}`` where the env key does NOT start with
-       ``KAMIWAZA_`` → resolve to ``default``, including when the
-       expression is embedded in a larger value. The host env isn't
-       consulted (we're nowhere near a docker-compose run); compose
-       semantics for "VAR is unset" simply use the default. The cluster
-       deployment then carries the literal default through to the pod
+    1. Non-platform placeholders use the current process environment and
+       normal Compose unset/empty semantics. Embedded substitutions and nested
+       defaults are resolved recursively. The cluster deployment carries the
+       resolved value through to the pod
        — and ``detect_service_url_rewrites`` (called by
        ``PayloadBuilder``) emits a ``service-ref-rewrites`` annotation
        so the operator can swap cross-service hostnames at deploy time.
@@ -529,8 +526,7 @@ def _resolve_shell_refs(env: Any) -> Any:
        point to laptop-only addresses (``host.docker.internal:7777``)
        that don't resolve in-cluster anyway.
 
-    3. ``${VAR}`` (no default) and ``${VAR:?error}`` (required) → drop.
-       No safe value to ship.
+    3. Unset bare substitutions and unsatisfied required forms are dropped.
 
     Plain values without ``${`` pass through unchanged.
     """
@@ -558,18 +554,64 @@ def _resolve_shell_refs(env: Any) -> Any:
 
 
 def _resolve_default_substitution(key: str, value: str) -> Optional[str]:
-    """Resolve default substitutions in a non-platform env value.
-
-    Returns None when:
-    - any substitution has no safe default
-    - the key is platform-injected (``KAMIWAZA_*``) — let envFrom win
-    """
+    """Resolve Compose substitutions in one non-platform environment value."""
     if key.startswith(_PLATFORM_INJECTED_PREFIX):
         return None
-    resolved = _DEFAULT_SUB_RE.sub(lambda match: match.group(2), value)
-    if "${" in resolved:
+    return _resolve_compose_value(value)
+
+
+def _resolve_compose_value(value: str) -> Optional[str]:
+    """Resolve every balanced ``${...}`` expression embedded in *value*."""
+    resolved: List[str] = []
+    cursor = 0
+    while (start := value.find("${", cursor)) >= 0:
+        end = _compose_substitution_end(value, start)
+        if end is None:
+            return None
+        resolved.append(value[cursor:start])
+        substitution = _resolve_compose_substitution(value[start : end + 1])
+        if substitution is None:
+            return None
+        resolved.append(substitution)
+        cursor = end + 1
+    resolved.append(value[cursor:])
+    return "".join(resolved)
+
+
+def _compose_substitution_end(value: str, start: int) -> Optional[int]:
+    """Find the closing brace paired with the ``${`` at *start*."""
+    depth = 1
+    cursor = start + 2
+    while cursor < len(value):
+        if value.startswith("${", cursor):
+            depth += 1
+            cursor += 2
+            continue
+        if value[cursor] == "}":
+            depth -= 1
+            if depth == 0:
+                return cursor
+        cursor += 1
+    return None
+
+
+def _resolve_compose_substitution(value: str) -> Optional[str]:
+    """Resolve one full substitution using Compose environment semantics."""
+    match = _COMPOSE_SUB_RE.match(value)
+    if match is None:
         return None
-    return resolved
+    name, operator, fallback = match.groups()
+    is_set = name in os.environ
+    host_value = os.environ.get(name, "")
+    if operator is None:
+        return host_value if is_set else None
+    if operator in (":-", ":?") and host_value:
+        return host_value
+    if operator in ("-", "?") and is_set:
+        return host_value
+    if operator in (":?", "?"):
+        return None
+    return _resolve_compose_value(fallback)
 
 
 def _resolve_list_entry(entry: Any) -> Optional[str]:
--- a/kamiwaza_extensions/compose_transformer.py
+++ b/kamiwaza_extensions/compose_transformer.py
@@ -777,7 +777,7 @@ def detect_service_url_rewrites(
 
 
 def _iter_env_entries(env: Any) -> List[Tuple[str, str]]:
-    """Yield ``(key, value)`` pairs from either env shape."""
+    """Yield ``(key, value)`` pairs from supported env shapes."""
     out: List[Tuple[str, str]] = []
     if isinstance(env, dict):
         for k, v in env.items():
@@ -788,8 +788,14 @@ def _iter_env_entries(env: Any) -> List[Tuple[str, str]]:
             if isinstance(entry, str) and "=" in entry:
                 k, v = entry.split("=", 1)
                 out.append((k, v))
-            elif isinstance(entry, dict) and "name" in entry and "value" in entry:
-                out.append((str(entry["name"]), str(entry["value"])))
+            elif isinstance(entry, dict):
+                if "name" in entry and "value" in entry:
+                    out.append((str(entry["name"]), str(entry["value"])))
+                    continue
+                if "name" not in entry:
+                    for k, v in entry.items():
+                        if isinstance(v, (str, int, float, bool)):
+                            out.append((str(k), str(v)))
     return out
 
 
--- a/kamiwaza_extensions/compose_transformer.py
+++ b/kamiwaza_extensions/compose_transformer.py
@@ -301,6 +301,7 @@ _COMPOSE_SUB_RE = re.compile(
     re.DOTALL,
 )
 _COMPOSE_BRACE_RE = re.compile(r"[{}]")
+_MAX_COMPOSE_SUBSTITUTION_DEPTH = 100
 
 # Env var names that should be left to the platform's ConfigMap envFrom
 # injection (operator writes the cluster-internal value; an explicit
@@ -543,7 +544,7 @@ def _resolve_shell_refs(env: Any) -> Any:
             if not isinstance(v, str) or "$" not in v:
                 out[k] = v
                 continue
-            resolved = _resolve_default_substitution(k, v)
+            resolved = _resolve_env_value(k, v)
             if resolved is not None:
                 out[k] = resolved
         return out
@@ -560,7 +561,7 @@ def _resolve_shell_refs(env: Any) -> Any:
     return env
 
 
-def _resolve_default_substitution(key: str, value: str) -> Optional[str]:
+def _resolve_env_value(key: str, value: str) -> Optional[str]:
     """Resolve Compose substitutions in one environment value."""
     is_platform_key = key.startswith(_PLATFORM_INJECTED_PREFIX)
     if is_platform_key and _has_braced_substitution(value):
@@ -581,8 +582,10 @@ def _has_braced_substitution(value: str) -> bool:
     return False
 
 
-def _resolve_compose_value(value: str) -> Optional[str]:
+def _resolve_compose_value(value: str, depth: int = 0) -> Optional[str]:
     """Resolve braced expressions while honoring Compose's ``$$`` escape."""
+    if depth >= _MAX_COMPOSE_SUBSTITUTION_DEPTH:
+        return None
     resolved: List[str] = []
     cursor = 0
     while cursor < len(value):
@@ -602,7 +605,7 @@ def _resolve_compose_value(value: str) -> Optional[str]:
         end = _compose_substitution_end(value, dollar)
         if end is None:
             return None
-        substitution = _resolve_compose_substitution(value[dollar : end + 1])
+        substitution = _resolve_compose_substitution(value[dollar : end + 1], depth)
         if substitution is None:
             return None
         resolved.append(substitution)
@@ -623,7 +626,7 @@ def _compose_substitution_end(value: str, start: int) -> Optional[int]:
     return final_brace if final_brace >= 0 else None
 
 
-def _resolve_compose_substitution(value: str) -> Optional[str]:
+def _resolve_compose_substitution(value: str, depth: int = 0) -> Optional[str]:
     """Resolve one full substitution using Compose environment semantics."""
     match = _COMPOSE_SUB_RE.match(value)
     if match is None:
@@ -638,30 +641,47 @@ def _resolve_compose_substitution(value: str) -> Optional[str]:
     if operator in ("-", "?") and is_set:
         return host_value
     if operator == ":+":
-        return _resolve_compose_value(fallback) if host_value else ""
+        return _resolve_compose_value(fallback, depth + 1) if host_value else ""
     if operator == "+":
-        return _resolve_compose_value(fallback) if is_set else ""
+        return _resolve_compose_value(fallback, depth + 1) if is_set else ""
     if operator in (":?", "?"):
         return None
-    return _resolve_compose_value(fallback)
+    return _resolve_compose_value(fallback, depth + 1)
 
 
-def _resolve_list_entry(entry: Any) -> Optional[str]:
-    """Apply ``_resolve_default_substitution`` to ``KEY=value`` list entries."""
+def _resolve_list_entry(entry: Any) -> Optional[Any]:
+    """Resolve one string or name/value dict environment-list entry."""
+    if isinstance(entry, dict):
+        return _resolve_dict_list_entry(entry)
     if not isinstance(entry, str) or "=" not in entry:
         return None
     key, value = entry.split("=", 1)
-    resolved = _resolve_default_substitution(key, value)
+    resolved = _resolve_env_value(key, value)
     if resolved is None:
         return None
     return f"{key}={resolved}"
 
 
+def _resolve_dict_list_entry(entry: Dict[str, Any]) -> Optional[Dict[str, Any]]:
+    """Resolve a Kubernetes-style ``name``/``value`` environment entry."""
+    name = entry.get("name")
+    value = entry.get("value")
+    if name is None or not isinstance(value, str):
+        return None
+    resolved = _resolve_env_value(str(name), value)
+    if resolved is None:
+        return None
+    out = dict(entry)
+    out["value"] = resolved
+    return out
+
+
 def _entry_has_shell_ref(entry: Any) -> bool:
     if isinstance(entry, str) and "=" in entry:
         return "$" in entry.split("=", 1)[1]
-    if isinstance(entry, dict):
-        return "$" in str(entry.get("value", ""))
+    if isinstance(entry, dict) and "name" in entry:
+        value = entry.get("value")
+        return isinstance(value, str) and "$" in value
     return False
 
 
--- a/kamiwaza_extensions/compose_transformer.py
+++ b/kamiwaza_extensions/compose_transformer.py
@@ -778,25 +778,37 @@ def detect_service_url_rewrites(
 
 def _iter_env_entries(env: Any) -> List[Tuple[str, str]]:
     """Yield ``(key, value)`` pairs from supported env shapes."""
-    out: List[Tuple[str, str]] = []
     if isinstance(env, dict):
-        for k, v in env.items():
-            if isinstance(v, (str, int, float, bool)):
-                out.append((str(k), str(v)))
-    elif isinstance(env, list):
-        for entry in env:
-            if isinstance(entry, str) and "=" in entry:
-                k, v = entry.split("=", 1)
-                out.append((k, v))
-            elif isinstance(entry, dict):
-                if "name" in entry and "value" in entry:
-                    out.append((str(entry["name"]), str(entry["value"])))
-                    continue
-                if "name" not in entry:
-                    for k, v in entry.items():
-                        if isinstance(v, (str, int, float, bool)):
-                            out.append((str(k), str(v)))
+        return _iter_env_mapping(env)
+    if not isinstance(env, list):
+        return []
+    out: List[Tuple[str, str]] = []
+    for entry in env:
+        out.extend(_iter_env_list_entry(entry))
     return out
+
+
+def _iter_env_mapping(env: Dict[Any, Any]) -> List[Tuple[str, str]]:
+    """Return scalar key/value pairs from an environment mapping."""
+    return [
+        (str(key), str(value))
+        for key, value in env.items()
+        if isinstance(value, (str, int, float, bool))
+    ]
+
+
+def _iter_env_list_entry(entry: Any) -> List[Tuple[str, str]]:
+    """Return key/value pairs from one supported environment-list entry."""
+    if isinstance(entry, str) and "=" in entry:
+        key, value = entry.split("=", 1)
+        return [(key, value)]
+    if not isinstance(entry, dict):
+        return []
+    if "name" in entry:
+        if "value" in entry:
+            return [(str(entry["name"]), str(entry["value"]))]
+        return []
+    return _iter_env_mapping(entry)
 
 
 def _rewrite_url_hosts(
--- a/kamiwaza_extensions/compose_transformer.py
+++ b/kamiwaza_extensions/compose_transformer.py
@@ -297,9 +297,10 @@ def _fallback_image_basename(
 # ``${VAR-default}`` form (use default only if unset). For our purposes
 # both collapse to the literal default — there's no host process between
 # us and Kubernetes, so the var is always "unset" by the time the pod
-# starts. ``${VAR:?error}`` and bare ``${VAR}`` aren't matched (no safe
-# default → drop downstream).
-_DEFAULT_SUB_RE = re.compile(r"^\$\{([A-Za-z_][A-Za-z0-9_]*):?-([^}]*)\}$")
+# starts. The expression may be embedded in a larger value, such as
+# ``neo4j/${NEO4J_PASSWORD:-changeme}``. ``${VAR:?error}`` and bare
+# ``${VAR}`` aren't matched (no safe default → drop downstream).
+_DEFAULT_SUB_RE = re.compile(r"\$\{([A-Za-z_][A-Za-z0-9_]*):?-([^}]*)\}")
 
 # Env var names that should be left to the platform's ConfigMap envFrom
 # injection (operator writes the cluster-internal value; an explicit
@@ -512,8 +513,9 @@ def _resolve_shell_refs(env: Any) -> Any:
 
     Two rules:
 
-    1. ``${VAR:-default}`` where ``VAR`` does NOT start with
-       ``KAMIWAZA_`` → resolve to ``default``. The host env isn't
+    1. ``${VAR:-default}`` where the env key does NOT start with
+       ``KAMIWAZA_`` → resolve to ``default``, including when the
+       expression is embedded in a larger value. The host env isn't
        consulted (we're nowhere near a docker-compose run); compose
        semantics for "VAR is unset" simply use the default. The cluster
        deployment then carries the literal default through to the pod
@@ -556,18 +558,18 @@ def _resolve_shell_refs(env: Any) -> Any:
 
 
 def _resolve_default_substitution(key: str, value: str) -> Optional[str]:
-    """Return ``default`` from ``${VAR:-default}`` for non-platform keys.
+    """Resolve default substitutions in a non-platform env value.
 
     Returns None when:
-    - the value isn't a single ``${VAR(:-)default}`` substitution
+    - any substitution has no safe default
     - the key is platform-injected (``KAMIWAZA_*``) — let envFrom win
     """
     if key.startswith(_PLATFORM_INJECTED_PREFIX):
         return None
-    m = _DEFAULT_SUB_RE.match(value.strip())
-    if not m:
+    resolved = _DEFAULT_SUB_RE.sub(lambda match: match.group(2), value)
+    if "${" in resolved:
         return None
-    return m.group(2)
+    return resolved
 
 
 def _resolve_list_entry(entry: Any) -> Optional[str]:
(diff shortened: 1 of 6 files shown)


m9e:
# PR Review Report: ENG-10260: Resolve compound Compose defaults in kz-ext dev

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/261
**Branch**: `fix/dev-compound-env-defaults` -> `develop`
**Scope**: 2 files, +44/-10 · reviewed at head `bcce3ba`, merge-base `8d21d43`
**Engines**: 5 launched (Superpowers, Claude, Codex, Kamiwaza, Antigravity)

The core fix is correct and genuinely tested. Both new tests kill mutants, and one is a **sole killer** — better than most single-scenario test additions. CI is 9/9 green, and the change is additive across the unit suite (3,143 -> 3,145 passing, with one pre-existing environmental failure that reproduces unchanged at the merge-base).

The concern is scope, not direction. Un-anchoring `_DEFAULT_SUB_RE` converts a set of **fail-safe drops into materialized values**, and it feeds those values into downstream paths with known blind spots. Every item below is backed by executed merge-base-vs-HEAD output, not inspection.

---

## High Priority

| H1 | `compose_transformer.py:519-523,605` | **The docstring's cited safety net provably does not cover the PR's named customer.** The amended rule-1 docstring justifies resolving-to-default by pointing at `detect_service_url_rewrites`, which "emits a `service-ref-rewrites` annotation so the operator can swap cross-service hostnames at deploy time." But `_URL_HOST_RE` is `(?P<scheme>https?://)…` — **HTTP/HTTPS only**. Executed on a Graphiti-shaped compose: `NEO4J_URI: "bolt://${NEO4J_HOST:-neo4j}:7687"` was `DROPPED` at merge-base and is now materialized as `bolt://neo4j:7687`, and `detect_service_url_rewrites` returns a rewrite for the sibling `http://` value but **nothing** for `NEO4J_URI`. The operator prefixes service names (`myapp-dev-abc-neo4j`), so the bare `neo4j` alias will not resolve in-cluster. Net effect for the exact Graphiti/Neo4j scenario this PR exists to fix: merge-base shipped no `NEO4J_URI` (app falls back to its own default); HEAD ships a **present-but-unresolvable** one. The gap pre-exists for plain values, but this PR newly routes substitution-bearing values into it. Fix: narrow the docstring's safety claim, or widen `_URL_HOST_RE` to `bolt`/`redis`/`postgres`/`amqp`/`mongodb`. | SP, CL, primary |
| H2 | `compose_transformer.py:303,569` | **Compose's `$$` escape is now silently corrupted rather than dropped.** In Compose, `$$` escapes a literal `$`, so `$${VAR:-d}` means the literal string `${VAR:-d}`. The anchor previously made this unmatchable; the unanchored regex now matches at offset 1. Executed: merge-base `DROPPED` -> HEAD `'$d'`. Neither ref is *correct*, but HEAD ships a corrupted value into the pod where the base shipped nothing. The converse also regresses in the other direction: `"$${LIT} ${A:-1}"` is dropped at both refs — a legitimate value lost. No repo compose uses `$$` under `environment:` today (verified by grep over all `*.yml`/`*.yaml`), so this is **latent, not live** — but any user compose with a `$$`-escaped literal is now mangled. Zero test coverage either direction. | SP, CL, KZ, AG, primary |

## Medium Priority

| M1 | `compose_transformer.py:567-572` | **The `KAMIWAZA_` platform-injection guard is keyed on the env *key*, never on the referenced *variable name*** — so this PR widens the leak from "value is exactly one substitution" to "any substitution embedded anywhere." Executed: `NEO4J_URI: "http://${KAMIWAZA_HOST:-localhost}:7687"` -> merge-base `DROPPED`, HEAD `'http://localhost:7687'`; this is what the comment at :305-311 forbids. **Severity held at Medium because the merge-base control shows the class is pre-existing**: the repo's own primary fixture (`test_compose_transformer.py:34`, `OPENAI_BASE_URL: ${KAMIWAZA_ENDPOINT:-http://host.docker.internal:8080}`) already materialized `host.docker.internal` at 8d21d43. The PR extends an existing hole rather than opening a new one. A ~4-line var-name-aware fix (return `match.group(0)` from the `sub` callback when `match.group(1).startswith(_PLATFORM_INJECTED_PREFIX)`, letting the existing guard drop the entry) was applied and **passed all 1,777 `tests/unit/extensions` tests unchanged** — confirming this is an unguarded gap, not a pinned contract. | SP, CL, KZ, primary |
| M2 | `compose_transformer.py:569` | **Defaults containing `}` are now mangled instead of dropped.** `([^}]*)` stops at the first `}`, and trailing braces are passed through verbatim, so the value re-balances into something plausible-but-wrong. Executed: `${CFG:-{"a":1},"b":2}` -> merge-base `DROPPED`, HEAD `'{"a":1,"b":2}'`. (Note: the simpler `${CONFIG:-{"key":"val"}}` resolves *correctly* — see Refuted below.) | KZ, primary |
| M3 | `compose_transformer.py:570` | **The `"${" in resolved` sentinel only catches brace-form references, so a bare `$VAR` alongside a resolvable substitution leaks literally.** Executed: `$DB_PASS/${A:-1}` -> merge-base `DROPPED`, HEAD `'$DB_PASS/1'` — an unexpanded shell reference shipped into the pod env. (A bare `$VAR` *alone* is unaffected — see Refuted.) | primary, KZ |
| M4 | `tests/…/test_compose_transformer.py:499-529` | **The new tests cover the two spellings named in the PR body, not the class of change.** The **list path is entirely untested for the new behavior** — `test_resolves_default_substitution_list_form` still uses the anchored form. Executed new-but-unpinned behavior: `"KEY=neo4j/${PW:-changeme}"` -> merge-base `DROPPED`, HEAD `'KEY=neo4j/changeme'`; and **multiple substitutions in one value — the PR's headline capability** — `"${A:-1}/${B:-2}"` -> merge-base `DROPPED`, HEAD `'1/2'`. Also unpinned: `$$` escaping, whitespace preservation, `KAMIWAZA_*` embedded under a non-platform key, and the fail-closed nested case. | CL, SP, AG, KZ |
| M5 | `compose_transformer.py:488,526` | **Two docstrings now contradict the code and each other.** Rule 1 was correctly amended to say "the env **key**", but rule 2 (:526) and the public `resolve_env_placeholders` docstring (:488) still assert `${KAMIWAZA_*:-default}` -> dropped. Executed: that is false at **both** refs. The docs were already wrong; this PR widens the gap they misdescribe without correcting them. | SP, CL |

## Low Priority

| L1 | `compose_transformer.py:569` | `value.strip()` was silently dropped: `"  ${A:-1}  "` -> merge-base `'1'`, HEAD `'  1  '`. HEAD is arguably more compose-faithful, but the change is undocumented and untested. | SP, CX, KZ, CL, primary |
| L2 | `compose_transformer.py:303` | Three near-identical regexes now diverge in anchoring, **two sharing the name `_DEFAULT_SUB_RE`**: this one (unanchored, `[^}]*`) vs `dev_env_image_refs.py:47` (anchored, `[^}]+`) vs `registry_builder.py:32` (anchored). A same-named constant with opposite semantics in a sibling module is a footgun; consider `_EMBEDDED_DEFAULT_SUB_RE`. | SP, CL, primary |
| L3 | `payload_builder.py:269-291` | `_parse_env` emits env as a plain `env: [{name, value}]` list — no Secret. So the headline win materializes a credential literal (`NEO4J_AUTH: neo4j/changeme`) into the plaintext Deployment spec, readable by anyone with `describe pod` in the namespace. Intended for a dev flow and a functional improvement, but worth an explicit docstring acknowledgement. | CL, SP |

## Validated Remedy

Two small edits restore the fail-safe drops for H2 and M3 while preserving **every** behavior the PR's own and pre-existing tests require (verified against `NEO4J_AUTH`, `BACKEND_URL`, `OPENAI_BASE_URL`, multi-substitution, empty-default, and the `:?required` drop):

```python
_DEFAULT_SUB_RE = re.compile(r"(?<!\$)\$\{([A-Za-z_][A-Za-z0-9_]*):?-([^}]*)\}")
_RESIDUAL_REF_RE = re.compile(r"\$(?:\{|[A-Za-z_])")
...
    resolved = _DEFAULT_SUB_RE.sub(lambda m: m.group(2), value)
    if _RESIDUAL_REF_RE.search(resolved):
        return None
```

Executed result: `$${VAR:-x}` and `$DB_PASS/${A:-1}` return to `DROP`; `neo4j/${NEO4J_PASSWORD:-changeme}` -> `neo4j/changeme` and `http://${HOST:-localhost}:${PORT:-7474}` -> `http://localhost:7474` are unchanged. This does **not** address M2 (`}`-in-default), which needs a balanced-brace or trailing-`}` check.

## Refuted — Do Not Chase These

Several engine findings did not survive execution. Listing them so they aren't actioned:

- **"De-anchoring changes every other call site of `_DEFAULT_SUB_RE`"** (KZ, High) — **False.** The constant has exactly two references in `compose_transformer.py` (definition :303, use :569) and is not imported cross-module. The sibling regexes are independent constants, correctly left anchored because they *preserve* the substitution form while rewriting the default. The blast radius does not materialize.
- **"Bare `$VAR` values are now returned literally instead of dropped"** (KZ, High) — **Unreachable.** Both callers pre-filter on `"${" in v` (:540) / `_entry_has_shell_ref` (:552), so `"$DB_PASSWORD"` never reaches the function; it passes through byte-identically at both refs. (The *compound* variant is real — that is M3.)
- **"Values containing a literal `${` are newly dropped, breaking Helm/Kustomize pass-through"** (KZ, **Critical**) — **False.** `"format ${var}"` is dropped at merge-base *and* HEAD. Unchanged behavior.
- **"Nested `${A:-${B:-c}}` is a regression"** (KZ, High) — **False.** Dropped at both refs. A pre-existing limitation that correctly fails closed, worth a test but not a defect.
- **"`${CONFIG:-{"key":"val"}}` is truncated/corrupted"** (AG, High) — **False.** It resolves to `{"key":"val"}` correctly, as does `${CONFIG:-{"a":{"b":1}}}`; unmatched trailing braces reassemble. Only the asymmetric form in M2 actually mangles.

## Engine Reliability Notes

- **Kamiwaza lane did not run its configured model.** `kamiwaza/tokenator/GLM-5.2-NVFP4` is absent from `/v1/models`; the tool **failed closed** rather than silently substituting (correct behavior). Two substitute models were run explicitly — one of which is Claude-family, so it is weak evidence of engine independence. **This is not a Kamiwaza clean bill of health**, and 4 of its findings are refuted above.
- **Codex** required the known `--base` + positional-prompt workaround (`CODEX_EXIT=2`, 0-byte stdout, 173-byte stderr — a usage rejection, never "no findings"). It was re-run correctly with an isolated `CODEX_HOME`/`HOME` and verified base scoping before reporting.

## Test Coverage Assessment

**Coverage verdict**: Adequate for the two spellings shipped; **insufficient for the class of change.**

Mutation results (run against `test_compose_transformer.py` at HEAD, production file restored byte-identical after each):

| Mutant | Result | Killed by |
|---|---|---|
| M1: full revert to the anchored regex + `match`/`group(2)` | 1 failed, 112 passed | `test_resolves_default_substitution_embedded_in_value` |
| M2: delete the `if "${" in resolved` guard | 2 failed, 111 passed | new test **and** a pre-existing one |
| M3: weaken guard to `resolved.strip().startswith("${")` | 1 failed, 112 passed | `test_drops_embedded_substitution_without_safe_default` — **sole killer** |

Both new tests carry real weight; M3 shows the second is the only thing in the entire suite pinning the guard against an embedded-position weakening.

**Missing tests**: list-form embedded substitution; multiple substitutions in one value; `$$` escape (both directions); `KAMIWAZA_*` var under a non-platform key; whitespace preservation; nested fail-closed lock; an integration assertion through `detect_service_url_rewrites` for a non-HTTP cross-service URL.

## Reviewer Verdicts

| Superpowers | Request Changes |
| Claude | Request Changes |
| Codex | Approve (explicit clean bill; independently confirmed the pre-filter argument) |
| Kamiwaza | Comment (configured model unavailable; substitutes only; 4 findings refuted) |
| Antigravity | Comment (Risk: Low, but two High findings — one refuted) |

**Vote Tally**: Approve: 1 | Comment: 2 | Request Changes: 2 (out of 5)

## Positive Observations

- **`sub` with a *callable* replacement is the right call** — a string template would have turned backslashes and `\1` in defaults into an escape-injection bug.
- **The `"${" in resolved` post-condition is genuinely clever fail-closed design** — it composes across *all* substitutions, so one unresolvable reference still drops the whole entry (`neo4j/${VAR:?err}` -> dropped). Mutation M3 confirms it is load-bearing.
- **The sibling anchored regexes were correctly left alone.** The author did not over-apply the change; anchoring is right in `registry_builder`/`dev_env_image_refs` because they rewrite refs while preserving the substitution form.
- **The PR fixed a real pre-existing docstring inaccuracy** at :516 (the `KAMIWAZA_` check has always been on the env key, not `VAR`).
- **No ReDoS** — no nested quantifiers; measured 1-5 µs on 5 KB pathological inputs.
- **Useful side effect**: compound `AGENT_SERVER_IMAGE` / `SANDBOX_ALLOWED_IMAGE_PREFIXES` values previously dropped before `rewrite_env_image_refs` now survive to be rewritten, strengthening the ENG-7110 lockstep invariant.
- **Validation claims check out**, with one note: the changed test file goes 111 -> 113, and `tests/unit` 3,143 -> 3,145 with the single failure (`test_push_call_uses_effective_insecure_not_persisted_flag`) reproducing identically at the merge-base — host-dependent Docker `insecure-registries`, not attributable to this PR. The body's "2,485 passed / 1,143 deselected" could not be reproduced (this host collects 3,146 in `tests/unit`) — almost certainly a different marker selection, not a defect. `make lint` targets `kamiwaza_sdk/`, not the changed `kamiwaza_extensions/`, so the "28 unrelated pre-existing findings" caveat is accurate and consistent with the green CI Ruff job.

## Overall Verdict

**Risk Level**: Medium
**Reviewer Consensus**: 2/5 request changes; escalated by rule because High findings survive verification

The direction is right and the fix should land. What blocks it is that un-anchoring trades several **fail-safe drops for silently-wrong materialized values**, and the two paths that were supposed to make materialization safe — the `KAMIWAZA_` guard and `detect_service_url_rewrites` — are both blind in exactly the case the PR names (H1). H2 and M3 are closed by the four-line remedy above; H1 needs either a narrowed docstring claim or a widened scheme regex. The list-form and multi-substitution tests (M4) matter most, because both are real new behavior with nothing pinning them.

---

*Posted as a Comment rather than a formal Request Changes: the reviewing account (`m9e`) authored this PR, and GitHub does not permit approving or requesting changes on your own pull request.*

Legend: SP=Superpowers, CL=Claude, CX=Codex, KZ=Kamiwaza, AG=Antigravity, primary=synthesizing reviewer

m9e:
Follow-up at `4eaf6cd` after revalidating the earlier review against the current head and the supported operator:

- **H2 (`$$` escape): fixed.** The resolver now consumes `$$` before looking for `${...}`, so an escaped host variable remains literal and cannot be materialized into the deployment. Set/unset and mixed escaped/resolved cases are covered.
- **Brace/default corruption: fixed.** The balanced scanner now matches Docker Compose when a fallback contains JSON or asymmetric literal braces, including when the host value wins. Both branches are covered.
- **H1 (non-HTTP sibling URLs): refuted by the deployment integration.** The SDK annotation detector is HTTP-specific, but the operator always follows exact annotation rewrites with its generic `rewriteServiceEnvVars` pass. That pass handles arbitrary `scheme://` URLs, userinfo DSNs, and host:port values; operator tests cover Redis and PostgreSQL DSNs, and the path is present in supported v0.13.0-v1.0.x releases. No SDK expansion is needed.
- **`KAMIWAZA_*` policy/docs: clarified.** The guard is intentionally keyed on the destination env key to prevent a resolved explicit entry from shadowing operator ConfigMap injection. Referencing a platform-named host variable under another key is an explicit author alias, not a shadowing path. The docstrings now describe the actual key-based, placeholder-bearing rule.
- **Coverage gaps: addressed for the changed behavior.** Added embedded list-form/multiple-substitution, escaped-dollar, balanced/unmatched-brace, and ambient-environment isolation coverage. Unbraced `$VAR`, `+`/`:+`, project `.env`, and malformed-input error parity are outside this PR's stated braced-operator/exported-shell scope; the docs no longer claim full Compose interpolation support.

Validation: `make test` reports 3,380 passed, 2 skipped, 260 deselected; changed-file Ruff/isort and targeted mypy pass. An independent post-fix review compared more than 20 parser/termination cases with the actual Docker Compose 5.1 container environment and found no blocker.

m9e:
# PR Review Report: ENG-10260: Resolve compound Compose defaults in kz-ext dev

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/261
**Branch**: fix/dev-compound-env-defaults → develop
**Scope**: 2 files, +296/−47
**Process**: Four-model parallel review (Claude deep analysis w/ empirical Docker Compose 5.1.0 parity testing, Codex `gpt-5.5`, Kamiwaza DeepSeek-V4-Flash-0731 via kzinfer, Antigravity), findings cross-verified against the diff before synthesis.

---

## Critical Issues (Must Fix)

None. All four reviewers found the core scanner correct; the Claude lane verified parity with Docker Compose 5.1.0 (`docker compose config`) on nested defaults, balanced/unmatched braces, `$$` escapes, the full `:-`/`-`/`:?`/`?` set/empty/unset matrix, and multi-substitution values.

## High Priority Issues (Should Fix — 2+ reviewers or verified single-source)

| 1 | compose_transformer.py:299 | `:+`/`+` (Compose "alternate value") operators aren't matched by `_COMPOSE_SUB_RE`, so `${VAR:+x}` entries are **silently dropped even when resolvable** (real Compose yields `x` when set, `""` when unset — empirically verified). Support them (~2 lines) or document the exclusion. Pre-existing behavior, but now contradicts the PR's stated goal of host-resolution fidelity. | CL, KZ, AG | Correctness |
| 2 | compose_transformer.py:539, 638 | `$$` unescaping is gated on `${` appearing in the value: `ab$$cd` ships literally as `ab$$cd` (Compose: `ab$cd`) while `pa$$word ${X:-y}` correctly ships `pa$word y`. Same token, different meaning depending on unrelated content. Gate on `"$" in v` or document. Empirically verified. | CL | Correctness/Consistency |
| 3 | compose_transformer.py:622 | Unsatisfied `:?`/`?` silently drops the entry; real Compose hard-fails with the author's error message. A developer's `${DB_PASSWORD:?required}` typo becomes an opaque runtime failure in the pod. Log a warning naming the variable and error text at the `kz-ext dev` call site. | CL, AG | Debuggability |
| 4 | compose_transformer.py:640 / dev.py:787 | Host-env capture is silent: any exported shell var matching a placeholder name (`OPENAI_API_KEY`, `USER`, `PORT`…) is baked into the cluster-visible deploy payload. Intended per PR body, but an informational log listing captured variable **names** (never values) would let developers audit what left their machine. | CL, AG | Security/Auditability |

| 5 | compose_transformer.py:345 | Class docstring still says "All operations are pure (no I/O)" — now false; `resolve_env_placeholders` reads `os.environ`. | CL | Documentation |
| 6 | compose_transformer.py:628 | Greedy `rfind("}")` fallback can pair `${` with a **later** substitution's closing brace when an unmatched `{` in a default precedes another `${…}` (`${X:-{a} ${Y:-b}`). Real Compose rejects such input. Obscure; tests cover only the single-substitution case. | CL, AG | Edge case |
| 7 | compose_transformer.py:557 | KAMIWAZA drop is key-based only: `BACKEND_URL=${KAMIWAZA_X:-laptop-default}` still resolves from host env or default, potentially shadowing ConfigMap intent. Pre-existing design, but host-env consultation amplifies it. Worth pinning intent with a test. | KZ | Design |
| 8 | compose_transformer.py:580 | Mixed braced+unbraced values (`${A:-x}-$B`) ship half-resolved (`x-$B` literal) where Compose resolves both. Deliberate "braced-only" scope — document it in the `_resolve_shell_refs` docstring. | CL | Documentation |
| 9 | compose_transformer.py:559 | `_resolve_default_substitution` name is stale, and reading `os.environ` directly forces `monkeypatch` gymnastics in every test — an injectable `environ: Mapping` parameter would restore determinism. | CL | Design/Testability |
| 10 | compose_transformer.py:559 | `.strip()` removal: `" ${V:-d} "` now yields `" d "` instead of `d`. More Compose-correct, but an unmentioned behavior change. | KZ | Behavior note |

**Refuted finding**: one reviewer flagged "unbounded recursion / cycle on `A=${B:-${A:-x}}`" as Critical — this is incorrect. Host env values are returned verbatim and never re-resolved, so recursion depth is bounded by literal brace nesting in the compose file itself. (A pathological multi-thousand-deep hand-written literal could hit Python's recursion limit; negligible.)

## Test Coverage Assessment

- **Coverage verdict**: Adequate — 12 new/updated tests including a parametrized empty-vs-unset matrix for all four supported operators, plus escapes, nested defaults, literal/unmatched braces, and list-form multi-substitution. The `KZ_TEST_*` rename + explicit `delenv` retrofit was essential (old names like `OPENAI_API_KEY` would have flaked once host env is consulted).
- **Missing test cases**: `${VAR:+x}` (pin current drop), `$$` without `${` (`ab$$cd`), mixed resolvable+unresolvable value, trailing bare `$`, unclosed `${VAR`, empty default `${VAR:-}`, non-KAMIWAZA key referencing `${KAMIWAZA_X:-default}`.

## Reviewer Verdicts

| Claude | Comment |
| Codex (gpt-5.5) | Approve |
| Kamiwaza (kzinfer fallback — no sandbox/tools; DeepSeek-V4-Flash-0731) | Request Changes (rested on refuted critical) |
| Antigravity | Approve |

**Vote Tally**: Approve: 2 | Comment: 1 | Request Changes: 1 (out of 4 successful reviewers)

- The balanced-brace scanner is empirically faithful to real Compose 5.1.0 on every hard case tested — including the pathological `${X:-{foo} ${Y:-bar}}` — in ~60 lines.
- Empty-vs-unset semantics for all four operators are exactly right and pinned by a tight parametrized matrix.
- `$$` handling is inline (no placeholder round-trip) and never leaks the escaped variable's host value.
- Test-hygiene migration to `KZ_TEST_*` names prevented a whole class of developer-machine flakes.
- All new functions sit comfortably under the repo's CodeScene caps; the catalog-vs-K8s split (catalog keeps live placeholders) is preserved correctly.

## Overall Verdict

**Risk Level**: Low
**Recommendation**: Comment
**Reviewer Consensus**: 3/4 agree the PR is mergeable with non-blocking follow-ups (everything fails "safe" — dropped, not wrong-valued); the lone Request Changes rested on a refuted finding.

---

Legend: CL=Claude, CX=Codex, KZ=Kamiwaza, AG=Antigravity (agy)

m9e:
Follow-up at `ba32e6b` on the fresh four-lane review:

- **Alternate-value operators: fixed.** `${VAR:+alternate}` and `${VAR+alternate}` now follow Docker Compose's non-empty/set truth table, including nested alternates and escaped dollars.
- **Plain `$$` consistency: fixed.** Escapes now collapse in every string/list environment value, not only values that also contain `${...}`. Escaped braced expressions still remain literal and never consult the host variable.
- **Purity documentation: fixed.** The transformer docs now explicitly identify `resolve_env_placeholders` as the operation that reads (but does not modify) the process environment.
- **Required-form failure behavior: intentional.** Direct deploy continues its documented fail-closed policy by dropping unresolved or unsatisfied entries rather than aborting the whole deployment. Changing that contract to a Compose-style process error is outside this focused compatibility fix.
- **Host capture visibility: intentional and disclosed.** Reading exported host values is the feature requested by ENG-10260, and the PR security note explicitly says those resolved values enter the submitted secret-bearing extension resource. Logging captured names would be an observability enhancement, not a correctness blocker.

Validation at this head: `make test` reports 3,384 passed, 2 skipped, 260 deselected; the transformer suite reports 129 passed; changed-file Ruff/isort and targeted mypy pass. A final independent audit compared all six set/empty/unset branches plus `$$` behavior with an actual Docker Compose 5.1 container and found no blocking defect.

m9e:
# PR Review Report: ENG-10260: Resolve compound Compose defaults in kz-ext dev

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/261
**Branch**: fix/dev-compound-env-defaults -> develop
**Scope**: 2 files, +348/-56

Four-model review: Claude (CL), Codex (CX), Kamiwaza DeepSeek-V4-Flash via Kbox/Pi (KZ), Antigravity (AG). All four lanes completed successfully. Claude's lane ran empirical parity checks against a real Docker Compose 5.1 (`docker compose config`); the orchestrator independently verified the top finding against the merge-base source.

---

None — no finding carries all-reviewer critical consensus.

| 1 | compose_transformer.py:663-664 (with 649-657) | **Data-loss regression for dict-form env entries.** `_entry_has_shell_ref` was widened from `"${"` to `"$"`, but `_resolve_list_entry` still returns `None` for any non-string entry — so `{"name": "FOO", "value": "cost$value"}` (or any dict value containing a bare `$` or `$$` escape) is now silently dropped, where pre-PR it passed through unchanged. Verified three ways: KZ by execution, CL empirically (`_resolve_shell_refs([{"name": "PRICE", "value": "cost$$value"}, "PLAIN=ok"])` → `['PLAIN=ok']`), and orchestrator by merge-base source diff. Dict-form entries are a supported input (`_iter_env_entries`, `payload_builder._parse_env`). No test covers them. Fix: resolve `entry["value"]` for dict entries in `_resolve_list_entry`, or gate the dict branch of `_entry_has_shell_ref` on `_has_braced_substitution`; add a regression test. | KZ, CL | Correctness / Regression |
| 2 | compose_transformer.py:598-600 | **Unbraced `$VAR` parity gap.** Unbraced references are neither resolved nor dropped — literal `$VAR` ships to the pod. Verified vs Docker Compose 5.1: `$KZPROBE/data` interpolates there; `${A:-$B}` resolves the inner `$B`. Pre-existing (not a regression) and explicitly out of the PR's stated "braced" scope, but it silently produces wrong deployed values for compose files that work locally — most surprising inside defaults where the braced machinery otherwise resolves. Either resolve named refs, drop entries containing them per the module's no-unsafe-value policy, or ticket as explicit follow-up. | CL (single-source, confidence 85, empirically verified) | Compose parity |

| 3 | compose_transformer.py:602-604, 644-646; commands/dev.py:787 | **Silent drops mask configuration errors.** Malformed interpolation (`${A:-noclose`) and unsatisfied `:?`/`?` silently drop the whole entry — discarding the author's `:?error` message — where Docker Compose 5.1 hard-errors with a descriptive message (CL verified). And since results now also vary with the developer's host env (new in this PR), "why is my var missing in the pod" gets harder to diagnose. Have the dev.py call site report which keys were dropped and why; optionally also print which host variables were consumed. | KZ, CL, AG | Ergonomics / debuggability |
| 4 | compose_transformer.py:641-646 | **Unbounded recursion on nested defaults.** A ~2000-deep `${A:-${A:-…}}` raises an unhandled `RecursionError`, crashing `kz-ext dev` with a traceback (CL verified). Input is the developer's own file, so self-inflicted; an iterative loop or small depth cap with a clean error would be more robust. (AG's related O(N²) claim is disputed — CL found the scanner linear.) | CL, AG (related) | Robustness |
| 5 | compose_transformer.py:641-643 | **`:+`/`+` alternate is recursively resolved** via `_resolve_compose_value`; an alternate containing an unresolvable `${...}` drops the whole value, where Compose emits the alternate without re-scanning. Add a confirming test or document the divergence. | KZ (single-source) | Compose parity |
| 6 | compose_transformer.py:563-568 | **`KAMIWAZA_` guard keys on the env key, not the referenced variable**: `FOO: ${KAMIWAZA_API_KEY}` reads and ships the developer's exported value (CL verified). Key-based is right for the ConfigMap-shadowing rule and matches Compose; worth a line of documentation on the capture surface. | CL (single-source) | Documentation |
| 7 | compose_transformer.py:563; test_compose_transformer.py:577-579 | **Naming/doc nits**: `_resolve_default_substitution` now handles all six operators, escapes, and embedded substitutions — name and docstring understate it (e.g. `_resolve_env_value`); `test_alternate_default_form_dash_only` docstring is stale. | CL (single-source) | Documentation |

## Adjudicated / Rejected Findings

- **AG (Critical): host-env exposure — OVERRULED by reviewer-of-record ruling.** The `kz-ext dev` path is non-prod developer tooling; reading the invoking shell's environment there is the PR's stated, documented feature and matches the `docker compose up` trust model. Not classified as a security issue. (For the record: the behavior IS new in this PR — the merge-base never read `os.environ` — and AG rated it Critical, KZ/CL suggestion-level.) The residual non-security kernel (host-env-dependent results complicate debugging) is folded into finding #3.
- **AG (High): "greedy unmatched-brace fallback diverges from Compose"** — overruled by empirical evidence. CL verified asymmetric literal braces (`${A:-{x}y}` → `{x}y`) match Docker Compose 5.1 exactly, and KZ independently verified the brace balancing as correct Compose behavior. The legitimate kernel (Compose errors on truly malformed interpolation where this code silently drops) is folded into finding #3.
- **AG (High): "O(N²) quadratic execution"** — disputed by CL (linear scanner, no quadratic scans); the real, verified pathological-input issue is recursion depth (#4).

## Test Coverage Assessment

- **Coverage verdict**: Adequate overall — unusually strong for the new resolution paths (parametrized empty-vs-unset matrix for all six operators, `$$` escapes, nested defaults, JSON braces, embedded/multiple substitutions; 129/129 pass, plus 141 in adjacent suites). Needs work on the specific gaps below.
- **Missing test cases**:
  - Dict-form list entry (`{"name": ..., "value": "cost$$value"}`) — would have caught finding #1
  - Unbraced `$VAR` passthrough — pin intended behavior (finding #2)
  - Malformed interpolation (`${A:-noclose`) → entry dropped, value not mangled
  - `${VAR:?err}` with var set-but-empty (only unset is covered)
  - Nested default with unresolvable inner substitution (`${A:-${B}}`, both unset)
  - Deep-nesting bound once a cap exists (finding #5)
  - Platform key with escaped-only value (`KAMIWAZA_X: "$${LIT:-x}"`) — behavior change vs pre-PR, currently unpinned

## Reviewer Agreement Matrix

| Category | Claude | Codex | Kamiwaza | Antigravity | Consensus |
|----------|--------|-------|----------|-------------|-----------|
| Security (host env) | Suggestion | — | Suggestion | Critical | Overruled by user — non-prod dev path, documented Compose-parity feature |
| Correctness (dict regression) | Important | — | Important | — | High |
| Compose parity | Important (unbraced `$VAR`) | — | Suggestion (`:+` re-scan) | High (brace fallback — rejected) | High (#2), Medium (#5) |
| Performance/robustness | Suggestion (recursion) | — | — | High (quadratic — disputed) | Medium |
| Ergonomics (silent drops) | Suggestion | — | Suggestion | Medium | Medium |
| Tests | Needs specific gaps | — | Praised coverage | Low (gaps) | Adequate + targeted additions |

## Reviewer Verdicts

| Claude | Comment (explicit; Confidence: High) |
| Codex | Approve (derived — clean pass, zero findings) |
| Kamiwaza | Request Changes (explicit; Risk: Medium) |
| Antigravity | Request Changes (explicit; Risk: Critical — but its Critical and both High findings were overruled in adjudication) |

**Vote Tally**: Approve: 1 | Comment: 1 | Request Changes: 2 (out of 4 successful reviewers)

- Empirically verified Compose 5.1 parity where it counts: all six operator branches under set/empty/unset, `$$` escapes, balanced JSON defaults, asymmetric literal braces, and multiple substitutions per value match real `docker compose config` output exactly (CL).
- The balanced-brace scanner with greedy final-brace fallback is a simple, linear, faithful model of Compose's parser — no pathological regex backtracking (CL, KZ).
- Excellent test hygiene: `KZ_TEST_*` prefixes plus explicit `delenv`/`setenv` make the suite immune to host-env pollution — essential now that the code reads `os.environ` (CL).
- Destination separation preserved: pre-resolution compose is captured for the catalog overlay (platform substitutes at install time); only the direct K8s payload is resolved; `KAMIWAZA_*` ConfigMap precedence intact (CL, KZ).
- Strictly less lossy than before for valid syntax (embedded substitutions like `neo4j/${PASSWORD:-x}` were previously dropped wholesale) (CL).
- The `KAMIWAZA_*` drop rule correctly narrowed to placeholder-bearing entries only (KZ). Docstrings updated accurately, including the honest caveat about reading the process environment (CL, KZ).

## Overall Verdict

**Risk Level**: Medium
**Reviewer Consensus**: 2/4 voted Request Changes outright; rule basis is the confirmed High-priority regression (#1), which was verified by two reviewers plus the orchestrator and has a small, well-understood fix. With #1 fixed and a regression test added, this converts to Approve/Comment territory.

---

Legend: CL=Claude, CX=Codex, KZ=Kamiwaza (Kbox/Pi, DeepSeek-V4-Flash-0731), AG=Antigravity (agy)

m9e:
Follow-up at `1b59cc0` on the latest four-model review:

- **#1 dict-entry data loss: fixed end to end.** The transformer now resolves tolerated `name`/`value` environment-list entries without changing their shape, and `PayloadBuilder` serializes that shape as one correctly named env entry. The regression test covers plain `$$`, a braced default, final payload env, and the matching `service-ref-rewrites` annotation. (This is a repository-tolerated extension rather than valid Docker Compose syntax, but existing detector/builder surfaces already accepted it.)
- **#2 unbraced `$VAR`: intentionally deferred.** The gap is real but unchanged from the merge-base; this PR and its public docs explicitly scope host capture to supported braced syntax. Expanding the capture grammar should be a separate compatibility change.
- **#3 malformed/required diagnostics: intentionally deferred.** Fail-closed dropping predates this PR and is the documented direct-deploy contract. The new tests pin malformed and set-empty required forms. Reporting dropped keys/reasons remains a useful observability follow-up, not a correctness blocker here.
- **#4 deep nesting: fixed.** Nested evaluation is bounded and fails closed instead of raising an unhandled `RecursionError`; sequential substitutions do not consume the depth budget.
- **#5 alternate re-scanning: refuted.** Docker Compose 5.1 recursively interpolates selected `+`/`:+` alternates. `${A:+${B}}` resolves to `B`'s value when selected, matching this implementation; a regression test now pins it.
- **#6 `KAMIWAZA_*` source names: intentional.** The guard is destination-key based to protect ConfigMap `envFrom` precedence. A non-platform key that aliases an exported platform-named host variable is explicit author behavior and is already covered by the PR security disclosure.
- **#7 naming/docs: fixed.** The helper and dash-only test names now describe their expanded/current roles.

Validation: `make test` reports 3,392 passed, 2 skipped, 260 deselected; the transformer suite reports 135 passed; changed-file Ruff/isort and targeted mypy pass. An independent final seam audit found no blocker and specifically verified shape preservation, depth accounting, Compose nested-alternate parity, and CodeScene-safe helper complexity.

m9e:
# PR Review Report: ENG-10260: Resolve compound Compose defaults in kz-ext dev

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/261
**Branch**: fix/dev-compound-env-defaults -> develop
**Scope**: 4 files, +568/-66

## Critical Issues
| # | File:Line | Issue and Evidence | Flagged By | Category |
|---|---|---|---|---|
| — | — | None | — | — |

## High Priority Issues
| # | File:Line | Issue and Evidence | Flagged By | Category |
|---|---|---|---|---|
| — | — | None | — | — |

## Medium / Low Suggestions
|---|---|---|---|
| 1 | `compose_transformer.py:629-649` | **Host-env capture is silent.** `_resolve_compose_substitution` now reads `os.environ`, so a bare `${SECRET}` that was previously dropped ("no safe value to ship") now embeds the developer's real host value into the `CreateExtension` payload (cluster CR env, and potentially the `service-ref-rewrites` annotation). This is the PR's intent (Compose parity; `test_uses_exported_value_for_bare_placeholder` codifies it with a var named `SECRET_ENCRYPTION_KEY`), but the only caller (`commands/dev.py:787`) prints nothing about which keys were captured from the host vs. collapsed to defaults. Recommend a console notice listing host-resolved keys, plus a changelog/docs entry, so a developer does not unknowingly ship a laptop secret to a shared cluster. | CL, KZ |
| 2 | `compose_transformer.py:544-559` | **Silent drops now cover a wider, host-dependent surface.** Entries resolving to `None` (unterminated `${VAR:-no-close`, unsatisfied `${VAR:?msg}`, depth-cap trips) vanish with zero diagnostics, and resolution is now host-sensitive — the same compose file can deploy with a var present on one machine and silently missing on another. Compose aborts loudly on `:?`; this pipeline drops the author's own "required" signal silently. Drop-silently predates the PR, but the PR substantially widens when it fires. Recommend logging dropped keys in `kz-ext dev`. | CL, KZ |
| 3 | `compose_transformer.py:679-685` / `payload_builder.py:460-463` | **Mapping-shape dict entries in an environment list still bypass resolution.** `_entry_has_shell_ref` returns False for dicts without a `name` key, yet `_parse_env_dict_item` converts `{"FOO": "${BAR:-x}"}` items — so that shape ships the placeholder verbatim and skips the `KAMIWAZA_*` drop rule. Pre-existing gap, verified by the lead in the checkout; worth closing (or explicitly rejecting) now that the `name`/`value` shape was fixed. | CL |
| 4 | `dev_env_image_refs.py:115-122` | **`name`/`value` dict entries skip the ENG-7110 image-ref rewrite.** The rewriter's list branch only handles string entries (`isinstance(entry, str)`), so an `AGENT_SERVER_IMAGE` declared in dict shape now survives resolution (new `_resolve_dict_list_entry`) but is skipped at the rewrite step. Verified by the lead. Follow-up for shape parity. | CL |
| 5 | `payload_builder.py:464-467` | **Extra Kubernetes keys (e.g. `valueFrom`) are stripped at payload build** while `_resolve_dict_list_entry` preserves them at resolution (`out = dict(entry)`). If `valueFrom` is intentionally unsupported, fine — but the two layers should agree (carry or explicitly reject). | CL, AG |
| 6 | `compose_transformer.py:597-599` | **`$$` unescape crosses into Kubernetes' own `$(VAR)` expansion regime.** A compose author's `"$$(...)"` (intending literal `$(...)`) arrives in-pod as `$(...)` and can be expanded by K8s dependent-env expansion if the name matches a container env var. Narrow; worth an operator-docs note. | CL |
| 7 | `compose_transformer.py` (`_resolve_compose_value` docstring) | **Unbraced `$VAR` stays literal** (tested by `test_unbraced_reference_remains_literal`) while real Compose interpolates it. Deliberate deviation — document it explicitly in the docstring so future maintainers don't read it as a bug. | CL, AG |
| 8 | `tests/unit/extensions/test_compose_transformer.py` | **Missing test:** `KAMIWAZA_*` key with a literal non-braced `$` (e.g. `cost$5`) is kept because `_has_braced_substitution` is False — no test locks in that intended behavior. A mixed literal-`$`-plus-substitution case (`"cost$v ${A:-b}"`) is also untested. | KZ |

## Test Coverage Assessment

Strong across all reviewers. ~20 focused unit tests cover the full operator matrix (`:-`, `-`, `:?`, `?`, `:+`, `+`) across set/empty/unset axes, `$$` escapes (plain, mixed, platform-key), embedded and multi-substitution values, nested defaults/alternates, literal-brace defaults (balanced JSON and unmatched `{`), the depth-100 recursion cap failing closed, malformed unterminated substitutions, and the `name`/`value` dict list shape — plus an end-to-end payload test proving a dict entry survives resolution, payload build, and the `service-ref-rewrites` annotation. Existing tests were correctly migrated to `KZ_TEST_*` names with `monkeypatch` isolation, exactly what the new host-env sensitivity requires. Minor gaps: items 3 and 8 above, and no test for `$$(...)` values (item 6).

## Reviewer Status and Verdicts
| Reviewer | Status | Capability | Verdict | Notes |
|---|---|---|---|---|
| Claude | success | repo tools, isolated checkout | Comment | Risk Low; verified integration seams in `dev.py`, `dev_env_image_refs.py` |
| Codex | success | native `codex review`, repo tools | No explicit verdict | "No actionable regression was identified in the authoritative diff"; also compiled changed modules |
| Kamiwaza (DeepSeek-V4-Flash-0731) | success (degraded) | diff-only, no tools (kzinfer transport; not truncated) | Comment | Risk Medium; one finding adjudicated benign by lead audit (see below) |
| Antigravity (Gemini 3.1 Pro) | success | repo tools, isolated checkout | Approve | Risk Low |

**Vote Tally**: Approve: 1 | Comment: 2 | Request Changes: 0
(Codex issued no formal verdict; its native conclusion — no actionable regressions — is reported verbatim and excluded from the tally.)

## Positive Observations

- The balanced-brace scanner (`_compose_substitution_end`) with the greedy `rfind` fallback cleanly solves the JSON-in-default problem, and the scan-then-anchored-regex design avoids catastrophic backtracking (AG).
- Fail-closed everywhere: malformed input, depth-cap trips, and unsatisfied required forms return `None` rather than shipping a mangled value; `_MAX_COMPOSE_SUBSTITUTION_DEPTH = 100` bounds recursion well under Python's stack limit (CL, KZ, AG).
- Substitution results are correctly not re-interpolated — host values containing `${...}` ship literally, matching Compose (CL).
- The `KAMIWAZA_*` drop rule is preserved and correctly gated on *unescaped* braced substitutions, with a dedicated `$${...}` escape test (CL).
- `_parse_env_dict_item` fixes a real pre-existing mangling bug: `{"name": "KEY", "value": "v"}` list items previously exploded into nonsense `{"name": "name", "value": "KEY"}` entries (CL, KZ, AG).
- The catalog-vs-K8s split in `dev.py` (catalog keeps placeholders for install-time substitution; only the direct K8s payload is resolved) is untouched, with comments updated to match (CL).

## Lead Cross-Cutting Audit

I read the complete authoritative diff independently and verified the semantics matrix against Compose behavior (all six operators across set/empty/unset), the brace-matching edge cases (balanced JSON defaults, unmatched-`{` greedy fallback), and the recursion bounding, before reading any lane result.

Adjudications with repository evidence:

- **Kamiwaza's `_entry_has_shell_ref` concern is benign.** The function is consumed in exactly one place — the list-branch gate in `_resolve_shell_refs` (`compose_transformer.py:554`) — not by URL rewriting as the diff-only lane hypothesized. The `"${"` → `"$"` widening is *required* so `$$` escapes in list entries get processed, and plain `cost$5`-style values round-trip unchanged through `_resolve_compose_value`. No behavior hazard.
- **Claude's integration-seam findings verify.** `rewrite_env_image_refs`'s list branch skips non-string entries (`dev_env_image_refs.py:117`), so dict-shape image vars miss the ENG-7110 rewrite (item 4); `dev.py:787` emits no console feedback about host-resolved or dropped keys (items 1–2); the mapping-shape bypass (item 3) reproduces from the code paths.
- **Seam check, resolution ↔ payload build:** the layers disagree on extra dict keys (preserve vs. strip, item 5) — no correctness failure today, but a latent trap if `valueFrom` support is ever assumed.
- **The one genuine design pivot** — consulting the host environment, which the pre-PR comment explicitly rejected — is deliberate, scoped to the direct K8s dev-deploy path only (the catalog overlay keeps placeholders for install-time substitution), and matches the ENG-10260 title. Its residual risk is operational (items 1–2), not correctness.

## Overall Verdict
**Risk Level**: Low
**Recommendation**: Comment
**Rationale**: No verified Critical or High issue remains — four independent lanes plus the lead audit found no correctness regression, and the implementation is high quality with thorough, well-isolated tests. Comment rather than Approve because the review surfaced material nonblocking findings worth acting on: the intentional-but-silent host-environment capture (a developer can unknowingly ship a laptop secret to a shared cluster) and the widened silent-drop surface both deserve console feedback in `kz-ext dev`, and the dict-shape parity gaps (items 3–5) are cheap to close while the area is fresh.

---
Legend: CL=Claude, CX=Codex, KZ=Kamiwaza, AG=Antigravity

m9e:
Follow-up at `86bdb75` on the latest fresh review's remaining seam suggestions:

- **#1 host-env capture notice / #2 silent-drop notice: intentionally deferred.** Both are worthwhile observability improvements, but they require a diagnostics/result contract across the resolver and `kz-ext dev` rather than a local parser patch. Host capture is the explicit feature being delivered and is disclosed in the PR security note; fail-closed dropping predates this PR. Neither reviewer classified these as correctness blockers.
- **#3 mapping-fragment list entries: fixed for the repository-tolerated shape.** Each mapping member now resolves independently, including `$$` handling and the destination-key `KAMIWAZA_*` drop rule. Docker Compose itself rejects this list shape, so it remains compatibility tolerance rather than an advertised Compose syntax.
- **#4 `name`/`value` image-ref rewrite parity: fixed.** Literal `AGENT_SERVER_IMAGE` values in the tolerated dict shape now pass through the ENG-7110 rewrite; `valueFrom` entries remain untouched.
- **#5 `valueFrom` preservation: fixed end to end.** Payload serialization now retains Kubernetes env-entry fields instead of stripping them. The platform schema describes these entries as K8s env format, copies them into the CR, and the operator's `corev1.EnvVar` contract preserves `ValueFrom`.
- **#6 Kubernetes `$(VAR)` follow-on: documented.** The resolver doc now calls out that `$$` collapse can produce a token Kubernetes later expands.
- **#7 unbraced `$VAR` scope: documented explicitly.** The resolver intentionally handles braced Compose substitutions only.
- **#8 literal/mixed dollar coverage: added.** Tests now pin a literal non-braced dollar under a `KAMIWAZA_*` key and a mixed raw-dollar-plus-braced-substitution value.

Validation on the pushed head: `make test` reports **3,394 passed, 2 skipped, 260 deselected**; the three focused extension suites report **231 passed**; changed-file Ruff/isort, targeted mypy, and `git diff --check` pass. A separate integration/CodeScene audit found no blocker and confirmed the SDK → platform → operator `valueFrom` contract.

m9e:
# PR Review Report: ENG-10260: Resolve compound Compose defaults in kz-ext dev

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/261
**Branch**: fix/dev-compound-env-defaults -> develop
**Scope**: 6 files, +691/-81

## Critical Issues
| # | File:Line | Issue and Evidence | Flagged By | Category |
|---|-----------|--------------------|------------|----------|
| — | None | | | |

## High Priority Issues
| # | File:Line | Issue and Evidence | Flagged By | Category |
|---|-----------|--------------------|------------|----------|
| — | None | | | |

## Medium / Low Suggestions
|---|-----------|------------|------------|
| 1 | `compose_transformer.py:611-654` | Unresolvable substitutions drop the **entire env entry silently** — no warning is printed anywhere (`compose_transformer.py` has no console/logger call; `commands/dev.py:787` discards the info). Sharpest for `:?`: the compose author demanded a *loud* failure and gets the quietest one (real Compose v5.1.0 renders `postgres://user:${DB_PASS}@db/app` with unset `DB_PASS` as `postgres://user:@db/app`; this code drops `DATABASE_URL` entirely). Same root cause makes `${V:+${UNSET}}` drop the whole entry where Compose yields an empty inner value. Fail-closed dropping is a documented, defensible policy — add a one-line warning per dropped key. | CL, KZ |
| 2 | `compose_transformer.py:641-644`; `commands/dev.py:787` | Host-env capture is silent: nothing tells the developer *which* values came from their shell (e.g. an exported `OPENAI_API_KEY` previously dropped is now shipped; incidental `DEBUG`/`PORT`/`USER` collisions shadow compose defaults). Resolved values also flow into the `service-ref-rewrites` annotation (shown by `test_payload_builder.py:965-973`), which is more widely readable than pod env. Suggest printing the **names** (never values) of host-resolved vars during `kz-ext dev`. Intent is documented in the PR body; only visibility is missing. | CL |
| 3 | `dev_env_image_refs.py:121-133`; `compose_transformer.py:779-793` | Env-walker matrix is inconsistent for mapping-fragment list entries: they are resolved (`_resolve_mapping_list_entry`) and payload-built (`_parse_env`), but invisible to `rewrite_env_image_refs` (falls through unchanged) and to `_iter_env_entries`/`detect_service_url_rewrites` (requires `name`+`value`). A fragment-declared `AGENT_SERVER_IMAGE` deploys unrewritten (ImagePullBackOff — the ENG-7110 failure mode). Gap is preexisting for two walkers, but this PR legitimizes the shape end-to-end — finish the matrix or reject fragments everywhere. | CL |
| 4 | `commands/dev.py:787` then `:830`; `dev_env_image_refs.py:145-158` | Resolve-then-rewrite ordering can clobber a deliberate host override: exporting `AGENT_SERVER_IMAGE=<same-repo>:custom-tag` gets resolved, then image-rewritten to the dev-built ref, discarding the tag override without notice. Note it in the existing caveat block or skip host-sourced refs in the rewriter. | CL |
| 5 | `dev_env_image_refs.py:47` | `_DEFAULT_SUB_RE` still uses `[^}]+`, so nested defaults (`${AGENT_SERVER_IMAGE:-${BASE:-ghcr.io/x/agent}:2.0.2}`) won't unwrap on the catalog-overlay path — the catalog-side rewriter now lags the resolver this PR upgraded. | CL |
| 6 | `tests/unit/extensions/` | Minor coverage gaps: no direct test that a `valueFrom` entry passes through `resolve_env_placeholders` unchanged; no test for `${VAR:-}` (empty default — traced correct); no drop-path test for a `{"name","value"}` entry with unresolvable value; no string-list-form `KAMIWAZA_X=${...}` platform-drop test. | KZ, CL |
| 7 | `compose_transformer.py:553,664` | Consistency nits, mostly preexisting: `_resolve_list_entry` doesn't strip the key before the `KAMIWAZA_` prefix check (leading-space entry evades the platform drop) while the image rewriter does strip; a YAML integer key with a `$`-bearing value would raise `AttributeError` in the dict path (`_resolve_mapping_list_entry` already does `str(key)`); bare `- VARNAME` host-inherit entries still pass through as valueless K8s entries now that host-env parity is otherwise real. | CL |

## Test Coverage Assessment

Strong — the standout of the PR. Two lanes independently executed the three touched suites: **231 passed** (Claude in an isolated venv; Codex during its native review). New tests cover all six operators (`:-`, `-`, `:?`, `?`, `:+`, `+`) parametrized across set/empty/unset, nested defaults and alternates, embedded mid-value substitutions, `$$` escapes (standalone, mixed, and on platform keys), balanced JSON braces in defaults plus the unmatched-`{` greedy fallback, the 100-depth fail-closed cap (200-deep input), malformed unclosed braces, unbraced `$VAR` literals, trailing bare `$`, `KAMIWAZA_*` drops across shapes, all three env-list shapes through resolution, `valueFrom` preservation in the image rewriter and `_parse_env`, and a full resolve→build round-trip asserting `service-ref-rewrites` annotation content. Test env vars are `KZ_TEST_*`-namespaced with proper `monkeypatch` hygiene — which matters now that the code reads `os.environ`. Gaps are minor (row 6 above).

## Reviewer Status and Verdicts
| Reviewer | Status | Capability | Verdict | Notes |
|----------|--------|------------|---------|-------|
| Claude | success | repo-tools-readonly (isolated checkout) | Approve | Ran 231 tests; empirically probed Docker Compose v5.1.0 parity on the hard interpolation cases; richest findings |
| Codex | success | repo-tools (native `codex review`) | None stated | Native output: "No actionable defects"; ran the 231 affected tests and a live `docker compose config` quoting probe; no explicit verdict — excluded from tally |
| Kamiwaza (DeepSeek-V4-Flash-0731) | degraded-success | diff-only-no-tools via kzinfer (kbox sandbox provisioning failed, exit 69); full untruncated diff | Approve | Flagged the `valueFrom` test gap and the `${V:+${UNSET}}` strictness divergence |
| Antigravity (Gemini 3.1 Pro) | success | repo-tools-readonly TUI | Approve | Pane stalled on a first-run workspace-trust dialog, then recovered and completed; no findings |

**Vote Tally**: Approve: 3 | Comment: 0 | Request Changes: 0

## Positive Observations

- The interpolation engine **empirically matches Docker Compose v5.1.0**: two lanes independently probed `docker compose config` (JSON-brace defaults, unmatched-`{` greedy fallback, `${V:-}`, `$$` escapes, multi-substitution values, set/empty/unset across all operators) and found no mismatch beyond the documented, deliberate drop-instead-of-empty/error policy.
- `_parse_env_dict_item` fixes real pre-existing data corruption: at the merge base, a K8s-style `{"name": "KEY", "value": "v"}` list item exploded into two bogus entries and `valueFrom` payloads were stringified into junk.
- Recursion is depth-limited and fail-closed (`_MAX_COMPOSE_SUBSTITUTION_DEPTH = 100`), recursing only on fallback nesting, so `RecursionError` is gone with no pathological blowup.
- The `KAMIWAZA_*` drop was refined correctly: platform keys with only `$$` escapes or bare `$` now pass through (escapes collapsed per Compose) while any unescaped `${` still defers to ConfigMap envFrom.
- Broadening the value gate from `"${" in v` to `"$" in v` makes `$$` collapsing apply to plain values (`cost$$value` → `cost$value`) — a quiet correctness fix matching what real Compose ships.
- The security consequences of host-env resolution are proactively documented in the PR body, including the `$(VAR)` post-`$$`-collapse Kubernetes expansion caveat.

## Lead Cross-Cutting Audit

I read the complete authoritative diff independently before consuming any lane result. My trace of the `_resolve_compose_substitution` operator table against Compose unset-vs-empty semantics matches on all six operators, and the parametrized test grid covers the full matrix. The balanced-brace scanner plus greedy `rfind` fallback is correct on the JSON-default and asymmetric-brace cases, and malformed `${` fails closed. I specifically checked the seam the PR widens: `_entry_has_shell_ref` broadened from `"${"` to `"$"` — its list-branch consumers pass non-ref entries through unchanged, `valueFrom` entries are never selected (value is not a string), and `$$`-only entries now correctly route through escape collapsing. The dedup above merges Claude's silent-drop finding with Kamiwaza's `${V:+${UNSET}}` strictness note (same root cause: `None` propagation drops the whole value). The cross-lane corroboration is unusually strong — two independent live Compose-parity probes and two independent 231-test runs. The one integration seam that remains genuinely unfinished is the mapping-fragment walker matrix (row 3): resolution and payload-building accept a shape that rewriting and annotation cannot see; it is nonblocking because the shape is nonstandard and the gap predates the PR, but it belongs on the ENG-10260 follow-up list.

## Overall Verdict
**Risk Level**: Low
**Recommendation**: Comment
**Rationale**: No Critical or High issues were found by any of the four lanes or the lead audit; the interpolation engine is demonstrably Compose-accurate and exceptionally well-tested. The recommendation is Comment rather than Approve because the review surfaced useful nonblocking findings that deserve author attention: silent whole-entry drops (including the inverted loudness of `:?`), silent host-env capture with values reaching the widely-readable `service-ref-rewrites` annotation, and the incomplete mapping-fragment walker matrix. None block merge; a one-line warning on drop and on host-resolution would close the two visibility gaps cheaply.

---
Legend: CL=Claude, CX=Codex, KZ=Kamiwaza, AG=Antigravity

m9e:
Follow-up at `ed1e386` on the latest Low-risk four-lane review:

- **Mapping-fragment rewrite parity: fixed.** The tolerated mapping-fragment list shape now participates in both ENG-7110 image-reference rewriting and cross-service URL rewrite detection, matching its resolver and payload-builder behavior. Regression tests cover both walkers.
- **Direct `valueFrom` pass-through coverage: added.** A Kubernetes `name`/`valueFrom` entry now has a direct resolver regression assertion in addition to the payload and image-rewriter tests.
- **Silent host capture / fail-closed drop diagnostics: intentionally deferred.** These remain worthwhile, nonblocking UX improvements, but require a diagnostics/result contract between the pure resolver and `kz-ext dev`. The host-capture behavior is the explicit feature and is disclosed in the PR security note; fail-closed dropping is the existing documented policy. Values should never be logged, and even key-only reporting deserves a separate scoped change.

Validation on the pushed head: `make test` reports **3,396 passed, 2 skipped, 260 deselected**; the focused transformer/payload/image-rewrite suites report **233 passed**; changed-file Ruff/isort, targeted mypy, and `git diff --check` pass.

m9e:
Interim status on the latest head: no new Critical or High correctness/security finding has appeared. CodeScene did flag its **critical code-health rule** (`Deep, Nested Complexity`) for `_iter_env_entries` at `ed1e386`; that is a merge gate rather than a runtime defect. The mapping-fragment behavior and tests are green, and I am flattening the walker into small shape-specific helpers now without changing semantics.

m9e:
CodeScene follow-up at `8a4d9e8`: flattened `_iter_env_entries` into three small, shape-specific helpers, preserving the mapping-fragment rewrite behavior while removing the flagged `Deep, Nested Complexity`. Focused transformer/payload/image-rewrite suites remain **233 passed**; Ruff, isort, targeted mypy, and `git diff --check` pass.

m9e:
Interim **Critical/High candidate finding** from the current Kamiwaza lane at `8a4d9e8` (posted immediately per review-loop policy): it claims Kubernetes `name`/`valueFrom` entries are dropped because `_resolve_dict_list_entry` rejects entries without a string `value`, and labels the list path's resolution as unconditional.

Immediate adjudication evidence currently **refutes** the claim:

- `_resolve_shell_refs` calls `_resolve_list_entry` only when `_entry_has_shell_ref(entry)` is true; a `name`/`valueFrom` entry has no string `value`, so the gate is false and the original entry is appended unchanged.
- The exact direct regression assertion is present in `test_resolves_name_value_dict_list_entries` and passes on this head.
- The current focused transformer/payload/image-rewrite run is **233 passed**, including that assertion.

I am leaving the head unchanged while the remaining independent lanes and the lead cross-cutting audit verify this candidate. If they find a real triggering path, I will fix it; otherwise the final report will record it as refuted with this evidence.

m9e:
# Current-head review report (`8a4d9e8`)

**Critical / High:** None verified.

## Interim candidate disposition

The Kamiwaza lane's first, validator-rejected draft alleged that `name`/`valueFrom` entries are dropped and labeled that Critical/High. That claim is **refuted**:

- `_resolve_shell_refs` calls `_resolve_list_entry` only when `_entry_has_shell_ref(entry)` is true.
- For a `name`/`valueFrom` entry, there is no string `value`, so the gate is false and the original entry is appended unchanged.
- The exact direct assertion in `test_resolves_name_value_dict_list_entries` passes on `8a4d9e8` (`1 passed`; full focused set `233 passed`).
- The format-correct Kamiwaza retry no longer reports the claim and returns **Low / Approve**; Antigravity independently inspected the code and explicitly confirms `valueFrom` preservation (**Low / Approve**).

## Reviewer status

- **Codex:** success, no actionable finding; confirmed 233 focused tests pass.
- **Kamiwaza / DeepSeek-V4-Flash-0731:** Low risk, Approve; untruncated diff-only transport.
- **Antigravity / Gemini 3.1 Pro:** Low risk, Approve; repository-tool review.
- **Claude repository-tools lane:** unavailable for a durable verdict after hitting its account session limit late in the review. This is recorded as a degraded lane, not treated as approval. The immediately preceding `ed1e386` Claude audit was Low/Approve and its only actionable consistency seam is what `ed1e386`/`8a4d9e8` closed.
- **Lead cross-cutting audit:** no blocker. The `8a4d9e8` delta only flattens `_iter_env_entries` into shape-specific helpers; tests preserve dict, string-list, `name`/`value`, `valueFrom`, and mapping-fragment behavior. CodeScene now passes.

## Remaining nonblocking items

- Key-only diagnostics for host-env capture and fail-closed drops would improve `kz-ext dev` UX, but require a scoped resolver→CLI diagnostics contract; never log values.
- A tolerated mapping fragment with `None` still serializes as the string `"None"`; this is pre-existing, nonstandard Compose input and not a blocker.

## Validation / gates

- `make test`: **3,396 passed, 2 skipped, 260 deselected**
- Focused transformer/payload/image-rewrite suites: **233 passed**
- Ruff, isort, targeted mypy, and `git diff --check`: pass
- GitHub: Unit Tests, Ruff, mypy, coverage, cyclomatic complexity, package build, seeder build, Linear gate, and CodeScene: **all pass**

**Risk:** Low  
**Recommendation:** Ready from the code/review perspective. GitHub remains `REVIEW_REQUIRED` with `jxstanford` requested; that is the remaining policy gate.

m9e:
# PR Review Report: ENG-10260: Resolve compound Compose defaults in kz-ext dev

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/261
**Branch**: fix/dev-compound-env-defaults -> develop
**Scope**: 6 files, +781/-93

## Critical Issues
| # | File:Line | Issue and Evidence | Flagged By | Category |
|---|-----------|--------------------|------------|----------|
| None | — | — | — | — |

## High Priority Issues
| # | File:Line | Issue and Evidence | Flagged By | Category |
|---|-----------|--------------------|------------|----------|
| None | — | — | — | — |

No verified blocking findings. The Kamiwaza lane's sole blocking claim — that `_resolve_dict_list_entry` drops `{"name": ..., "valueFrom": ...}` entries and fails its own test — was refuted with direct evidence: `_resolve_shell_refs` gates list entries on `_entry_has_shell_ref` (compose_transformer.py:558-566, outside the diff hunks visible to that diff-only lane), which returns False for valueFrom-only entries, so they pass through unchanged. The exact test it predicted would fail (`test_resolves_name_value_dict_list_entries`) passes in the isolated checkout, and the Antigravity lane independently confirmed valueFrom preservation.

## Medium / Low Suggestions
|---|-----------|------------|------------|
| 1 | compose_transformer.py:635-655 | **[Medium]** Host-env resolution is silent. `${BACKEND_URL:-http://backend:8000}` now resolves to whatever the developer's shell has exported; a stray export also suppresses the corresponding `service-ref-rewrites` entry, producing hard-to-trace cluster deploy failures. Suggest `run_dev_remote` print a short summary of env keys resolved from the host environment and keys dropped as unresolvable. Nonblocking per the standing kz-ext dev adjudication (dev-path host-env findings are ergonomics, Medium at most). | CL |
| 2 | compose_transformer.py:730-776, payload_builder.py:54 | **[Medium]** Host-resolved values (potentially secrets, e.g. `API_URL: "http://backend:8000/?token=${API_TOKEN}"`) are serialized into the `extensions.kamiwaza.io/service-ref-rewrites` annotation as plain `from`/`to` strings — a broader surface than the pod env (annotations are readable CR metadata and commonly logged). Consider skipping or masking annotation emission for values whose resolution consumed host-env data. Nonblocking per the same adjudication. | CL |
| 3 | dev_env_image_refs.py:47 | **[Medium-Low]** The catalog-overlay image rewrite regex `_DEFAULT_SUB_RE` remains single-level, so a compound default on a coupled image env name (e.g. `AGENT_SERVER_IMAGE: ${AGENT_SERVER_IMAGE:-${AGENT_IMAGE:-ghcr.io/org/agent:2.0.2}}`) is not unwrapped on the `catalog_compose` path, risking an ENG-7110-style ImagePullBackOff on catalog install. Pre-existing limitation the PR narrows rather than introduces; extending the rewriters to reuse the new balanced-brace scanner would finish the job. | CL |
| 4 | compose_transformer.py:503-504 | **[Low]** `$$(VAR)` semantics changed: pre-PR `prefix-$$(POD_IP)` shipped verbatim (kubelet-escaped to literal `$(POD_IP)`); post-PR it collapses to `prefix-$(POD_IP)`, which kubelet expands. Acknowledged in the docstring; worth a release-note line. | CL |
| 5 | compose_transformer.py:653-654 | **[Low]** Unsatisfied `${VAR:?msg}` drops the entry silently; real Compose aborts with the message. A console warning (or hard error for `:?` specifically) would surface misconfiguration at deploy time instead of pod runtime. | CL |
| 6 | compose_transformer.py:700-708 | **[Low]** Compose host-pass-through forms — bare list entry `"SOME_VAR"` and dict-form `VAR: null` — still pass through unresolved, an asymmetry now that `${VAR}` reads the host env. Resolve them or document the exclusion alongside the unbraced-`$VAR` note. | CL |
| 7 | compose_transformer.py:668-675 | **[Low]** The greedy unmatched-brace fallback in `_compose_substitution_end` scans `rfind("}")` over the entire remaining value, so a value mixing an unbalanced-brace default with a later `${...}` substitution (e.g. `${A:-{x} ${B:-y}`) is swallowed into one expression, diverging from Compose on this malformed edge. Fail-closed either way; noting for parity completeness. | Lead |
| 8 | compose_transformer.py:485-505 | **[Low]** The host-env consultation change deserves prominent release-note treatment beyond the updated docstrings, since it reverses the previous "defaults always win" behavior for every existing dev workflow. | KZ, CL |

## Test Coverage Assessment
Strong and environment-hygienic: every host-env-sensitive test monkeypatches `KZ_TEST_*` names, making the suite immune to the runner's shell. Coverage spans all six operator forms with empty-vs-unset distinctions, nested defaults and alternates, `$$` escapes in dict/list/platform-key positions, literal JSON braces and the greedy unmatched-brace fallback, embedded/multi-substitution values, name/value and mapping-fragment list shapes through resolution → image rewrite → payload build, depth-limit and malformed-input fail-closed paths, and `_parse_env` valueFrom passthrough. The 233 tests in the three touched modules pass (verified independently by the Claude and Codex lanes). Minor gaps: no test for the platform-key drop in the string list shape (`"KAMIWAZA_X=${...}"`), and no dev.py-level test pinning the interaction where a host-exported var suppresses a `service-ref-rewrites` entry (the failure mode in suggestion 1).

## Reviewer Status and Verdicts
| Reviewer | Status | Capability | Verdict | Notes |
|----------|--------|------------|---------|-------|
| Claude | success | repo tools, isolated checkout | Approve | Operator-by-operator Compose-parity check; ran the touched test modules (233 passed) |
| Codex | success | native `codex review`, repo tools | (none stated) | rc 0; reported no actionable correctness defects and passing focused suites; excluded from tally — no explicit verdict to record |
| Kamiwaza (DeepSeek-V4-Flash-0731) | degraded success | kzinfer diff-only, no tools, diff untruncated | Request Changes | Sole blocking finding refuted by lead with direct evidence (gating code + passing test); kbox-pi transport failed provisioning (exit 69), fell back per contract |
| Antigravity (Gemini 3.1 Pro) | success | repo tools, interactive TUI | Approve | Lane recovered from a workspace-trust-dialog stall; valid durable review with DONE sentinel |

**Vote Tally**: Approve: 2 | Comment: 0 | Request Changes: 1

## Positive Observations
- The operator dispatch in `_resolve_compose_substitution` is a faithful, compact encoding of Compose's set/empty matrix; neither the Claude lane nor the lead audit could construct a supported input where it disagrees with Compose (CL, AG, Lead).
- `_parse_env_dict_item` fixes a real pre-existing corruption bug: K8s-style `{"name": "X", "value": "y"}` items previously exploded into `{"name": "name", ...}` nonsense entries; they now round-trip correctly with `valueFrom` preserved (CL and AG independently).
- Fail-closed discipline is consistent — malformed substitutions, unbalanced braces, over-deep nesting (`_MAX_COMPOSE_SUBSTITUTION_DEPTH`), and unresolvable refs drop entries rather than shipping mangled values (CL, KZ, AG).
- The `KAMIWAZA_*` envFrom-shadowing guard is preserved, including honoring `$$`-escaped platform values as explicit author choices (CL).
- Tests are hygienic and comprehensive; the greedy-brace fallback elegantly matches Compose behavior (AG, KZ).

## Lead Cross-Cutting Audit
Independent read of the full authoritative diff plus targeted checkout verification:
- Traced every operator branch of `_resolve_compose_substitution` against Compose's unset-vs-empty matrix; all branches agree, including `?`/`:?` empty-vs-unset distinctions and `+`/`:+` alternates resolving their fallback recursively.
- Adjudicated the Claude/Kamiwaza disagreement with direct evidence: read the `_entry_has_shell_ref` gate in the checkout and executed the exact disputed test (`1 passed`). The Kamiwaza Request Changes verdict rests entirely on that refuted finding — a capability artifact of diff-only review, not a code defect.
- Seam check (resolution → `detect_service_url_rewrites` → annotation → payload): the new end-to-end test pins name/value resolution through payload env to annotation consistency. The same seam is where host-derived values reach the annotation (suggestion 2).
- Seam check (image-ref rewrite vs resolution ordering): rewriting operates on unresolved default expressions; a host-set image var wins unrewritten, which is semantically correct — an explicit developer override.
- Applied the standing adjudication that the direct `kz-ext dev` path is non-production: host-env-exposure findings are capped at Medium ergonomics severity; production-path exposure does not apply since catalog templates remain unresolved for platform install-time substitution.

## Overall Verdict
**Risk Level**: Low
**Recommendation**: Comment
**Rationale**: No verified Critical or High finding remains — the single Request Changes vote rests on a refuted diff-only misread, confirmed by executing the disputed test against the checkout. The interpolation engine is Compose-correct in every audited branch, fails closed on malformed input, fixes a genuine pre-existing `_parse_env` corruption bug, and ships thorough, environment-hygienic tests. Comment rather than Approve because the nonblocking findings deserve author attention: silent host-env resolution and its suppression of service-ref rewrites (observability), host-derived values reaching the `service-ref-rewrites` annotation, the catalog-overlay rewriter's remaining blindness to compound defaults, and two release-note items (`$$(VAR)` kubelet-escape change, host-env behavior reversal).

---
Legend: CL=Claude, CX=Codex, KZ=Kamiwaza, AG=Antigravity

PR #262, opened 2026-08-13
ENG-3068: Ignore stale Failed phase during extension PATCH rollout

m9e:
## Summary

- Check the extension status endpoint before treating a top-level `Failed` phase as terminal.
- Continue polling when the status endpoint reports an active rolling update.
- Preserve fail-fast behavior when no rollout is active or status cannot be verified.

## Why

During PATCH, the extension resource can briefly retain the previous generation's `Failed` phase while the new generation is already rolling out. `kz-ext dev` should follow the active rollout instead of aborting on stale status.

## Validation

- `uv run pytest tests/unit/extensions/test_deployment_poller.py`: 8 passed
- Changed files pass Ruff

Linear: https://linear.app/kamiwaza/issue/ENG-3068

### DIFF
--- a/kamiwaza_extensions/deployment_poller.py
+++ b/kamiwaza_extensions/deployment_poller.py
@@ -67,6 +67,12 @@ class DeploymentPoller:
                 last_phase = phase
 
             if phase == "Failed":
+                if use_status_endpoint and self._is_rolling_update(
+                    client, extension_name
+                ):
+                    console.print("  [dim]Status: rolling update in progress[/dim]")
+                    time.sleep(poll_interval)
+                    continue
                 msg = self._extract_failure_message(ext)
                 raise DeploymentFailedError(
                     f"Extension deployment failed: {msg}"
@@ -103,6 +109,15 @@ class DeploymentPoller:
             f"Extension did not reach Running state within {timeout}s "
             f"(last phase: {last_phase})"
         )
+
+    @staticmethod
+    def _is_rolling_update(client: KamiwazaClient, extension_name: str) -> bool:
+        """Return whether a newer PATCH generation is still rolling out."""
+        try:
+            status = client.extensions.get_extension_status(extension_name)
+        except Exception:
+            return False
+        return status.rolling_update is True
 
     @staticmethod
     def _check_via_status_endpoint(
--- a/tests/unit/extensions/test_deployment_poller.py
+++ b/tests/unit/extensions/test_deployment_poller.py
@@ -78,6 +78,29 @@ class TestWaitForReady:
         with pytest.raises(DeploymentFailedError, match="ImagePullBackOff"):
             poller.wait_for_ready(client, "test", timeout=10)
 
+    @patch.object(DeploymentPoller, "_check_via_status_endpoint", return_value=(None, ""))
+    @patch.object(DeploymentPoller, "_check_pods_ready", return_value=(True, "2/2 ready"))
+    @patch("kamiwaza_extensions.deployment_poller.time.sleep")
+    def test_ignores_stale_failed_phase_during_patch_rollout(
+        self, mock_sleep, mock_pods, mock_status, poller
+    ):
+        client = MagicMock()
+        client.extensions.get_extension.side_effect = [
+            _make_ext("Failed"),
+            _make_ext("Provisioning"),
+            _make_ext("Running"),
+        ]
+        client.extensions.get_extension_status.return_value = ExtensionStatus(
+            name="test",
+            phase="Failed",
+            rolling_update=True,
+        )
+
+        result = poller.wait_for_ready(client, "test", timeout=10)
+
+        assert result.phase == "Running"
+        client.extensions.get_extension_status.assert_called_once_with("test")
+
     @patch("kamiwaza_extensions.deployment_poller.time.monotonic")
     @patch("kamiwaza_extensions.deployment_poller.time.sleep")
     def test_raises_on_timeout(self, mock_sleep, mock_monotonic, poller):


m9e:
# PR Review Report: ENG-3068 — Ignore stale Failed phase during extension PATCH rollout

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/262
**Branch**: `fix/eng-3068-ignore-stale-failed-patch-status` → `develop`
**Reviewed head**: `099285d5ee477e63527c5b023e0ea26a5ccd1d29` (re-checked; unchanged throughout)
**Merge-base**: `8d21d435eeafd59b41e8b657f780a3e7c0c0832e`
**Scope**: 2 files, +38/-0

> Authored by the same account that is reviewing (`m9e`), so GitHub will not accept an Approve or
> Request-Changes from me. Posted as a **Comment**; the verdict is stated at the bottom.

---

## Mechanism first: what `rolling_update` actually means

The guard's entire safety argument rests on a definition that lives in another repo. Looked up rather than
assumed — `kamiwaza/serving/garden/extensions/extensions_service.py:204-214` (checkout at `origin/develop`,
`git rev-list --left-right --count HEAD...origin/develop` = `0 0` after fetch):

```python
def _extension_status_rolling_update(metadata, cr_status) -> bool:
    generation = metadata.get("generation")
    observed_generation = cr_status.get("observedGeneration")
    return bool(generation is not None and observed_generation is not None
                and generation != observed_generation)
```

Pinned by the server's own tests (`tests/unit/serving/garden/extensions/test_patch_and_status.py:764`
gen=2/obs=2 → `False`; `:810` gen=3/obs=2 → `True`).

So `rolling_update` means **"the operator has not yet observed the current generation"** — a genuine staleness
signal. **The fix targets the right thing, at the right layer.** Everything below is about whether the guard is
*reachable*, *defended*, and *honestly tested* — not about the idea being wrong.

---

## High

### H1 — `deployment_poller.py:70` · `use_status_endpoint` is a one-way latch that silently disarms this entire PR

`use_status_endpoint` is cleared inside the `phase == "Running"` branch whenever `_check_via_status_endpoint`
returns `(None, "")` — which happens not only on 404/405/501 but on **any other `APIError` (5xx, auth) and any
bare `Exception`** (lines 137-146). It is never re-armed.

So **one transient `/status` error while the extension is still `Running` on the old generation** permanently
disables the new stale-`Failed` guard for the rest of the poll. "Extension is `Running` when you PATCH it" is
the *normal* case for a PATCH rollout — precisely the ENG-3068 scenario.

EXECUTED, two arms, deterministic fake clock, real `ExtensionStatus`, `sys.path` asserted against the tree
(poll 1 = `Running` with `/status` raising `APIError(500)` once; polls 2+ = `Failed`, `rolling_update=True`):

| Arm | Outcome | `/status` calls |
|---|---|---|
| **HEAD** (conjunct present) | `DeploymentFailedError: ...ImagePullBackOff: pull access denied` at **t=3.0s** | **1** |
| Conjunct dropped (`numstat 1 1`) | `DeploymentTimeoutError: ...within 120s (last phase: Failed)` at t=120.0s | 40 |

The builds diverge completely and **nothing tests the divergence** — which is why the mutant that drops the
conjunct survives (H2 table, arm M2). Under HEAD the guard is skipped even though the server would have
answered `rolling_update=True`: **the ENG-3068 symptom stays reachable through one earlier 5xx.**

Originated by one review lane, then independently re-executed by me before publishing.

**Fix**: drop the conjunct — it buys nothing, since `_is_rolling_update` already returns `False` on any
exception, so the only cost of always calling it is one wasted request on clusters with no `/status`. Or give
the `Failed` path its own flag rather than sharing the readiness-detail latch.

### H2 — `deployment_poller.py:116-119` · The fail-closed guarantee is both unpinned and silent

The description promises *"Preserve fail-fast behavior when no rollout is active **or status cannot be
verified**."* The second half is `except Exception: return False`. Nothing tests it, and it says nothing.

EXECUTED mutation battery (own `git clone -s` tree; baseline `8 passed in 0.27s`; every arm asserted
pattern-count == 1 and printed `git diff --numstat` as proof it landed; `PYTHONDONTWRITEBYTECODE=1`):

| Arm | Mutation | Applied | Result | Verdict |
|---|---|---|---|---|
| NOOP | split the return into two statements | 2/1 | 8 passed | **SURVIVED** (harness is not false-killing) |
| M1 | `is True` → `bool(...)` | 1/1 | 1 failed, 7 passed (10.63s) | KILLED — *incidentally*, see L-4 |
| **M2** | drop the `use_status_endpoint and` conjunct | 1/1 | **8 passed** | **SURVIVED** → H1 |
| **M3** | `except Exception: return False` → `return True` | 1/1 | **8 passed** | **SURVIVED** |
| M4 | delete the whole guard (positive control) | 0/6 | 1 failed, 7 passed | KILLED |
| M5 | remove `time.sleep(poll_interval)` | 0/1 | 8 passed | SURVIVED (sleep is mocked — not meaningful) |
| M6 | `continue` → fall through | 0/1 | 1 failed, 7 passed | KILLED |
| M7 | hoist the guard above `if phase == "Failed":` | 1/1 | 1 failed | KILLED (via `StopIteration`, not an assertion) |

Inverting fail-closed to fail-**open** (every `/status` error read as "a rollout is active", so a genuine
`Failed` polls to the full timeout) is invisible to the suite. A later refactor can silently delete the
guarantee the description sells.

**And the swallow is silent.** EXECUTED: with `get_extension_status` raising `APIError("Unauthorized",
status_code=401)`, `wait_for_ready` raises `DeploymentFailedError` and stderr contains neither `"401"` nor
`"Unauthorized"`. The sibling `_check_via_status_endpoint:141-142` prints `Status endpoint error: {exc}` for
exactly this class. On an expired token the user sees a *stale* failure message with no hint that the
staleness guard was skipped. The fail-closed *direction* is right; the silence and the missing pin are not.

**Validated remedy (executed, not merely proposed)** — `10 passed` at head, and
`test_fails_fast_when_status_endpoint_errors` is a **sole killer for M3**:

```python
@patch("kamiwaza_extensions.deployment_poller.time.sleep")
def test_fails_fast_when_status_endpoint_errors(self, mock_sleep, poller):
    client = MagicMock()
    client.extensions.get_extension.return_value = _make_ext("Failed", message="ImagePullBackOff")
    client.extensions.get_extension_status.side_effect = RuntimeError("boom")
    with pytest.raises(DeploymentFailedError, match="ImagePullBackOff"):
        poller.wait_for_ready(client, "test", timeout=10)

@patch("kamiwaza_extensions.deployment_poller.time.sleep")
def test_fails_fast_when_not_rolling(self, mock_sleep, poller):
    client = MagicMock()
    client.extensions.get_extension.return_value = _make_ext("Failed", message="ImagePullBackOff")
    client.extensions.get_extension_status.return_value = ExtensionStatus(
        name="test", phase="Failed", rolling_update=False
    )
    with pytest.raises(DeploymentFailedError, match="ImagePullBackOff"):
        poller.wait_for_ready(client, "test", timeout=10)
```

Neither covers H1. A **third** test is needed and is the sole killer for M2: `Running` → `/status` raises once
→ `Failed` with `rolling_update=True` → assert the poller still waits.

### H3 — `test_deployment_poller.py:81-102` · The new test is unrepresentative, and one assertion encodes a false production claim

EXECUTED: run the new test's *exact* fixture semantics with `_check_via_status_endpoint` and
`_check_pods_ready` **unpatched** →

```
DeploymentTimeoutError: Extension did not reach Running state within 120s (last phase: Running)
t=120.0, 39 get_extension_status calls
```

It only reaches `Running` because both readiness helpers are stubbed: the `(None, "")` stub trips the latch
into the kubectl path on the very first `Running` poll. In real code, `_check_via_status_endpoint:148-153`
treats a truthy `rolling_update` as *not ready* and loops.

So `client.extensions.get_extension_status.assert_called_once_with("test")` (diff line 63) is a **mock
artifact** — in production the `Running` branch calls that endpoint again, so "called once" is not a contract.

Compounding it, an assertion-drop matrix over the M2/M3/M4/M6/M7 arms comes back **completely flat**: with the
body reduced to the `wait_for_ready(...)` call plus `pass`, every kill still lands (M4/M6 raise; M7 exhausts
the 3-element `side_effect` into `StopIteration`). NOOP survived in all four rows, so the harness is not
false-killing. Both visible assertions are inert *and* one of them is wrong about production.

The test does carry weight — it is the only test that kills M4 and M6 — but purely as a "does not raise" test.

---

## Medium

**M-1 — `deployment_poller.py:108-111` · The timeout path destroys the diagnostic string, and it is worse remotely.**
Whenever the guard holds the loop to the deadline, `_extract_failure_message`'s detail is discarded for a bare
`(last phase: Failed)` — visible in H1's own two-arm output. For a **local** connection `diagnose_dev_timeout`
recovers well (operator `ImagePullBackOff` → `CLUSTER_NOT_READY`). For a **remote** connection it does not:
`dev_diagnostics.py:59-69` returns category `unknown` immediately with "this could be operator state or an
app-level startup failure". So remote `kz-ext dev` users pay up to 300 s *and* lose the message *and* get a
vague diagnosis. Cheap, invariant-independent fix: record the last `_extract_failure_message(ext)` seen while
skipping and interpolate it into `DeploymentTimeoutError`.

**M-2 — `deployment_poller.py:115` · The docstring asserts a cross-repo contract whose key leg is unverified.**
`"""Return whether a newer PATCH generation is still rolling out."""` overstates the server. It returns
`generation != observedGeneration` — *"the operator has not acknowledged the current spec"* — which **also
covers operator down / crashlooping / mid-upgrade**, where the condition is permanent, not transient. The API
server bumps `metadata.generation` on the PATCH regardless of operator health; only the controller ever writes
`observedGeneration`.

That leaves one load-bearing invariant unproven: *does the operator write `observedGeneration = generation` on
**failing** reconciles, or only on successful ones?* If only on success, `rolling_update` stays `True` on a
permanently-broken generation and the swallow is real. **The controller source could not be located** —
bounded absence: `kz/kamiwaza` @ `origin/develop` (read-side only), `kz/deploy` (CRDs + Helm only;
`charts/extension-operator/` has no source), `kz/operators` @ `origin/develop` (unrelated), and
`find /Users/matt/code -maxdepth 4` for `*controller*.go` / kubebuilder `PROJECT` → zero hits. Two indirect
arguments favour the invariant holding: the CRD carries both `status.observedGeneration` and per-condition
`observedGeneration` alongside a `phase` enum (standard controller-runtime split — if `observedGeneration` were
success-gated, `phase: Failed` would be *permanently* unobservable after any PATCH, i.e. the ENG-3068 symptom
would not be "brief"); and the platform's contract test conjoins `not rolling_update AND ready_replicas >= 2`
(`tests/contract/extensions/test_zero_downtime_patch.py:400-405`), which would be dead weight if
`rolling_update == False` implied a healthy rollout. **Graded ~80% confident, not settled.** Please state the
invariant in the PR body and ideally pin it with a contract test on a failing reconcile. M-1's fix makes it
largely moot either way.

**M-3 — `deployment_poller.py:62` + `70-72` · The guard is two non-atomic CR reads, so it does not close its own race.**
EXECUTED: `get_extension` returns `Failed(message="PREVIOUS generation ImagePullBackOff")` while
`get_extension_status` returns `rolling_update=False` (operator reconciled between the two GETs) → raises the
**stale** message anyway. Both server handlers read `cr_status` from the same CR, so a single
`get_extension_status` call yields phase *and* `rolling_update` atomically — `ext.phase` is redundant here.
Narrow window, but it is the same window the PR exists to fix.

---

## Low

- **L-1 — `:73`** The `rolling update in progress` line is not deduplicated the way `last_phase`/`last_ready`
  are. Executed: 40 identical lines over 120 s; ~100 at the caller's real 300 s default.
- **L-2 — `:54`** Stale docstring: `Raises: DeploymentFailedError: Extension reached Failed phase.` is no
  longer unconditional.
- **L-3 — `:70`** `get_extension_status` is a heavy server call (lists pods **and** events per invocation —
  `extensions_service.py:1352-1354`), now hit every `poll_interval` while `Failed`. Also: an extension that
  goes `Failed` without ever passing through `Running` issues one doomed call per poll against a 404 endpoint.
- **L-4 — `:120`** `is True` vs `bool(...)` is production-equivalent — `ExtensionStatus.rolling_update` is a
  pydantic `bool` (`schemas/extensions.py:245`), always coerced. The identity check only matters against
  `MagicMock`, which is why M1 died *accidentally*, via a 10.6 s real-clock spin in `test_raises_on_failed`
  (which does not patch `time.monotonic`). `test_fails_fast_when_not_rolling` replaces that accident with a
  real pin.
- **L-5 — out of scope, worth a line** The CRD `phase` enum is
  `Pending|Provisioning|Running|Degraded|Failed|Terminating` (`deploy/crds/kamiwazaextension-crd.yaml:3543-3551`).
  The poller treats only `Failed` as terminal, so a stuck `Degraded` already burns the full 300 s. Pre-existing.

---

## Refuted during review (recorded so they are not re-raised)

- **"This swallows genuine deployment failures."** Raised as **P1 by Codex Pass A**. Refuted *as stated*: once
  the operator reconciles the new generation and it genuinely fails, `observedGeneration == generation`,
  `rolling_update` goes `False`, and `Failed` is terminal again — executed: `rolling_update` True→True→False
  with phase pinned `Failed` raises `DeploymentFailedError: ...ImagePullBackOff` after exactly 3 status calls.
  Caveat in M-2.
- **"The SDK drops an `observedGeneration` the server exposes; the poller should check staleness directly."**
  My own framing, **refuted**. The server's `ExtensionStatus` (`schemas.py:787-803`) has **no generation field
  and no `extra="allow"`**, so the response body physically cannot carry it. `rolling_update` *is* the direct
  staleness check, just computed server-side — there is no narrower signal available at the SDK boundary.
- **"First-deploy (CREATE) regression."** Cannot occur. `_extension_status_rolling_update` requires **both**
  `generation` and `observedGeneration` non-`None`; on a fresh CREATE the status is empty (→ `False`) or
  `observedGeneration == 1 == generation` (→ `False`). `_delete_and_recreate` (`dev.py:1146`) also starts at
  generation 1. Additionally, k8s does not bump `metadata.generation` for the metadata-only annotation changes
  in `_build_patch_kwargs` (`dev.py:22-47`) on a CRD with a `status` subresource, so a no-op redeploy still
  fails fast on a pre-existing `Failed`.
- **"Exit-code / caller regression."** `commands/dev.py:1168-1196`: `DeploymentTimeoutError` routes through
  `diagnose_dev_timeout` and exits `CLUSTER_NOT_READY` **only** for operator-not-ready — otherwise `1`, the
  same code `DeploymentFailedError` produces. No regression. Note the real budget is
  `KAMIWAZA_DEV_TIMEOUT` default **300 s**, not `wait_for_ready`'s unused `timeout=120` default. The
  identically-named `kamiwaza_sdk.exceptions.DeploymentFailedError` (`exceptions.py:204`) is a distinct class —
  no aliasing, `seeding/cli.py:217` and `serving.py` are unaffected.
- **"The PR reddens the suite."** `tests/unit/extensions/`: merge-base **1 failed, 1775 passed, 1 skipped** →
  head **1 failed, 1776 passed, 1 skipped**. Delta is exactly **+1 passing = the new test**. The failure
  (`test_dev_canonical_refs.py::TestInsecurePreflightSource::test_push_call_uses_effective_insecure_not_persisted_flag`)
  reproduces at `8d21d435` — pre-existing and host-environmental. The skip is a BSD-`sed` host limitation.

---

## Test coverage assessment

**Verdict: Needs Work** — adequate for the happy path, absent for every guarantee.

The new test proves one thing: a `Failed` phase with `rolling_update=True` does not raise. It does not pin the
fail-closed path (M3 survived), does not pin the latch (M2 survived, = H1), its two visible assertions are
inert, one of them is a mock artifact, and it stubs out the very interaction where H1 lives.

Three tests needed: the two validated in H2, plus the latch test.

## Positive observations

- **The fix targets the right signal, and the alternatives genuinely do not exist.** `rolling_update` is
  `generation != observedGeneration` — and the server exposes nothing narrower, so this is not a lazy proxy.
  Choosing it over a timer or retry counter is correct.
- Fail-closed by intent (`except → False`, identity check rather than truthiness) — the right default for a
  guard that suppresses error reporting.
- +38/-0, purely additive, scoped to the single branch it needs to affect.
- The CREATE path is safe by construction, not by accident — the server's two-`None` check makes a first-deploy
  swallow impossible.
- The timeout fallback is better served than expected on local connections: `diagnose_dev_timeout` already
  distinguishes operator-not-ready from app-failure.

## Reviewer verdicts

| Reviewer | Lane / model | Verdict | Contribution |
|---|---|---|---|
| Main loop | Opus 5 | Request Changes | mutation battery, merge-base control, server-contract lookup, independent re-execution of H1 and H3 |
| Review lane 1 | Opus 5, own clone | Request Changes | **originated H1** (latch); flat assertion matrix |
| Review lane 2 | Opus 5, own clone | Request Changes | executed 401-silence (H2), PROBE_A/B (H3), M-3 race; refuted the CREATE and observedGeneration theories; suite counts |
| Codex Pass A | gpt-5.4, 1635 B stdout | Request Changes | P1 **refuted**; P2 corroborates H2's silence |
| Codex Pass B | gpt-5.4, 888 B stdout | Comment | P2 corroborates H2's silence |
| Kamiwaza | `kamiwaza/tokenator/DeepSeek-V4-Flash-0731`, `truncated:false` | Approve (Risk: Low) | independently named both missing negative tests and the resilience gap |

*Kamiwaza lane disclosure*: the tool's hardcoded default (`GLM-5.2-NVFP4`) is de-provisioned, so the lane ran
with an explicit `--model-name`; the echoed `model` field was verified. DeepSeek is an independent family, so
this is a genuine outside opinion rather than a second Claude.

**Vote tally**: Request Changes 4 | Comment 1 | Approve 1 (of 6 successful lanes).

---

## Overall

**Risk Level**: Medium

The design is right and I would take this change — two of my own suspicions (a weaker-than-necessary staleness
proxy, a first-deploy regression) were refuted during review, and the layer choice holds up. The problem is
that **as shipped, the fix is disarmed before it can run**: a single transient `/status` error during an
earlier `Running` poll latches it off permanently (H1, executed twice independently), so the ENG-3068 symptom
remains reachable — and the suite cannot see it, because both the latch and the fail-closed path are surviving
mutants.

Four small fixes close it:

1. **Drop the `use_status_endpoint` conjunct** (or give the `Failed` path its own flag) — H1.
2. **Add the three tests** — the two validated above plus the latch test — H2, and make the new test
   representative by dropping the `_check_via_status_endpoint` stub and replacing the `assert_called_once_with`
   mock artifact — H3.
3. **Log the swallowed exception** in `_is_rolling_update`, mirroring `_check_via_status_endpoint:141-142` — H2.
4. **Carry `_extract_failure_message` into `DeploymentTimeoutError`** — M-1, which also makes the unresolved
   operator-contract question (M-2) largely moot.

Folding the two CR reads into one `get_extension_status` call (M-3) is worth doing if it is cheap.

m9e:
# PR Review Report: ENG-3068: Ignore stale Failed phase during extension PATCH rollout

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/262
**Branch**: fix/eng-3068-ignore-stale-failed-patch-status -> develop
**Scope**: 2 files, +38/-0

## Critical Issues
| # | File:Line | Issue and Evidence | Flagged By | Category |
|---|---|---|---|---|
| - | - | None | - | - |

## High Priority Issues
| # | File:Line | Issue and Evidence | Flagged By | Category |
|---|---|---|---|---|
| - | - | None. One lane (CL) rated finding M1 below as High / Request Changes; a neutral four-model adjudication council (Codex, Kamiwaza/GLM, Antigravity, fresh Claude; orchestrator abstained) ruled it Medium / non-blocking 3-1 in round 1 and 4-0 in round 2 after the dissenter reviewed the frequency and F10 requeue-interval evidence. Adjudication transcripts retained in the run directory. | - | - |

## Medium / Low Suggestions
|---|---|---|---|
| M1 | `kamiwaza_extensions/deployment_poller.py:69-75` | **Persistent reconcile-step failure on a PATCH now waits the full timeout instead of failing fast (Medium).** The platform computes `rolling_update` as `metadata.generation != status.observedGeneration` (kamiwaza develop `extensions_service.py`). The operator advances `ObservedGeneration` only in its end-of-pass status write (operators develop@0eea2e8 `kamiwazaextension_controller.go:3977`); its `setFailedCondition` path (8 call sites: RBAC/Secret/Sandbox/ConfigMap/Service/NetworkPolicy/Ingress/Istio) writes `Phase=Failed` and requeues every 30 s without touching it. So an API-server rejection, admission denial, quota, or ingress collision on a PATCHed spec keeps `rolling_update=True` indefinitely, and the new branch loops for `KAMIWAZA_DEV_TIMEOUT` (300 s), prints ~100 `rolling update in progress` lines, then raises `DeploymentTimeoutError (last phase: Failed)` and runs the operator-oriented `diagnose_dev_timeout`. Pod-level failures (ImagePullBackOff, CrashLoop) still fail fast because the normal pass writes phase and `ObservedGeneration` together. Pre-PR the same case failed immediately but with a stale/empty message, so this is a bounded latency and diagnostics regression on a narrow path, not a correctness failure. Recommended follow-up: cap the stale-Failed tolerance at 2-3 operator requeue intervals (60-90 s) and, on timeout with `last_phase == "Failed"`, include `_extract_failure_message(ext)` plus any Warning `ExtensionStatus.events`; add a regression test for the persistent-Failed+rolling case. | CL, KZ |
| M2 | `kamiwaza_extensions/deployment_poller.py:116-119` | `_is_rolling_update` swallows every exception silently and returns False. Fail-closed is the right direction, but the sibling `_check_via_status_endpoint` (`:137-146`) distinguishes 404/405/501 from other errors and prints `Status endpoint error`. A transient 5xx or validation error during the stale window silently reproduces the original ENG-3068 abort with no hint. Mirror the sibling: quiet on 404/405/501, dim log otherwise. Also add a one-line comment stating the fail-closed intent so a future refactor does not "simplify" it the wrong way. | CL, KZ |
| M3 | `tests/unit/extensions/test_deployment_poller.py:71-79` | The `status.rolling_update is True` identity check is load-bearing for the pre-existing `test_raises_on_failed`: that test uses a bare `MagicMock()` client, so `get_extension_status(...).rolling_update` is a truthy MagicMock and only the `is True` check keeps the test from spinning to timeout. Nothing documents this. Set an explicit `ExtensionStatus(rolling_update=False)` return value there (or add a dedicated negative test) so the fail-fast path is deliberately covered. | CL, KZ, AG |
| M4 | `tests/unit/extensions/test_deployment_poller.py` | Missing negative-path tests for the contract the PR description states: (a) `Failed` + `get_extension_status` raising must raise `DeploymentFailedError`; (b) `Failed` + `rolling_update=True` persisting past the deadline must raise `DeploymentTimeoutError`; (c) `Failed` observed after `use_status_endpoint` was disabled by an earlier `Running` poll must raise immediately with no status call. | CL, KZ, AG |
| L1 | `kamiwaza_extensions/deployment_poller.py:73` | `Status: rolling update in progress` prints on every poll (every 3 s, up to ~100 lines per timeout) with no dedupe, unlike every other status line in the loop. Print once per stale window (flag or fold into `last_phase`). | CL, KZ, AG |
| L2 | `kamiwaza_extensions/deployment_poller.py:65-73` | The phase-change line prints `Status: Failed` before the stale check runs, so every post-failure redeploy shows `Failed` followed by `rolling update in progress`. Evaluate the stale check first, or print `Status: Failed (previous generation, awaiting reconcile)` in the stale case. | CL |
| L3 | `kamiwaza_extensions/deployment_poller.py:115` | Docstring "a newer PATCH generation is still rolling out" overstates the signal. Given the platform definition, "the operator has not yet completed a pass for the latest spec generation; status fields may be stale" is accurate and stops a future maintainer from "improving" the check with replica counts. | CL |
| L4 | `tests/unit/extensions/test_deployment_poller.py:81-102` | The new test patches `_check_via_status_endpoint` to `(None, "")` (endpoint absent) while `get_extension_status` returns a valid `ExtensionStatus`; no real platform behaves both ways. Leave the sibling unpatched and give `get_extension_status` a `side_effect` of `[ExtensionStatus(rolling_update=True), ExtensionStatus(services=[ready svc])]`; assert `call_count == 2`. Also yield `Failed` more than once to exercise repeated `continue`. | CL, AG |
| L5 | `kamiwaza_extensions/deployment_poller.py:37-108` | `wait_for_ready` is now 72 lines, over the repo's advisory 70-line CodeScene cap noted in `.ai/rules/codescene-authoring.md`; the hard radon gate still passes. Extracting the `Failed`-phase handling into a helper keeps it net-neutral. | KZ |
| L6 | Known limitation (doc/Linear note) | The fix only engages when the previous generation completed at least one operator pass. If the previous `Failed` came from `setFailedCondition` on a first-generation create, `observedGeneration` is omitted, the platform guard yields `rolling_update=False`, and the stale `Failed` still aborts exactly as before. Worth a comment or a follow-up ticket; the durable fix is operator-side (advance or record the failed generation in `setFailedCondition`). | CL |

## Test Coverage Assessment
`uv run pytest tests/unit/extensions/test_deployment_poller.py`: 8 passed (independently confirmed by CL, CX, KZ). Ruff clean on both changed files (CL, KZ); mypy clean (KZ). The new test genuinely exercises the `continue` branch: with `_check_via_status_endpoint` patched out, the single asserted `get_extension_status("test")` call can only come from `_is_rolling_update`. Gaps: the fail-closed exception path, the explicit `rolling_update=False` fail-fast path, the persistent-rolling timeout path, and the `use_status_endpoint=False` path are all untested (M3, M4), and one pre-existing test passes only by MagicMock accident (M3). Nothing in the SDK pins the platform semantics of `rolling_update` that this change now depends on.

## Reviewer Status and Verdicts
| Reviewer | Status | Capability | Verdict | Notes |
|---|---|---|---|---|
| Claude (CL) | success | repo tools, read-only checkout, ran tests | Request Changes (Risk: Medium) | Found M1 with operator-source evidence; rated it High. Council adjudicated M1 to Medium/non-blocking (4-0 in round 2), so the lead does not carry the Request Changes forward. |
| Codex (CX) | success | native `codex review`, repo tools, ran tests | none stated | Codex CLI upgraded 0.152.1 -> 0.153.4 mid-run (pinned model required it). Output was one sentence, "no actionable regressions, 8 tests passed", with no Risk/Verdict lines; excluded from the tally. Did not find M1. |
| Kamiwaza (KZ) | success | Kbox/Pi tool-enabled read-only checkout; GLM-5.3-NVFP4 on tokenator (vllm, local weights verified); not truncated | Approve (Risk: Low) | Found M1 independently and called it nonblocking; found M3, L1, L5. |
| Antigravity (AG) | success | interactive agy (Gemini 3.1 Pro), repo tools; trust dialog cleared manually | Comment (Risk: Low) | Found L1, M3, M4, L4. Did not find M1; as council panelist it initially rated M1 High, then reversed to Medium in round 2. |

**Vote Tally**: Approve: 1 | Comment: 1 | Request Changes: 1

## Positive Observations
- Minimal, correctly placed fix: the stale check runs before `_extract_failure_message`, so when `rolling_update` is False the behavior is byte-identical to before (CL, KZ).
- Fail-closed by design: any exception, missing endpoint, or non-True flag falls back to the pre-PR `DeploymentFailedError`; platforms without `/status` pay one extra call and fail exactly as before (CL, KZ, AG).
- The loop stays deadline-bounded after `continue`; no infinite-wait risk (KZ, CL).
- Reuses `client.extensions.get_extension_status` rather than adding API surface; no schema or service changes (KZ).
- The PR description states the fail-fast design intent accurately, which made the trade-off easy to evaluate (CL).

## Lead Cross-Cutting Audit
Independent trace by the lead, before reading lane outputs: platform `rolling_update` is exactly `generation != observedGeneration` (kamiwaza develop `extensions_service.py:322`); the operator's normal path writes `ObservedGeneration` and `Phase` in one status update, so a genuinely failed new generation at pod level still fails fast after one pass; the operator's reconcile-error path writes `Phase=Failed` without advancing `ObservedGeneration` (confirmed on both an 2026-08-07 snapshot and current develop@0eea2e8, where the only `ObservedGeneration` write is the end-of-pass one). That is the M1 mechanism; CL and KZ reached it independently. The reverse race the lead initially suspected (observedGeneration published before phase) does not exist on the normal path and is retracted. First-generation failures never engage the fix (L6). The `is True` identity check is load-bearing for an existing test via a MagicMock accident (M3). Integration seam: the poller's only caller is `kz-ext dev` (`commands/dev.py:1168`), whose timeout handler runs an operator-oriented diagnosis, which is why M1's misdirected diagnostic matters. Severity of M1 was contested (CL High vs KZ nonblocking), so per standing practice a neutral council adjudicated it rather than a vote: round 1 was 3-1 Medium, round 2 was 4-0 Medium after the F10 fact (operator requeue interval 30 s, transient 1 s) showed that a grace window needs calibration to 2-3 requeue intervals and belongs in a tested follow-up rather than a rushed inline patch. No Critical or High issue remains.

## Overall Verdict
**Risk Level**: Medium
**Recommendation**: Comment
**Rationale**: No verified Critical or High finding remains. The PR fixes a real, frequent wrong-abort on the redeploy-after-failure path with a fail-closed design and passing tests. It introduces one bounded, adjudicated-Medium regression (M1: persistent reconcile-step failures on a PATCH now cost the full 300 s timeout and a misleading diagnosis instead of an immediate but equally uninformative error) plus a cluster of nonblocking test-coverage and polish items. Merge is reasonable as-is; a prompt follow-up should land the calibrated grace window, surface the real failure message on timeout, and add the missing negative-path tests (M1-M4).

Use the engineering marketplace pr-review skill for final approval

---
Legend: CL=Claude, CX=Codex, KZ=Kamiwaza, AG=Antigravity

PR #264, opened 2026-08-14
ENG-10307: refresh direct shared-IDP SDK sessions

jxstanford:
## Summary
- add a programmatic shared-IDP password-grant authenticator with automatic refresh-token rotation
- keep refresh state in the configured token store and retry one request after an authorization failure
- restore the independently owned shared-realm exact-five validation lane on the release SDK base

## Validation
- 57 change-surface tests passed
- Ruff lint and format checks passed
- targeted mypy passed
- complexity and diff checks passed
- live release/1.2.0 validation on spark-3/spark-4 passed all five required shared-IDP cases, including gated mesh retrieval and receiver-provenance mesh job execution

## Target
`release/1.2.0`

## Tracking
[ENG-10307](https://linear.app/kamiwaza/issue/ENG-10307)

### DIFF
--- a/kamiwaza_sdk/__init__.py
+++ b/kamiwaza_sdk/__init__.py
@@ -2,6 +2,10 @@
 from importlib.metadata import version, PackageNotFoundError
 
 from .client import KamiwazaClient
+from .shared_idp_authentication import (
+    SharedIdpAuthConfig as SharedIdpAuthConfig,
+    SharedIdpAuthenticator as SharedIdpAuthenticator,
+)
 
 # Export as kamiwaza_sdk for the import pattern: from kamiwaza_sdk import KamiwazaClient as kz
 kamiwaza_sdk = KamiwazaClient
--- a/kamiwaza_sdk/client.py
+++ b/kamiwaza_sdk/client.py
@@ -258,16 +258,22 @@ class KamiwazaClient:
                 or os.environ.get("KAMIWAZA_API_TOKEN")
             )
             self.authenticator = ApiKeyAuthenticator(api_key) if api_key else None
+        self._owns_authenticator = self.authenticator is not None
 
         # Don't authenticate during initialization - let it happen on first request
 
     def close(self) -> None:
-        """Release the underlying requests.Session transport.
+        """Release the authentication and platform HTTP transports.
 
         Idempotent — repeated close() calls are safe (Session.close()
         does its own idempotency).
         """
-        self.session.close()
+        close_authenticator = getattr(self.authenticator, "close", None)
+        try:
+            if self._owns_authenticator and callable(close_authenticator):
+                close_authenticator()
+        finally:
+            self.session.close()
 
     def __enter__(self) -> "KamiwazaClient":
         return self
@@ -662,6 +668,7 @@ class KamiwazaClient:
         )
         # Preserve exact parent auth state; __init__ may otherwise consult env vars.
         scoped.authenticator = self.authenticator
+        scoped._owns_authenticator = False
         scoped.session.headers.update(self.session.headers)
         scoped.session.cookies.update(self.session.cookies)
         scoped._default_headers = dict(self._default_headers)
@@ -863,7 +870,7 @@ class KamiwazaClient:
 
     @property
     def workrooms(self):
-        if not hasattr(self, '_workrooms'):
+        if not hasattr(self, "_workrooms"):
             self._workrooms = WorkroomService(self)
         return self._workrooms
 
--- a/kamiwaza_sdk/shared_idp_authentication.py
+++ b/kamiwaza_sdk/shared_idp_authentication.py
@@ -1,0 +1,257 @@
+"""Direct shared-identity-provider authentication for automated SDK clients."""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import time
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any
+
+import requests  # type: ignore[import-untyped]
+from pydantic import ValidationError
+
+from .authentication import Authenticator
+from .exceptions import AuthenticationError
+from .schemas.auth import TokenResponse
+from .token_store import FileTokenStore, StoredToken, TokenStore
+
+_DEFAULT_TIMEOUT_SECONDS = 30.0
+_REFRESH_LEEWAY_SECONDS = 30.0
+
+
+@dataclass(frozen=True)
+class SharedIdpAuthConfig:
+    """Configuration for a browser-independent shared-realm OIDC session."""
+
+    issuer: str
+    client_id: str
+    username: str
+    password: str = field(repr=False)
+    client_secret: str | None = field(default=None, repr=False)
+    scope: str = "openid"
+    verify: bool | str = True
+    timeout_seconds: float = _DEFAULT_TIMEOUT_SECONDS
+    token_root: Path | str | None = None
+
+    @property
+    def normalized_issuer(self) -> str:
+        """Return the issuer without a trailing slash."""
+        return self.issuer.rstrip("/")
+
+
+class _OidcGrantError(AuthenticationError):
+    """Internal typed OAuth grant failure used to control safe fallback."""
+
+    def __init__(
+        self,
+        message: str,
+        *,
+        status_code: int | None,
+        oauth_error: str | None,
+    ) -> None:
+        super().__init__(message, status_code=status_code)
+        self.oauth_error = oauth_error
+
+
+def shared_idp_token_path(
+    config: SharedIdpAuthConfig,
+    *,
+    root: Path | str | None = None,
+) -> Path:
+    """Return a deterministic token path scoped to issuer, client, and user."""
+    identity = [config.normalized_issuer, config.client_id, config.username]
+    encoded = json.dumps(identity, separators=(",", ":"), ensure_ascii=True)
+    digest = hashlib.sha256(encoded.encode("utf-8")).hexdigest()
+    token_root = Path(
+        root or config.token_root or Path.home() / ".kamiwaza" / "shared-idp"
+    )
+    return token_root / f"{digest}.json"
+
+
+def _default_token_store(config: SharedIdpAuthConfig) -> FileTokenStore:
+    token_path = shared_idp_token_path(config)
+    token_path.parent.mkdir(mode=0o700, parents=True, exist_ok=True)
+    token_path.parent.chmod(0o700)
+    return FileTokenStore(token_path)
+
+
+class SharedIdpAuthenticator(Authenticator):
+    """Maintain a direct shared-realm session without browser interaction."""
+
+    def __init__(
+        self,
+        config: SharedIdpAuthConfig,
+        *,
+        token_store: TokenStore | None = None,
+        http_session: requests.Session | None = None,
+    ) -> None:
+        self.config = config
+        self.token_store = (
+            token_store if token_store is not None else _default_token_store(config)
+        )
+        self._http = http_session or requests.Session()
+        self._access_token: str | None = None
+        self._refresh_token: str | None = None
+        self._expires_at: float | None = None
+        self._load_cached_token()
+
+    def authenticate(self, session: requests.Session) -> None:
+        """Attach a valid bearer token, refreshing it proactively."""
+        if self._needs_refresh():
+            self.refresh_token(session)
+            return
+        self._attach_access_token(session)
+
+    def refresh_token(self, session: requests.Session) -> None:
+        """Refresh the shared-realm token or perform a programmatic login."""
+        refresh_token = self._refresh_token
+        if not refresh_token:
+            response = self._password_grant()
+            self._store_response(response, previous_refresh=None)
+            self._attach_access_token(session)
+            return
+
+        previous_refresh: str | None = refresh_token
+        try:
+            response = self._refresh_grant(refresh_token)
+        except _OidcGrantError as exc:
+            if exc.oauth_error != "invalid_grant":
+                raise
+            self._clear_tokens()
+            response = self._password_grant()
+            previous_refresh = None
+        self._store_response(response, previous_refresh=previous_refresh)
+        self._attach_access_token(session)
+
+    def get_access_token(self, session: requests.Session) -> str | None:
+        """Return a current bearer token for callers that need token access."""
+        self.authenticate(session)
+        return self._access_token
+
+    def invalidate_session(self, session: requests.Session) -> bool:
+        """Expire the access token while retaining refresh capability."""
+        self._access_token = None
+        self._expires_at = None
+        session.headers.pop("Authorization", None)
+        return True
+
+    def _needs_refresh(self) -> bool:
+        if not self._access_token or self._expires_at is None:
+            return True
+        return time.time() >= self._expires_at - _REFRESH_LEEWAY_SECONDS
+
+    def _password_grant(self) -> TokenResponse:
+        data = {
+            "grant_type": "password",
+            "client_id": self.config.client_id,
+            "username": self.config.username,
+            "password": self.config.password,
+            "scope": self.config.scope,
+        }
+        return self._request_token(self._with_client_secret(data))
+
+    def _refresh_grant(self, refresh_token: str) -> TokenResponse:
+        data = {
+            "grant_type": "refresh_token",
+            "client_id": self.config.client_id,
+            "refresh_token": refresh_token,
+        }
+        return self._request_token(self._with_client_secret(data))
+
+    def _with_client_secret(self, data: dict[str, str]) -> dict[str, str]:
+        if not self.config.client_secret:
+            return data
+        return {**data, "client_secret": self.config.client_secret}
+
+    def _request_token(self, data: dict[str, str]) -> TokenResponse:
+        endpoint = f"{self.config.normalized_issuer}/protocol/openid-connect/token"
+        try:
+            response = self._http.post(
+                endpoint,
+                data=data,
+                timeout=self.config.timeout_seconds,
+                verify=self.config.verify,
+            )
+        except requests.RequestException as exc:
+            raise AuthenticationError(
+                "Shared identity provider is unavailable"
+            ) from exc
+
+        payload = self._response_payload(response)
+        if response.status_code >= 400:
+            raise self._grant_error(response.status_code, payload)
+        if payload is None:
+            raise AuthenticationError("Shared identity provider returned invalid JSON")
+        try:
+            return TokenResponse.model_validate(payload)
+        except ValidationError as exc:
+            raise AuthenticationError(
+                "Shared identity provider returned an invalid token response"
+            ) from exc
+
+    @staticmethod
+    def _response_payload(response: requests.Response) -> dict[str, Any] | None:
+        try:
+            payload = response.json()
+        except ValueError:
+            return None
+        return payload if isinstance(payload, dict) else None
+
+    @staticmethod
+    def _grant_error(
+        status_code: int,
+        payload: dict[str, Any] | None,
+    ) -> _OidcGrantError:
+        oauth_error = payload.get("error") if payload else None
+        description = payload.get("error_description") if payload else None
+        detail = description or oauth_error or "token grant failed"
+        return _OidcGrantError(
+            f"Shared identity provider rejected the token grant: {detail}",
+            status_code=status_code,
+            oauth_error=oauth_error if isinstance(oauth_error, str) else None,
+        )
+
+    def _store_response(
+        self,
+        response: TokenResponse,
+        *,
+        previous_refresh: str | None,
+    ) -> None:
+        refresh_token = response.refresh_token or previous_refresh
+        if not refresh_token:
+            raise AuthenticationError(
+                "Shared identity provider did not return a refresh token"
+            )
+        expires_at = time.time() + response.expires_in
+        stored = StoredToken(
+            access_token=response.access_token,
+            refresh_token=refresh_token,
+            expires_at=expires_at,
+        )
+        self.token_store.save(stored)
+        self._access_token = stored.access_token
+        self._refresh_token = stored.refresh_token
+        self._expires_at = stored.expires_at
+
+    def _attach_access_token(self, session: requests.Session) -> None:
+        if not self._access_token:
+            raise AuthenticationError(
+                "Shared identity provider access token is unavailable"
+            )
+        session.headers["Authorization"] = f"Bearer {self._access_token}"
+
+    def _clear_tokens(self) -> None:
+        self._access_token = None
+        self._refresh_token = None
+        self._expires_at = None
+        self.token_store.clear()
+
+    def _load_cached_token(self) -> None:
+        cached = self.token_store.load()
+        if not cached:
+            return
+        self._access_token = cached.access_token
+        self._refresh_token = cached.refresh_token
+        self._expires_at = cached.expires_at
--- a/kamiwaza_sdk/shared_idp_authentication.py
+++ b/kamiwaza_sdk/shared_idp_authentication.py
@@ -5,11 +5,14 @@ from __future__ import annotations
 import hashlib
 import json
 import time
+from contextlib import AbstractContextManager, nullcontext
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
+from urllib.parse import urlsplit
 
 import requests  # type: ignore[import-untyped]
+from filelock import FileLock
 from pydantic import ValidationError
 
 from .authentication import Authenticator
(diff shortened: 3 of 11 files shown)


m9e (changes requested):
# PR Review Report: ENG-10307 — refresh direct shared-IDP SDK sessions

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/264
**Branch**: `fix/eng-10307-shared-idp-refresh-release-1.2.0` → `release/1.2.0`
**Head reviewed**: `0f4d78ad` · **Merge-base**: `0ccc0c43` · **Scope**: 11 files, +1254/−136
**CI at head**: 11/11 check-runs green (incl. Unit Tests, Mypy, Ruff, Coverage, Cyclomatic Complexity, CodeScene, live Extension contract).

Five independent engines ran (Superpowers, Claude-deep, Codex ×2 passes, Kamiwaza `DeepSeek-V4-Flash-0731`, Antigravity). Every finding below is backed by an executed probe or a two-arm mutation, and the SHA each measurement came from is stated.

---

## Critical Issues

### C1 — A stale refresh token is replayed after `invalidate_session()`, because the cross-process adoption guard is skipped on exactly that path
`kamiwaza_sdk/shared_idp_authentication.py:178-192` (`_adopt_updated_cached_token`) + `:208-213` (`invalidate_session`)
**Flagged by: CL-me(executed), CX-A(P1), AG(Critical), KZ(High)** — 4 lanes, independently.

`_adopt_updated_cached_token` is this PR's protection against replaying a refresh token that a peer process already rotated. It bails out immediately on `known_token is None`. `invalidate_session()` produces precisely that state — it nulls `_access_token`/`_expires_at` but **retains `_refresh_token`** — so the guard is disarmed on the one path where the SDK is already in trouble. `WorkroomService._request_with_scope_recovery` (`services/workrooms.py:51-56`) drives that path on every stale-workroom-scope 403.

Two-arm proof at `0f4d78a` (fake Keycloak, rotation + reuse detection = the `Revoke Refresh Token` posture a shared federation realm would run):

```
setup:   A and B both load the SHARED FileTokenStore -> access A1 / refresh R1
         A's token expires, A rotates -> cache now holds R2; B still holds R1 in memory

ARM 1  B reaches refresh via invalidate_session()  (known_token == None)
       B RESULT: RAISED _OidcGrantError: rejected the token grant: Invalid refresh token (reuse detected)
       B grants: ['refresh_token']   IDP family revoked: True
       A afterwards: COLLATERAL FAILURE _OidcGrantError  <-- the process holding the GOOD session also dies

CONTROL  B reaches refresh via ordinary expiry  (known_token present)
       CONTROL: ok, token A2   grants: []   revoked: False   <-- adopts the rotated cache, spends nothing
```

**Floor, without reuse detection** (re-ran with the reuse check disabled): `B RESULT: RAISED ... Invalid refresh token`, no collateral. So even on a permissive Keycloak this is a hard auth failure on a path that had a perfectly valid token sitting in the cache one `load()` away. `password_regrant_on_invalid_grant` defaults to `False`, so the fallback does not rescue it.

The two `test_shared_file_cache_reloads_*` tests are the control arm made explicit: they prove the author designed for this hazard and tested it — for `known_token is not None` only.

**A naive fix does not work, and the tests say why.** Dropping the `known_token is None` clause outright fails `test_invalidation_retains_refresh_and_get_access_token_renews` (SP's mutant M9 KILLED). That test is right: after an invalidation the cache still holds the *same* access token that was just rejected, and re-adopting it would make the retry a no-op. Verified remedy that satisfies both constraints — remember *which* token was invalidated, so the comparison can distinguish "cache still holds the rejected token" from "a peer rotated the cache":

```python
# __init__
self._invalidated_token: StoredToken | None = None
# invalidate_session(), before nulling
self._invalidated_token = self._current_token()
# _apply_cached_token(), after applying
self._invalidated_token = None
# _refresh_token_locked()
known_token = self._current_token() or self._invalidated_token
```
Executed at `0f4d78a`: **33/33 unit tests still pass** (`test_shared_idp_authentication.py` + `test_client_workroom_scope.py`), ARM 1 becomes `B RESULT: ok, token A2 · grants: [] · revoked: False`, and the control arm is unchanged.

### C2 — N concurrent callers issue N token grants; with a non-`FileTokenStore` there is no mutual exclusion at all
`kamiwaza_sdk/shared_idp_authentication.py:130-135` (`authenticate`) + `:169-176` (`_refresh_guard`)
**Flagged by: SP(Critical), CL(Critical), AG(High), CL-me(executed)** — 4 lanes.

`authenticate()` evaluates `_needs_refresh()` **outside** the lock and `_refresh_token_locked` never re-evaluates it after acquiring the lock. And `_refresh_guard()` returns `nullcontext()` for every store that is not a `FileTokenStore` — there is no `threading.Lock` anywhere.

8 threads, one authenticator, one expired token, at `0f4d78a`:
```
    FileTokenStore: 8 concurrent threads -> 8 token grants ['password','refresh_token' x7]
InMemoryTokenStore: 8 concurrent threads -> 8 token grants ['password' x8]
```
Two lanes reproduced the sharper variant independently — with `InMemoryTokenStore` the *same* refresh token is presented repeatedly:
- SP (4 threads): `store=memory grants=4 refresh_tokens_sent=['refresh-1','refresh-1','refresh-1','refresh-1'] DUPLICATES: 3`
- CL (3 threads): `REFRESH TOKEN REUSED across concurrent grants: True | n_grants: 3`

Under reuse detection the second presentation kills the whole session chain — the exact failure this PR exists to prevent. **`InMemoryTokenStore` is what this PR's own integration fixture uses** (`_programmatic_persona_session`, live file :337). Even where the `flock` does serialize, N sequential rotations happen where one would do, and each rotation retires the token any peer process cached.

Fix: an unconditional `threading.RLock`, plus a re-check of `_needs_refresh()` *inside* the guard. The re-check must not swallow the client's forced post-401 refresh (`client.py:409` calls `refresh_token()` deliberately on a non-expired token) — so this needs an explicit `force=` split between "proactive" and "forced" refresh.

### C3 — The required-edge release gate is now tautological: it validates `REQUIRED_EDGE_CASES` against itself
`tests/integration/test_federation_required_edge_contract.py:56-66`, `tests/integration/required_federation_edge.py:8-17`
**Flagged by: CL(Critical, 2-arm), CX-A(P3), SP(High), CL-me(executed)** — 4 lanes.

`test_required_edge_collection_guard_requires_all_five_cases` builds its `items` **from** `required_edge.REQUIRED_EDGE_CASES` and then asserts that set is satisfied. At merge-base it also carried two literal anchors, both deleted here:

```python
assert "test_unonboarded_primary_realm_user_rejected_by_receiver_allowlist" in required_edge.REQUIRED_EDGE_CASES
assert not any("native_realm_token" in case for case in required_edge.REQUIRED_EDGE_CASES)
```

The second is an explicit prohibition on the exact substitution this PR makes. It was **deleted rather than inverted**, and nothing at head pins the set: `grep -rn REQUIRED_EDGE_CASES tests/ conftest.py` finds only the two self-referential comprehensions. Despite the test's name, `len(REQUIRED_EDGE_CASES) == 5` has never been asserted.

Two-arm mutation (my run — replace the whole five-case set with two bogus names):
```
HEAD 0f4d78a        -> 30 passed          (SURVIVED)
merge-base 0ccc0c43 -> 6 failed, 23 passed (CAUGHT)
```
CL reproduced it with a narrower mutant (rename the 5th entry to a function that exists nowhere): base `1 failed, 28 passed` → head `30 passed`.

The PR body says this "restores the independently owned shared-realm exact-five validation lane". The exact-five *pin* is the thing it removed — a required-case name can now point at a nonexistent function and the whole contract suite stays green.

---

## High Priority Issues

### H1 — `_owns_authenticator` is "was an authenticator present at `__init__`", not ownership — so `close()` tears down a caller-supplied authenticator
`kamiwaza_sdk/client.py:261, 271-276` · **Flagged by all 5 lanes** (CL-me executed, CX-B P3, AG High, SP High, CL High, KZ suggestion).

`self._owns_authenticator = self.authenticator is not None` is at method-body level, so it is `True` for an injected authenticator too. Executed at `0f4d78a`:
```
client._owns_authenticator for a CALLER-SUPPLIED authenticator: True
client.close() invoked caller-supplied authenticator.close(): True
```
CL reproduced the sharing case: `c1 owns: True  c2 owns: True` for two clients holding the *same* authenticator; `c1.close()` closes the OIDC transport out from under `c2`. That directly contradicts the multi-client sharing pattern the token store exists to enable. The `workroom_scope` carve-out (`client.py:671` `scoped._owns_authenticator = False`) shows the intent — it patches one instance of the problem.

Codex found the same defect from the opposite direction: a client built first and *then* given `client.authenticator = ...` has the flag stuck at `False` and never closes it. There is a real call site — `kamiwaza_sdk/seeding/federation/cli.py:465`. (Currently benign: only `SharedIdpAuthenticator` defines `close()`, so no transport leaks there today. Latent.)

Fix: set the flag only on the internally-constructed branch, or take an explicit `owns_authenticator: bool` kwarg. Note `test_client_close_releases_internally_owned_oidc_session` calls this "internally owned" — it isn't.

### H2 — The integration-test rewrite dropped persona-provisioning coverage entirely
`tests/integration/test_federation_shared_idp_gated_retrieval_live.py:366-405` · **Flagged by: CL (2-arm mutation)**

Deleting `test_required_personas_use_primary_realm_login_and_receiver_onboarding` removed the only tests of `_provision_personas`:

| mutant | base `0ccc0c4` | head `0f4d78a` |
|---|---|---|
| delete the `receiver._request("POST", .../users)` onboarding call | CAUGHT (1 failed) | **SURVIVED (30 passed)** |
| drop the `TS` persona from the persona map | CAUGHT (1 failed) | **SURVIVED (30 passed)** |

Base asserted `receiver._request.call_count == 3` and `set(personas) == {"U","S","TS","unonboarded"}`. Head asserts neither; the replacement test covers a single persona session, not the loop.

### H3 — The required negative case changed layer (authorization → authentication), leaving no negative coverage of the receiver allowlist
`required_federation_edge.py:14`, live file `:533-545` · **Flagged by: SP, CL, CX-B(P2)**

`test_unonboarded_primary_realm_user_rejected_by_receiver_allowlist` (asserted `unauthorized_brokered_user`, a receiver *authorization* check) was swapped for `test_native_realm_token_rejected_at_receiver_shared_idp_boundary` (asserts `peer_jwt_validation_failed`, a receiver *authentication* check). `_provision_personas` still writes allowlist rows for U/S/TS, but the required five now contain **no negative test that the allowlist denies anything**.

Compounding (CX-B, CL): the new case drives `live_kamiwaza_session_client`, and that fixture prefers `KAMIWAZA_API_KEY` over a password login (`conftest.py:1190-1201`). When the lane runs with an API key set — the common case — the credential under test is a PAT, not the "initiator-native token" the name and docstring claim, and the rejection reason becomes environment-dependent. The old case used a dedicated `unonboarded` persona client and was deterministic.

### H4 — Three load-bearing guards survive mutation with the suite fully green
`kamiwaza_sdk/shared_idp_authentication.py`, `kamiwaza_sdk/token_store.py` · **Flagged by: SP and CL independently, matching arms**

| mutant | result |
|---|---|
| `_refresh_guard` → always `nullcontext()` (delete the `FileLock` outright) | **SURVIVED** (SP 28P / CL 63P) |
| `token_store.save`: `os.open` mode `0o600` → `0o644` | **SURVIVED** (SP 28P) |
| `_grant_error`: always `oauth_error="invalid_grant"` | **SURVIVED** (SP 28P) |
| `_needs_refresh`: drop `- _REFRESH_LEEWAY_SECONDS` | **SURVIVED** (SP 28P / CL 63P) |
| `_store_response`: `time.time() + 3600` instead of `+ response.expires_in` | **SURVIVED** (CL 63P) |
| `_clear_tokens`: drop `token_store.clear()` | **SURVIVED** (CL 63P) |

The `filelock>=3.20` runtime dependency this PR adds has **no test at all** — the two `test_shared_file_cache_reloads_*` tests run sequentially and pass with the lock deleted. On `0o600`: the only assertion checks the *end state*, which the trailing `self.path.chmod(0o600)` satisfies alone, so the layer that actually closes the world-readable window during the write is the untested one (add a `os.umask(0)`-hostile test). On `invalid_grant`: `test_refresh_transport_or_server_failure_does_not_password_regrant` is parametrized only with `password_regrant_on_invalid_grant=False`, so it never exercises the discrimination in its own name — under the mutant, a `503` or `invalid_client` would replay the plaintext password to a degraded endpoint.

### H5 — `verify=False` over HTTPS is ungated and unwarned, while plain `http://` is gated
`kamiwaza_sdk/shared_idp_authentication.py:37, 43-53` · **Flagged by: SP, CL**

`__post_init__` refuses `http://` without `allow_insecure_http=True` ("Reject plaintext credential transport"), but `verify: bool | str = True` accepts `False` silently — which hands the ROPC password and client secret to any MITM just as effectively. `KamiwazaClient` at least logs "SSL verification disabled" for its own transport; this config does not. CL adds that in the live lane `idp_verify` falls back to `platform_verify`, and `tests/integration/conftest.py:1188` does `os.environ.setdefault("KAMIWAZA_VERIFY_SSL","false")` — so the lane's default posture is to POST the persona password to the IDP with verification off.

### H6 — `tests/integration/required_federation_edge_personas.py` is now dead code
**Flagged by: SP, CL; verified independently** — the file is still tracked at head (4,949 B) and `grep -rn required_federation_edge_personas --include=*.py .` returns zero importers (at merge-base: two). It still contains `_require_primary_personas_absent`, a fail-closed refusal to overwrite an operator-owned primary-realm identity that has **no equivalent** in the new inline `_provision_personas`. Delete it, or fold that guard back in.

---

## Medium Priority

| 1 | `shared_idp_authentication.py:176` | `FileLock(...)` uses filelock's default `timeout=-1` — block forever. Verified: `filelock 3.32.3`, `timeout: float = -1`, guard resolves to `UnixFileLock … timeout: -1`. One wedged holder stalls every other SDK process on that identity, though the token POST itself is bounded at 30 s. Pass `~2 × timeout_seconds` and translate `filelock.Timeout` into `AuthenticationError`. | me, KZ, CX-A, SP, CL |
| 2 | `shared_idp_authentication.py:313` | No floor on `expires_in`. Executed: `expires_in=0` → **4 token grants for 4 `authenticate()` calls**; `expires_in=-60` likewise. Any `expires_in <= 30` (the leeway) means a full grant per request. | me, KZ, CL |
| 3 | `shared_idp_authentication.py:41, 155-166` | With the default `password_regrant_on_invalid_grant=False`, a dead refresh token is re-raised **without** `_clear_tokens()`, so it stays in the `FileTokenStore` on disk. Every future process reloads the poison and fails identically until someone `rm`s `~/.kamiwaza/shared-idp/<sha256>.json`. For a class documented "for automated SDK clients" that is the wrong default; at minimum clear the dead token so a restart recovers. | CL |
| 4 | `shared_idp_authentication.py:249` | Token endpoint hardcoded to Keycloak's `{issuer}/protocol/openid-connect/token` — no discovery, no override — under a provider-neutral public name (`SharedIdpAuthenticator`, generic `issuer` field) exported from `__init__.py`. Okta/Auth0/Entra will 404. Add `token_endpoint: str \| None`, or say "Keycloak" in the docstring. | AG, CL, SP |
| 5 | `shared_idp_authentication.py:319-322` | `token_store.save(stored)` runs *before* the in-memory fields update. If the save raises (full disk, read-only home) the newly rotated refresh token is dropped while the old one is already dead at the IDP — unrecoverable unless regrant is enabled. | SP |
| 6 | `shared_idp_authentication.py:302-312` | An empty `access_token` is **persisted to disk** before `_attach_access_token` rejects it; `TokenResponse.access_token` has no `min_length=1`. Validate before `save`. | CL |
| 7 | `token_store.py:69` | `tmp_path = self.path.with_suffix(".tmp")` is a fixed name shared by every writer of that path; a concurrent writer can tear the write. Use `tempfile.mkstemp(dir=self.path.parent)`. Also: no `try/finally` cleanup, so a failed write orphans the `.tmp`. | SP, AG |
| 8 | contract test `:209-258` | `test_persona_session_performs_password_grant_then_real_refresh` monkeypatches **both** `edge.SharedIdpAuthenticator` and `edge.KamiwazaClient`. The stub saves a `SimpleNamespace` into a real `InMemoryTokenStore`, so the `token_store.load().refresh_token` assertions are satisfied by an object that is not a `StoredToken`; and `assert persona["client"].verify is True` passes only because the stub defines `.verify` — the real `KamiwazaClient` has no such attribute (it uses `session.verify`), so that assertion could never hold against production. | CL |
| 9 | live file `:384` | `sub = mc.jwt_sub(token) or base` silently substitutes the username when the `sub` claim is missing (`jwt_sub` returns `""`). The receiver keys the brokered user on `<sub>@<cluster-uuid>`, so this registers a wrong allowlist row; later assertions fail for an unrelated-looking reason and teardown revokes the garbage row while the real one leaks. Merge-base asserted `sub and sub == str(current.sub)`. | CL |
| 10 | live file `:325-348` | Each persona builds a `KamiwazaClient` that is never `close()`d and never registered with the `ExitStack` — 3 platform + 3 OIDC sessions leak for the module's lifetime. | CL |
| 11 | `__init__.py:5-8` | The eager re-export makes `filelock` a hard import-time dependency of `import kamiwaza_sdk` for every user. Mitigating: `filelock` was already transitive via `huggingface-hub`, is Unlicense, and is locked with sdist+wheel hashes — no new supply-chain surface. Consider PEP-562 lazy re-export, and whether `>=3.20` (Sept 2025) is a needlessly aggressive floor when `mode=` has existed since 3.10. | SP, CL |
| 12 | `shared_idp_authentication.py:80-97` | `shared_idp_token_path` hashes issuer/client_id/username/scope but **not** `client_secret` or `password` — two configs differing only by secret share a cache entry. | CL, KZ |
| 13 | — | No README/`docs/` entry for the two new public exports; `grep -rn "SharedIdpAuth" README.md docs/ AGENTS.md` → 0 hits. | CL |

---

## Test Coverage Assessment

**Coverage verdict: Needs Work.** Baselines at `0f4d78a`: 33 unit (`test_shared_idp_authentication.py` 28 + `test_client_workroom_scope.py` 5), 30 contract — 63 total. The PR body's "57 change-surface tests passed" is conservative and its ruff/mypy claims verified true.

**Genuinely well covered** (each mutation-verified CAUGHT): retry-once-on-401 (asserts the header sequence *and* the grant sequence — notably strong); refresh-token reuse when the provider does not rotate; cross-*process* file-cache adoption on the `known_token is not None` path; redirect refusal before credential replay; scope normalization in the cache key; HTTPS-issuer enforcement; the full IDP error matrix; token-store file/dir permissions and custom-root preservation; secret non-disclosure in `repr`.

**Missing** (mutation SURVIVED — the behavior can be deleted with everything green): the `FileLock` mechanism entirely; any concurrency test at all; the 30 s refresh leeway; `_clear_tokens`' store clear; expiry computation from `expires_in`; `_provision_personas`; the `invalid_grant` discrimination under `password_regrant_on_invalid_grant=True`; the private-mode `os.open`; a second consecutive 401 after refresh; `get_access_token` returning `None`; and — per C1 — the `invalidate_session` + peer-rotation interaction.

---

## Assertion-Delta Audit — the "restore" is a net weakening

Checks present at `0ccc0c43` and **absent** at `0f4d78ad`:
1. The literal pin on the required case name in `REQUIRED_EDGE_CASES`. *No replacement.*
2. `assert not any("native_realm_token" in case ...)` — an explicit prohibition on this PR's substitution. **Deleted rather than inverted.**
3. `assert set(personas) == {"U","S","TS","unonboarded"}`.
4. `assert initiator.subjects.upsert.call_count == 4`.
5. `assert set(persona_clients) == {"fed-clr-u","fed-clr-s","fed-clr-ts","fed-clr-unonboarded"}`.
6. `assert receiver._request.call_count == 3` — exact onboarding-row count.
7. `direct_shared_login.assert_not_called()` — **legitimately** obsolete; enabling direct ROPC is the PR's purpose.
8. `assert personas[clearance]["client"] is persona_clients[username]` — persona→client identity binding.
9. Login targets the initiator `base_url`, threads the password, honors `verify`.
10. Exercise of `_require_primary_personas_absent` (the fail-closed refusal to overwrite an operator-owned identity). *No equivalent at head.*
11. Layer swap: `unauthorized_brokered_user` (authorization) → `peer_jwt_validation_failed` (authentication). Not preserved, see H3.

Genuinely added: `requires_owned_shared_realm` in the marker superset and `_required_item` keywords; the CA-bundle round-trip test (content, `0o600`, no-CA fallback); and — a real strengthening — retrieval/job cases now assert the token comes from a live `authenticator.get_access_token(client.session)` rather than a static dict entry (that mutant is CAUGHT).

**Verdict: WEAKENED.** Three independent 2-arm mutations flip from CAUGHT at `0ccc0c43` to SURVIVED at `0f4d78ad`. Item 7 is the only removal that is clearly correct.

---

## Reviewer Verdicts

| Reviewer | Verdict | Notes |
|----------|---------|-------|
| Superpowers | **Request Changes** | Risk High. 10-mutant battery, 6/10 killed; concurrency proof executed. |
| Claude (deep) | **Request Changes** | Confidence High. 2-arm base-vs-head mutation grid on every weakening claim. |
| Codex (2 passes) | **Request Changes** | Pass A P1 = C1, P2 = lock timeout, P3 = C3. Pass B = H1 (inverse) + H3. Zero overlap between passes, as usual. |
| Kamiwaza (`DeepSeek-V4-Flash-0731`) | **Comment** | Risk Medium, no Criticals. `truncated:false`, model echoed. Independently named C1, the lock timeout and the `expires_in` floor. |
| Antigravity | **Request Changes** | Risk High. Independently named C1 (Critical) and H1 (High). |

**Vote Tally**: Approve 0 | Comment 1 | Request Changes 4 — out of N=5.

---

## Positive Observations

- The per-path security posture of the new module is unusually good and mostly mutation-verified: `allow_redirects=False` **plus** an explicit 3xx rejection *before* any payload parse, so credentials are never replayed to an attacker-controlled URL; non-JSON and schema-invalid responses become typed `AuthenticationError`s; `password_regrant_on_invalid_grant` defaults to fail-closed.
- `_OidcGrantError` carrying the typed `oauth_error` so that only `invalid_grant` — never a transport failure or a 5xx — can trigger a password regrant is a careful, correct design.
- `SharedIdpAuthConfig` marks `password` and `client_secret` `repr=False` **and** has a test asserting neither appears in `repr()`. Rarely done.
- `shared_idp_token_path` scopes the cache by issuer + client_id + username + *normalized, order-independent* scope, with tests proving all four dimensions separate and that `"offline_access openid" == "openid   offline_access"`. That is the right key.
- `_adopt_updated_cached_token`'s read-under-lock is the right *shape* for cross-process rotation — C1 is a hole in one entry path, not a wrong design.
- `test_client_refreshes_and_retries_one_unauthorized_request` asserts the header sequence alongside the grant sequence; the retry is properly bounded by `did_refresh` — no 401 loop, no recursion, no lockout amplification.
- `test_existing_custom_token_root_permissions_are_not_changed` is a thoughtful, tested distinction between SDK-managed and caller-managed cache roots.
- The `filelock` choice is sound: Unlicense, already transitive via `huggingface-hub`, locked with sdist + wheel hashes.
- CI is fully green at head and the PR body's lint/type/test claims all check out.

---

## Overall Verdict

**Risk Level: High**
**Reviewer Consensus**: 4/5

The authenticator itself is carefully built, and most of its individual guards are real and tested. What blocks it is that the two mechanisms the PR exists to provide are each broken on one path — the cross-process rotation guard is skipped exactly when `invalidate_session()` has run (C1), and the refresh serialization does not serialize for the store the PR's own fixture uses, nor re-check under the lock (C2) — and neither is covered by a test, since deleting the `FileLock` outright leaves all 63 tests green. Separately, the integration lane described as "restoring the exact-five validation" measurably removed the pin that made it a gate (C3): a required case can now name a function that does not exist and the contract suite stays green.

C1 has a verified fix that keeps 33/33 green. C3 is a few lines of re-added assertions. C2 needs the proactive-vs-forced refresh split. Happy to re-review as soon as those land.

---

Legend: SP=Superpowers, CL=Claude, CX=Codex, KZ=Kamiwaza, AG=Antigravity. Reviewed at head `0f4d78ad`, merge-base `0ccc0c43`.

jxstanford:
# Review Report: ENG-10307: refresh direct shared-IDP SDK sessions

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/264  
**Exact head**: `0f4d78ad5c6cec0b330864589c483e84bafb73bb`  
**Engines**: Claude, Codex, Kamiwaza (3/3)

## Critical Issues

None.

## High Issues

None.

## Medium Issues

None.

## Test Coverage Assessment

- The focused authentication and scoped-client suite passed: 33 tests.
- The final concurrency regression passed and explicitly proves the second authenticator holds stale expired in-memory state before another process updates the shared cache.
- Ruff, formatting, targeted mypy, diff checks, and all required CI checks passed on the exact head.
- Coverage includes rotating and non-rotating refresh providers, cross-process cache serialization, explicit 401 refresh, redirect refusal, token-store permissions, TLS/CA handling, and authenticator ownership.

## Positive Observations

- Refresh serialization holds a deterministic file lock across cache reload, token grant, and atomic persistence.
- A waiting client adopts a newer valid access token even when the provider reuses the refresh token, avoiding redundant grants.
- Explicit invalidation cannot silently re-adopt the rejected bearer; it forces a real refresh.
- Password regrant after `invalid_grant` is opt-in and disabled by default.
- OIDC credential requests reject redirects and preserve the HTTPS-by-default transport boundary.

## CI & Merge Status

| Check | Status |
|---|---|
| Mergeable | MERGEABLE |
| Base | `release/1.2.0` |
| Required checks | All passing |

## Overall Verdict

**APPROVE** — All three engines approved the exact current head. No Critical, High, or Medium issue remains.

**Reviewed-SHA**: 0f4d78ad5c6cec0b330864589c483e84bafb73bb

PR #269, opened 2026-08-14
ENG-10385: make JSON Schema validation mandatory

nickz9:
## Summary

- declare `jsonschema>=4.26.0,<5.0.0` in the SDK dev/test dependency group and refresh `uv.lock`
- import `Draft202012Validator` normally so the Evidence v2 contract test cannot silently skip
- preserve the unrelated BSD/macOS `sed` skip as the explicit platform waiver

## Validation

- `uv sync --group dev --locked`
- `uv run pytest -q tests/e2e/scenarios/test_evidence_v2.py::TestSchemaFileSync::test_v2_record_validates_against_schema_file` — 1 passed
- `make test` — 3,272 passed, 1 skipped (GNU `sed` waiver), 260 deselected
- `uv run ruff check tests/e2e/scenarios/test_evidence_v2.py` — passed
- `uv run mypy --explicit-package-bases --disable-error-code import-untyped tests/e2e/scenarios/test_evidence_v2.py` — passed
- `git diff --check` and `uv lock --check` — passed

Repository-wide `make lint` and `make type-check` were also run. They report 28 Ruff errors and 37 mypy errors respectively in pre-existing, untouched SDK source files; this PR adds no source-file failure and the focused checks above are green.

Linear: ENG-10385

No Kajiya changes. No merge requested.

PR #271, opened 2026-08-14
[ENG-10397] Fix CLI serve-deploy live PAT scope and cleanup

nickz9:
## Summary

- keep the auth-only CLI PAT flow at `openid` scope while using the required explicit `admin` scope for `serve deploy`
- surface sanitized CLI stderr instead of an opaque `CalledProcessError`, redacting passwords and JWTs
- stop every deployment created by the live node and revoke both test PATs, including failure paths
- add focused contracts for scope selection, diagnostic redaction, and exact PAT cleanup identity

## Validation

- `make test`: 3,275 passed, 2 skipped, 260 deselected
  - jsonschema gate skip is addressed by pending SDK PR #269
  - GNU sed skip is the established BSD/macOS waiver
- focused unit tests: 9 passed
- Ruff: passed on all changed files
- mypy: passed on the new support module and unit tests
- package sdist/wheel build: passed
- Helm-only k0s/Lima/Istio live CLI module: 2 passed in 196.92s
- test PATs revoked and no new model Deployment, Pod, Service, or route remained

Linear: ENG-10397

No Kajiya, Plan MCP, Core, deploy, or federation changes.


nickz9:
# Review Report: [ENG-10397] Fix CLI serve-deploy live PAT scope and cleanup

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/271
**Branch**: `feature/eng-10397-sdk-cli-serve-deploy-live-test-uses-non-admin-pat-and-hides` → `release/1.1.0`
**Engines**: Claude Code, OpenAI Codex (2/3 available — Kamiwaza engine not configured)

## CI & Merge Status
| Item | Status |
|------|--------|
| Checks | 7 pass · 0 fail · 0 pending |
| Mergeable | `MERGEABLE` / `BLOCKED` |
| Review decision | `REVIEW_REQUIRED` |
| CodeScene | APPROVED |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 8/10 | Scope escalation correctly grounded in the repo's documented contract; the cleanup design is the weak point. |
| Security | 6/10 | Force-stops unowned deployments with an admin PAT; admin PAT leaks on the common skip path. |
| Test Quality | 7/10 | Redaction guard is mutation-proven, but `_cleanup_cli_resources` — the headline fix — has zero unit coverage. |
| Error Handling | 6/10 | Raw `{exc}` interpolated with no redaction or truncation; JWT regex over-matches and destroys diagnostics. |
| Code Quality | 8/10 | Clean, ruff/mypy clean, small functions; 8-param helper exceeds the repo's own cap. |
| Refactoring | 8/10 | Extracting a testable seam is the right move and is what makes the scope contract verifiable. |
| Documentation Accuracy | N/A | No Markdown changed — all 3 files are `.py`. |

## Critical Issues (Must Fix)
| 1 | `tests/integration/test_cli_live.py:42-54` | **Cleanup force-stops deployments the test does not own.** `cleanup_ids = current_deployment_ids - baseline_deployment_ids`, then every id in that set gets `stop_deployment(..., force=True)` — with **no ownership check, no creator check, and no test-specific name filter**. Any deployment created between the baseline snapshot and teardown is terminated, including one belonging to a concurrent CI run or another user on a shared live cluster. This compounds with the (otherwise correct) scope escalation in this same PR: the admin PAT gives the cleanup cluster-wide authority to actually do it. Verified by reading the code. Restrict teardown to the CLI-reported `deployment_id`, or filter by an identifier uniquely tied to this run. | Codex | security / destructive action |

## High Priority Issues
| 1 | `tests/integration/test_cli_live.py:118-137` | **The admin PAT leaks on the common path.** It is minted at `:118`, but `try:` does not open until `:137` — `client_factory`, `ensure_deployable_model_ready`, `target_model_file_id` and `list_deployments()` all run in between. `_ensure_deployable_target_ready` both raises *and* calls `pytest.skip()`, and `Skipped` derives from `BaseException`; either way the code sits **before** the `try`, so `_cleanup_cli_resources` never runs. That skip is the *expected* path on any host the `requires_deployable_model` gate excludes — so on most hosts this test mints an admin PAT, reports SKIPPED, and leaves it live for its full 900s TTL and cached on disk. Reproduced by the engine. Directly contradicts the PR's claim to "revoke both test PATs, including failure paths." Move `try:` to immediately after `:126`. | Claude | security |
| 2 | `tests/integration/test_cli_live.py:34-70` | `_cleanup_cli_resources` — the headline fix — has **zero** unit coverage. It and `_serve_deploy_args` stayed in the integration file rather than moving into `_kamiwaza_cli_live.py`, so unlike `run_cli`/`pat_jti` they have no seam and can only run against a live cluster. Mutating it is vacuous off-cluster: nothing fails. The baseline-reconciliation and failure-aggregation logic is entirely unverified — including the Critical above. | Claude | test coverage |
| 3 | `tests/integration/test_cli_live.py:49,61,67` | Cleanup failure strings interpolate raw `{exc}` with no redaction and no truncation, then surface it via the `assert` message. `APIError.__str__` embeds full response bodies (`client.py:539`). This bypasses `_redact_cli_stderr` entirely and re-introduces the log-flooding problem that `conftest.py`'s `_PROBE_ERROR_TRUNCATE` was added to solve — a lesson this repo already learned. | Claude | security |
| 4 | `_kamiwaza_cli_live.py:13-15` | `_JWT_PATTERN` matches **any** dotted triple, so `api.example.com`, `kamiwaza_sdk.cli.main`, `tests.integration.conftest` and `1.1.0` all become `[REDACTED_TOKEN]` (all verified). A Python traceback on stderr has every module path scrubbed, defeating the docstring's promise of "actionable CLI diagnostics." Anchor on the JWT header (`eyJ[A-Za-z0-9_-]{10,}\.…`) or require a minimum segment length. | Claude | error handling |
| 5 | `_kamiwaza_cli_live.py:97` | `assert cached["access_token"] == pat_token` would print **both full tokens** if this module were ever assertion-rewritten — demonstrated by registering it and getting `assert 'SUPER_SECRET_CACHED_PAT' == 'SUPER_SECRET_STDOUT_PAT'`. It is safe today only because the file is not a `test_*.py` and is not registered: an undocumented, fragile invariant sitting on a credential path. Compare digests, or pin the invariant with a comment. | Claude | security |

| 1 | `tests/integration/test_cli_live.py:91-92` | `test_cli_login_and_pat_flow` revokes its PAT with no `try/finally`, so a `client_factory` or `pat_jti` failure leaks it (openid scope, lower impact). Asymmetric with the hardening applied one test below. | Claude | correctness |
| 2 | `_kamiwaza_cli_live.py:60-70` | `cli_login_and_create_pat` takes **8 parameters**, against the ≤4 cap in this repo's own `.ai/rules/codescene-authoring.md`. A frozen dataclass for the connection args would bring it under. | Claude | code quality |
| 3 | `_kamiwaza_cli_live.py:22-29` | Password redaction only replaces the literal `--password` value; a JSON-escaped or URL-encoded echo would not match. Low risk given current CLI behaviour. | Claude | security |
| 4 | `kamiwaza_sdk/cli.py:53` | Out of scope, worth filing: `pat create --scope` defaults to `openid`, which per the documented scope mapping **cannot drive the CLI's own `serve deploy`**. The CLI's default PAT is unusable for its primary command and `--help` doesn't say so. | Claude | documentation |

## Test Coverage Assessment

**The redaction guard genuinely bites — 4/4 mutations caught**, which is the part of this PR that most needed proving:

| Mutation | Result |
|---|---|
| `_JWT_PATTERN` → never matches | **FAILED** `test_run_cli_reports_sanitized_stderr` — raw JWT surfaced |
| Password redaction → `pass` | **FAILED** same test — password surfaced |
| `--scope` hardcoded to `openid` | **FAILED** `test_cli_login_and_create_pat_uses_requested_scope[admin]` |
| `pat_jti` reads `sub` instead of `jti` | **FAILED** `test_pat_jti_extracts_exact_cleanup_identifier` |

**The cleanup guard does not bite.** `_cleanup_cli_resources` is referenced only at its `def` and its single call site; no unit test imports it, so mutations are vacuous off-cluster (High 2).

**Redaction coverage measured by shape**: canonical 3-part base64url JWTs redact (including quoted and `bearer=`-prefixed); opaque hex/base64 tokens, newline-wrapped JWTs, and std-base64 `+`/`/` variants do not. The misses are theoretical — PATs and session tokens in this system *are* 3-part base64url JWTs (`pat_jti` hard-requires it) — so the shapes actually in play are covered.

**Coverage delta — nothing dropped.** `test_cli_live.py` is +92/−90 but all six original assertions survive (four relocated verbatim into the support module), plus one new (`assert not cleanup_failures`) and four new unit tests. Test-method count 2 → 2; net coverage up.

**Results**: new unit file **4 passed**; full unit suite **3109 passed, 1 skipped** (the established BSD-sed waiver). `ruff` and `mypy` clean on all three files. Live tests collect (2) but cannot run without a cluster and abort loudly rather than skipping silently.

## Positive Observations
- **The `check=True` → `check=False` switch is a real security win.** The old code let `CalledProcessError` carry PAT-bearing stdout into the exception; `run_cli` now surfaces sanitized stderr only and drops stdout entirely.
- **The scope escalation is correctly grounded, not a convenient privilege grab.** I went looking for a masked authz bug and found the opposite: `conftest.py:1830-1836` documents that admin scope is required for serving endpoints (write scope resolves to user/editor/viewer and produces ~22 403s), and `test_write_scope_access.py` (ENG-2861) is a deliberate guard asserting admin-gated endpoints *do* 403 at write scope. This brings the CLI test in line with what the rest of the integration suite already does; `live_username` already defaults to `admin`.
- **The `runner` dependency-injection seam** is what makes the scope contract unit-testable at all, and the mutation results prove it works.
- **Exception chaining verified intact** — a cleanup failure in `finally` does not hide the original test failure; pytest prints both via `__context__`.
- **Nothing leaks into the published artifacts.** Both wheel and sdist were built and inspected: no `_kamiwaza_cli_live`. `[tool.setuptools.packages.find]` discovers packages only and no `py-modules` is declared, so a root module cannot be included — and `python-build.yml` already asserts wheel contents. Placement also has precedent (`_kamiwaza_pytest_options.py` is an existing sibling).
- The removed `except Exception: pass` was masking `stop_deployment` failures *and* there was no PAT revocation at all, so the previous state leaked a live deployment and a PAT on every run. The direction of travel is clearly right.

## Overall Verdict
**Risk Level**: High
**Verdict**: REQUEST_CHANGES

The intent is right and parts are well executed — the scope escalation is grounded in the repo's own documented contract rather than papering over an authz bug, the redaction guard is mutation-proven, and nothing reaches the wheel. But the new cleanup force-stops **any** deployment that appeared since its baseline snapshot, with no ownership check and now wielding an admin PAT, so on a shared cluster it can terminate another run's workloads. And the `try:` opens four lines too late, so on the `pytest.skip()` path this test is *designed* to take on most hosts, the admin PAT is never revoked — the opposite of the PR's stated goal. Both are small, local fixes; the cleanup logic also needs a unit seam so they can be proven rather than asserted.

**Reviewed-SHA**: `84c9a0968d9bad40377475b9e8e45429aa33da25`

---

nickz9:
# Multi-Engine PR Review Report: [ENG-10397] Fix CLI serve-deploy live PAT scope and cleanup

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/271
**Branch**: `feature/eng-10397-sdk-cli-serve-deploy-live-test-uses-non-admin-pat-and-hides` → `release/1.1.0`
**Engines**: Claude Code, OpenAI Codex (2/3 available — Kamiwaza-hosted model unavailable: no endpoint or system PAT configured)

## CI & Merge Status
| Item | Status |
|------|--------|
| Checks | Green — 7 passing, 0 failing |
| Mergeable | UNKNOWN (GitHub still computing at review time) |
| Review decision | REVIEW_REQUIRED |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 8/10 | Extraction into a shared audited auth module is the right shape; the PAT is created outside the caller's `try:`, which is what makes Critical 1 reachable. |
| Security | 6/10 | The dominant leak is correctly fixed at its root, but an `admin`-scoped PAT can still outlive the test for its full TTL. |
| Test Quality | 6/10 | Redaction and blast-radius contracts are genuine and mutation-proven; three protections stay green with the protection removed, and the new module is deselected from the documented unit lane. |
| Error Handling | 8/10 | Cleanup accumulates failures rather than short-circuiting, so a deployment-stop failure no longer suppresses PAT revocation. |
| Code Quality | 9/10 | Small single-responsibility helpers; ruff and mypy clean on all three files. |
| Refactoring | 9/10 | Both live tests now share one auth path; module placement follows the existing root-module convention and packaging is unaffected. |
| Documentation Accuracy | N/A | No Markdown changed (three `.py` files) — the PR-body claim inaccuracy is filed as a finding instead. |

## Critical Issues (Must Fix)
| 1 | `_kamiwaza_cli_live.py:187-191` | **An `admin`-scoped PAT can survive the test.** `cli_login_and_create_pat` creates the PAT server-side, then executes three raise points *before returning*: `assert pat_token`, `json.loads(config.token_path.read_text())` (also `KeyError` on `access_token`), and the digest-equality assert. The caller creates the PAT **outside** its `try:` (`tests/integration/test_cli_live.py:82-89`), so on any of those failures the `finally:` is never entered and `_cleanup_cli_resources` never runs — the admin PAT lives on the shared cluster for its full **900 s TTL**. The PR adds a unit test (`test_cli_pat_cache_mismatch_does_not_expose_tokens`) for exactly this digest-mismatch path, so it is a designed-for scenario, not a hypothetical. This also qualifies the PR's own claim to "revoke both test PATs in `finally` blocks". Fix: return immediately after `pat_token = result.stdout.strip()` and move the cache/digest verification into the caller's `try:`, or pass a revoker and wrap the post-creation block in `except BaseException: revoke; raise`. | Claude, Nick | Security / Cleanup |

## High Priority Issues
| 1 | `tests/unit/test_cli_live_support.py:166-181` | **The headline "PAT is revoked even when the deployment stop fails" invariant is not pinned.** This is the only test where `stop_deployment` raises, and it never asserts `revoke_pat` was called. Mutation-verified: inserting `_raise_cleanup_failures(...)` inside the deployment `except` — so the revoke is skipped entirely — leaves **all 10 tests green**. The code's ordering is correct; nothing defends it. Fix: assert `revoke_pat` was called once with the expected jti. | Claude | Test Quality |
| 2 | `tests/unit/test_cli_live_support.py:184-209` | **The "compare digests, not raw credentials" contract is not pinned**, and the root cause is interesting: `_kamiwaza_cli_live.py` is not in pytest's assert-rewrite set (no `register_assert_rewrite`, and the filename doesn't match `python_files`), so every bare assert in it raises `AssertionError('')` with no explanation text. Mutation-verified: reverting the digest comparison to a raw `cached["access_token"] == pat_token` comparison leaves all 10 tests green. The real protection today is that the module sits outside a rewritten test module; the digest is unverified belt-and-braces. Fix: add `pytest.register_assert_rewrite("_kamiwaza_cli_live")` to the root `conftest.py` — which both restores useful diagnostics and makes the test genuinely pin the digest. | Claude | Test Quality |
| 3 | `tests/integration/test_cli_live.py:96-106` | **A deployment is orphaned when the CLI exits before printing its summary.** `serve deploy --wait` can create the deployment and then fail during the wait (timeout or terminal error); `run_cli` raises, `deployment_id` stays `None`, and `_cleanup_cli_resources` skips the stop — leaving a GPU-holding deployment on the shared cluster. Malformed CLI output has the same effect. This is the deliberate flip side of "only stop what we were told about" (the right trade-off), but it is the most likely real-world failure of this test. Mitigation: run `serve deploy` without `--wait` to capture the id, then poll via the SDK inside the `try:`. | Claude, Codex | Resource leak |

| 1 | `tests/unit/test_cli_live_support.py` | **The new contracts are deselected from the documented unit lane.** The module never sets `pytestmark = pytest.mark.unit`, and `make test-unit` runs `uv run pytest -m "unit"`. I verified this directly: the new file has no marker while all four sampled sibling files in `tests/unit/` do, so every new redaction, cleanup and ownership contract is skipped by the documented lane. Add the module marker. | Codex, Nick | Test Quality |
| 2 | `_kamiwaza_cli_live.py:41-46` | `--password=VALUE` (single-arg form) is not redacted — verified leaking. Unreachable from this module, which always emits `["--password", value]`, but `run_cli` is exported and generic. Add a `startswith("--password=")` branch. | Claude, Nick | Security |
| 3 | `_kamiwaza_cli_live.py:15-17` | Regex edges: the negative lookbehind means a JWT glued to a preceding `[A-Za-z0-9_-]` is not redacted, and an unsigned JWT (empty signature) never matches. All realistic delimiters — space, `=`, quote, `/`, `:`, `(`, newline — redact correctly, so practical risk is low. | Claude | Security |
| 4 | `tests/unit/test_cli_live_support.py:167` | The truncation test places the token at the *start* of a 600-char string, so it would still pass if the code truncated before redacting (mutation-verified: survived). The code's order is correct; move the token to straddle `_CLEANUP_ERROR_MAX_LENGTH` to pin it. | Claude | Test Quality |
| 5 | `_kamiwaza_cli_live.py:50-56` | `_redact_cleanup_error` calls `_redact_cli_stderr(str(exc), [])` with an empty argv, so password redaction is disabled for cleanup errors and only the JWT pattern applies. Cleanup errors originate from SDK client calls that don't carry the password, so this is defence-in-depth rather than a live gap. | Nick | Security |
| 6 | `tests/integration/test_cli_live.py:82-85` | The admin PAT's lifetime spans the whole deploy test — model readiness plus the wait — rather than just `serve deploy`, because the CLI authenticates from the on-disk token cache. Bounded by the 900 s TTL. Tighter: mint `openid` for readiness and re-mint `admin` immediately before the deploy. | Claude | Least privilege |

## Test Coverage Assessment

The protections were **mutation-tested**, six mutations applied to `_kamiwaza_cli_live.py` with the full suite re-run after each and the source restored byte-identical:

| Mutation | Result |
|---|---|
| JWT redaction → no-op | **caught** (2 failed) |
| Password redaction removed | **caught** (1 failed) |
| Cleanup stops deployment even when id is `None` | **caught** (1 failed) |
| PAT revoke skipped when deployment stop fails | **survived** → High 1 |
| Digest comparison → raw token comparison | **survived** → High 2 |
| Truncate before redact | **survived** → Medium 4 |

So the redaction and blast-radius contracts are real and load-bearing; the revocation-ordering, digest and redact-before-bound contracts assert on shapes that pass with the protection removed. Baseline: 10 passed in 0.49 s; `ruff check` and `mypy` clean on all three files. Compounding this, Medium 1 means none of these 10 run in `make test-unit` at all.

## Positive Observations
- **The root cause is fixed, not the symptom.** `check=True` → `check=False` (`_kamiwaza_cli_live.py:79`) removes `CalledProcessError.__str__` echoing the full argv including `--password <value>`. That was the actual leak.
- **Redact-before-bound ordering is correct**, confirmed empirically with a token straddling the 500-char cap — bounding cannot leak a prefix.
- **The blast-radius narrowing is genuine and test-pinned.** The deployment id can only come from this invocation's own stdout JSON; `_cleanup_cli_resources` guards on `if deployment_id:`, and there is no `list_deployments`, no name/prefix matching, no namespace sweep and no reconcile anywhere in the path — `stop_deployment` appears exactly once. Changing the guard to `if True:` is caught by a test.
- **Cleanup accumulates failures instead of short-circuiting**, so the PAT revoke still executes after a deployment-stop failure, and the old `except Exception: pass` swallow is gone — failures now surface, redacted.
- **PATs are genuinely JWTs** (`pat_jti` raises unless the token has three parts), which is what makes the `eyJ...` redaction pattern actually cover them — I checked this specifically, since an opaque PAT format would have left the main sink unprotected.
- **`field(repr=False)` on `env` and `password`** shows real attention to the dataclass-repr sink, and `--showlocals` is not enabled anywhere in the repo or CI.
- **Scope minimization holds**: `scope` defaults to `openid`, `admin` is passed at exactly one call site, and that selection is parametrically pinned by a test. The auth-only test now also revokes its PAT, which it previously never did.

## Overall Verdict
**Risk Level**: Medium
**Verdict**: REQUEST_CHANGES

This PR fixes the real leak correctly and narrows a genuinely dangerous cleanup path, and the extraction into a shared audited module is a clear improvement. What blocks is that the central promise — "the PAT is always revoked" — has a reachable hole: `cli_login_and_create_pat` creates the PAT and then raises at three points before returning, entirely outside the caller's `try:`/`finally:`, so an `admin`-scoped credential survives for its full 900 s TTL. Both engines and my own trace landed on this independently, and the PR ships a unit test for precisely that digest-mismatch path.

Compounding it, mutation testing shows the two contracts that would have caught cleanup regressions — revoke-after-stop-failure and digest-not-raw-token — stay green with the protection removed, and the whole new contract module is deselected from `make test-unit` for want of a `pytestmark`. The fixes are small and local: move the post-creation verification inside the caller's `try:`, add the two missing assertions, register the module for assert rewriting, and add the unit marker.

---

**Reviewed-SHA**: b7a781f1969d65012a54b2618d1c987f7cfbb13e

nickz9:
# RE-REVIEW Report: [ENG-10397] Fix CLI serve-deploy live PAT scope and cleanup

**PR**: https://github.com/kamiwaza-ai/kamiwaza-sdk/pull/271
**Branch**: `feature/eng-10397-sdk-cli-serve-deploy-live-test-uses-non-admin-pat-and-hides` → `release/1.1.0`
**Engines**: Claude Code, OpenAI Codex (2/3 available — Kamiwaza engine not configured)
**Head**: `aa1b42eb` · prior REQUEST_CHANGES at `84c9a096` (since rebased away) · 4 files, +503/−100

## CI & Merge Status
| Item | Status |
|------|--------|
| Checks | **7 pass** · 0 fail · 0 pending |
| Mergeable | `MERGEABLE` / `BLOCKED` (`REVIEW_REQUIRED`) |
| Base | Rebased onto `effb0218` — 3 linear commits, no merges, nothing dropped |

## Round-1 Finding Status — all six fixed
| # | Prior finding | Status | Evidence |
|---|---|---|---|
| **C1** | Cleanup force-stopped unowned deployments | ✅ **FIXED — deleted, not guarded** | The baseline-diff logic is gone entirely. `grep stop_deployment\|list_deployments\|baseline` across all 4 files returns exactly **one** call site (`_kamiwaza_cli_live.py:124`), guarded by `if deployment_id:`. Provenance traced: the id comes from `json.loads(serve_result.stdout)` of *this test's own* `serve deploy`. Mutation M6 (`if deployment_id:` → `if True:`) is caught. |
| **I1** | Admin PAT leaked on the common skip path | ✅ **FIXED — proved by construction** | `ensure_deployable_model_ready` is now inside `try:` (`test_cli_live.py:96`), so `finally` covers `Skipped` (a `BaseException`). Driving the real test function with a readiness callable that calls `pytest.skip()` yields `raised=Skipped, cleanup_calls=[('cleanup', None, <PAT>)]` — cleanup fires. Same on the error path. Also confirmed the capacity gate is an autouse *fixture*, so gated hosts skip at setup **before** any PAT is minted. |
| **I2** | Cleanup had zero unit coverage | ✅ **FIXED** | Moved into the support module with 3 dedicated tests. **9/9 mutations caught**, including "swallow cleanup failures" and "skip the PAT revoke" — a real guard, not just a seam. |
| **I3** | Raw `{exc}` with no redaction or truncation | ✅ **FIXED** (one path remains, see Medium 2) | `_redact_cleanup_error` applies `_redact_cli_stderr` **then** truncates at 500 chars — correct ordering, so a JWT straddling the boundary is redacted before the cut. Both `_cleanup_cli_resources` failure paths route through it. |
| **I4** | `_JWT_PATTERN` over-matched any dotted triple | ✅ **FIXED** | Anchored on `eyJ` + ≥10 chars. Probed: real HS256 JWTs and the test PAT **redact**; `api.example.com`, `kamiwaza_sdk.cli.main`, `tests.integration.conftest`, `1.1.0` and `eyJab.cd.ef` all **survive intact**. Tracebacks stay readable. |
| **I5** | Raw token comparison under assertion rewriting | ✅ **FIXED** | `assert_cli_pat_cache_matches` compares sha256 digests only; grep finds no raw-token comparison in the diff. Mutation M5 (compare digest to itself) is caught, so the comparison genuinely detects mismatch. |
| S6/S7 | `try/finally` on the login test; 8-param helper | ✅ **FIXED** | `test_cli_login_and_pat_flow` now uses `try/finally`; `cli_login_and_create_pat` went 8 params → 4 via `CliAuthConfig`, meeting the repo's own ≤4 cap. |

## Dimension Scores
| Dimension | Score | Notes |
|-----------|-------|-------|
| Design Adherence | 9/10 | Cleanup scoped to a self-produced id; helpers extracted behind a testable seam. |
| Security | 8/10 | Credential paths re-traced end-to-end post-restructure; one un-redacted revoke remains (Medium 2). |
| Test Quality | 9/10 | 9/9 mutations caught and the redaction test is well-designed; the structural fix itself is unguarded (Medium 1). |
| Error Handling | 8/10 | Cleanup is fail-open per step — a failing stop does not prevent the PAT revoke; failures aggregate and raise once. |
| Code Quality | 9/10 | Every function ≤4 params and ≤45 lines; `ruff check` and `mypy` clean. |
| Refactoring | 9/10 | Moving the logic into the support module is what made I2 fixable at all. |
| Documentation Accuracy | 9/10 | Docstrings now describe the actual contract ("Stop only this test's reported deployment"). |

## Critical Issues (Must Fix)
| — | — | None. Codex independently reports "no actionable correctness defects." | — | — |

## High Priority Issues
| — | — | None. | — | — |

| 1 | `tests/integration/test_cli_live.py:95` | **The I1 fix has no regression guard.** Hoisting `ensure_deployable_model_ready` back above the `try` — reintroducing the exact round-1 bug shape — leaves **all 10 unit tests passing** (measured). The helper *functions* are thoroughly guarded; the test *structure* that was the actual defect is not. A unit test driving `test_cli_serve_deploy` with fake fixtures and asserting cleanup fires on `Skipped` would close it. | Claude | test coverage |
| 2 | `tests/integration/test_cli_live.py:48` | The one cleanup path bypassing `_redact_cleanup_error`: the bare `finally` calls `revoke_pat` directly, and `APIError` embeds the full unbounded response body (`client.py:470`). Not a demonstrated credential leak — the DELETE carries the jti, not the token — but it is the same unbounded-body class I3 fixed, and the helper already exists. Routing it through `_cleanup_cli_resources(client, None, pat_token)` would unify both tests. | Claude | consistency |
| 3 | `tests/integration/test_cli_live.py:99-113` | Dropping `--wait/--poll-interval/--timeout` in favour of `wait_deployment_ready()` removes the only live coverage of the CLI's own wait/poll path. The trade is **correct** — with `--wait`, a deploy timeout exits non-zero, `run_cli` drops stdout, and the `deployment_id` needed for cleanup is lost — but the rationale isn't recorded. A one-line comment would stop someone "restoring" `--wait`. | Claude | test coverage |
| 4 | `_kamiwaza_cli_live.py:37` | `_redact_cli_stderr` only redacts JWT-shaped tokens; an opaque `kwz_pat_…` survives. Moot today (platform PATs and session tokens are JWTs — `pat_jti` requires 3 parts), so future-proofing only. | Claude | security |

## Test Coverage Assessment
**9 of 9 mutations against the support module are caught**, which is what makes this round's fixes credible rather than asserted:

| Mutation | Caught by |
|---|---|
| Swallow cleanup failures | `test_cleanup_redacts_and_truncates_failure_details` |
| Skip the PAT revoke entirely | **3 tests** |
| Drop redaction from cleanup errors | `test_cleanup_redacts_and_truncates_failure_details` |
| Drop truncation | same |
| Neuter the cache digest comparison | `test_cli_pat_cache_mismatch_does_not_expose_tokens` |
| Remove the `if deployment_id:` guard | `test_cleanup_revokes_pat_when_no_deployment_was_reported` |
| `run_cli` leaks raw stderr | `test_run_cli_reports_sanitized_stderr` |
| `run_cli` appends stdout on failure | same |
| `--scope` hardcoded to `openid` | `test_cli_login_and_create_pat_uses_requested_scope[admin]` |
| **Structural: hoist readiness above `try`** | ❌ **none** — Medium 1 |

**Results**: `test_cli_live_support.py` **10 passed**; full unit suite **3121 passed, 1 skipped** (the skip is the established BSD-sed waiver); `ruff check` and `mypy` clean on all 4 files.

**Stated plainly**: the live/integration tests cannot run here — they need a cluster, `--live-base-url` and credentials. The real `serve deploy` round-trip, the admin-scoped PAT mint, and the real `stop_deployment`/`revoke_pat` calls are unverified by me; the evidence for those is the mutation-tested helpers plus the constructed-skip harness.

**Rebase integrity**: `effb0218..aa1b42e` is 3 linear commits touching only the 4 files in the final diff. The round-1 state matches `afa853b` exactly (`_kamiwaza_cli_live.py +112`, `test_cli_live_support.py +86`), and the two later commits are additive hardening. Nothing was lost in the rebase.

## Positive Observations
- **The Critical is gone rather than guarded.** The baseline-diff-and-force-stop logic was *deleted*; the module now has a single `stop_deployment` tied to an id this test produced itself. That is the right resolution — the ownership filter I suggested would have been the minimum, and removing the enumeration entirely is stronger.
- **The I1 fix was verified by construction, not by reading.** Putting the readiness call inside `try` handles `Skipped` correctly because it derives from `BaseException` and `finally` is unconditional — and driving the real test proved cleanup fires on both the skip and error paths.
- **Redaction ordering is correct — redact, then truncate** — and the test deliberately plants the PAT at offset 470 so it straddles the 500-char cut, asserting the first 10 chars are absent. That is a designed test, not a box-tick.
- **Cleanup is fail-open per step**: a failing `stop_deployment` does not prevent the PAT revoke; failures aggregate and raise once.
- **Packaging remains safe** — the root module follows the existing `_kamiwaza_pytest_options.py` precedent, `py-modules` is unset, and `packages.find` matches only `kamiwaza_*`, so it cannot reach the wheel.
- Complexity discipline holds throughout: ≤4 params, ≤45 lines per function, 211-line file — meeting both the CodeScene authoring rules and the stricter repo style caps.

## Overall Verdict
**Risk Level**: Low
**Verdict**: APPROVE

All six round-1 findings are fixed, and each was verified by evidence rather than inspection — the destructive baseline-diff cleanup is deleted outright, the skip-path PAT leak is disproved by constructing the skip, the JWT pattern now leaves tracebacks readable, and 9 of 9 mutations against the new helpers are caught. Codex independently found no actionable defects. The remaining items are one un-redacted revoke call and a missing structural regression guard for the `try`-placement fix itself — worth landing, neither worth blocking.

**Reviewed-SHA**: `aa1b42eb269dd502c22488419111b6bdfcd3ef0c`

---

nickz9:
## pr-verify verdict — historical output redacted

Historical verification details were removed from this public PR. Re-run pr-verify before relying on a current gate result.

kamiwaza-pr-verify (approved):
## pr-verify verdict — 🟢 **APPROVE**
_All required auto-checks passed (required-checks, linear-issue-present, no-debug-statements, pr-review-approved); 4 manual steps verified._
### Auto-checks
| | Check | Required | Status | Detail |
|---|-------|----------|--------|--------|
| ✅ | `required-checks` | required | pass | 7/7 checks green |
| — | `pr-evidence` | optional | absent | no pr-evidence sentinel block in PR body |
| ✅ | `linear-issue-present` | required | pass | found: ENG-10397 |
| ✅ | `no-debug-statements` | required | pass | diff scanned; no debug statements outside excluded paths (tests / scripts / bin / cli / tools / docs) |
| ✅ | `pr-review-approved` | required | pass | /pr-review APPROVE by nickz9 on aa1b42eb269d (self-review comment, SHA-pinned) |
### Manual steps
| | ID | Description | Observation |
|---|---|---|---|
| ✅ | `ms-1` | Verify the round-1 Critical is resolved: cleanup must no longer be able to stop  | Reproduced by the reviewer. Round 1 found that _cleanup_cli_resources computed cleanup_ids = current_deployment_ids - ba |
| ✅ | `ms-2` | Verify the admin PAT is no longer leaked on the skip path, which is the expected | Proved by construction, not by reading. Round 1 found the PAT was minted before try: opened, with client_factory, ensure |
| ✅ | `ms-3` | Verify the credential-safety controls actually bite, and re-trace every path a t | Reproduced by the reviewer. Nine of nine mutations against _kamiwaza_cli_live.py are caught by tests/unit/test_cli_live_ |
| ✅ | `ms-4` | Record the state of the live evidence and what remains unverified, so the gate d | Recorded by the reviewer. ENG-10397's live-evidence attachment 'CLI live final JUnit — 2/2 pass' is pinned to commit 84c |
### Gate
- verified_head_sha: `aa1b42eb269dd502c22488419111b6bdfcd3ef0c`
- pr_created_at: `2026-08-14T20:03:21Z`
- approval_submitted: `true`
- ✅ APPROVE submitted (as review) on commit \`aa1b42eb269d\`
<sub>Re-validated at 2026-08-18T17:48:07.262Z · tool_version 0.1.0 · format_version 0</sub>
<sub>Canonical format: [pr-verify REFERENCE.md](https://github.com/kamiwaza-internal/kamiwaza-engineering-marketplace/blob/main/plugins/pr-verify/skills/pr-verify/REFERENCE.md)</sub>

kamiwaza-pr-verify:
## pr-verify verdict — historical output redacted

Historical verification details were removed from this public PR. Re-run pr-verify before relying on a current gate result.

